PR #14, opened 2024-08-28
Various updates for EA

soluwalana:
## Description

Updates Helm Chart Dependencies
Updates log configuration
Updates debug mode selection based on DEBUG level
Improve caching for Dockerfile

## Checklist

### DIFF
--- a/src/nv_ingest/extraction_workflows/pdf/pdfium_helper.py
+++ b/src/nv_ingest/extraction_workflows/pdf/pdfium_helper.py
@@ -5,6 +5,7 @@
 
 import io
 import logging
+import os
 from math import ceil
 from math import floor
 from math import log
@@ -23,6 +24,7 @@ from nv_ingest.schemas.pdf_extractor_schema import PDFiumConfigSchema
 from nv_ingest.util.converters import bytetools
 from nv_ingest.util.image_processing.table_and_chart import join_cached_and_deplot_output
 from nv_ingest.util.image_processing.transforms import numpy_to_base64
+from nv_ingest.util.logging.configuration import configure_logging
 from nv_ingest.util.pdf.metadata_aggregators import Base64Image
 from nv_ingest.util.pdf.metadata_aggregators import ImageChart
 from nv_ingest.util.pdf.metadata_aggregators import ImageTable
@@ -53,6 +55,7 @@ from nv_ingest.util.triton.helpers import perform_model_inference
 
 
 logger = logging.getLogger(__name__)
+configure_logging(logger, os.getenv("INGEST_LOG_LEVEL", "INFO"))
 
 
 def extract_tables_and_charts_using_image_ensemble(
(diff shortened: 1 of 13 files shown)


src/nv_ingest/extraction_workflows/pdf/pdfium_helper.py
```diff
@@ -53,6 +55,7 @@
 
 
 logger = logging.getLogger(__name__)
+configure_logging(logger, os.getenv("INGEST_LOG_LEVEL", "INFO"))
```
drobison00: This shouldn't be required, logging should be globally configured here: https://github.com/NVIDIA/nv-ingest/pull/14/files#diff-a5b564197d741ea776ee29ab7192991961c613741edc265c9c623c58c0a623cdL657

src/pipeline.py
```diff
@@ -40,10 +40,11 @@
 from nv_ingest.stages.transforms.image_caption_extraction import generate_caption_extraction_stage
 from nv_ingest.util.converters.containers import merge_dict
 from nv_ingest.util.logging.configuration import LogLevel
+from nv_ingest.util.logging.configuration import configure_logging as configure_local_logging
 from nv_ingest.util.schema.schema_validator import validate_schema
 
 logger = logging.getLogger(__name__)
-logger.setLevel(logging.INFO)
+configure_local_logging(logger, os.getenv("INGEST_LOG_LEVEL", "INFO"))
```
drobison00, line 47: Should not be done here, this is handled here: https://github.com/NVIDIA/nv-ingest/pull/14/files#diff-a5b564197d741ea776ee29ab7192991961c613741edc265c9c623c58c0a623cdL657

src/nv_ingest/stages/pdf_extractor_stage.py
```diff
@@ -18,8 +19,10 @@
 from nv_ingest.schemas.pdf_extractor_schema import PDFExtractorSchema
 from nv_ingest.stages.multiprocessing_stage import MultiProcessingBaseStage
 from nv_ingest.util.exception_handlers.pdf import create_exception_tag
+from nv_ingest.util.logging.configuration import configure_logging
 
 logger = logging.getLogger(f"morpheus.{__name__}")
+configure_logging(logger, os.getenv("INGEST_LOG_LEVEL", "INFO"))
```
drobison00: Should not be required, this will be propagated via: https://github.com/NVIDIA/nv-ingest/pull/14/files#diff-a5b564197d741ea776ee29ab7192991961c613741edc265c9c623c58c0a623cdL657

drobison00 (changes requested):

src/pipeline.py
soluwalana: It was previously hardcoded to logging.INFO and no debug logs were produced from the container from this file when setting the `log_level` parameter via ENV or command line.

```
from morpheus.utils.logger import configure_logging

configure_logging(log_level=log_level)
```

Did not appear to affect the existing logger defined in this file, it creates a new logger at https://github.com/nv-morpheus/Morpheus/blob/branch-24.06/morpheus/utils/logger.py#L118

PR #44, opened 2024-09-05
Resolves #42 - Errors raised during the extraction stage are obscured by a raised payload merge failure.

### LINKED ISSUE #42: [BUG]: Errors raised during the extraction stage are obscured by a raised payload merge failure.
### Version

24.08

Source

### Describe the bug.

During document extraction, failures to process the document are currently being shadowed by a subsequent failure to update the message payload. To resolve this, we need update the way that error information is propagated back from the shared worker pool, and through the multiprocessing_stage.

### Completion requirements
- [ ] Modify existing multiprocess_stage code to correctly forward raised exception messages via the returned `work_package`
- [ ] Modify existing extract_and_decode function to raise an exception instead of trying to error an error_message structure (this should be built and handled at a higher level)
- [ ] Update the client CLI code to better indicate the processing events that lead to the Job failure and report them to the caller.

_No response_

### Relevant log output

_No response_

### Other/Misc.

_No response_

drobison00:
Resolves #42 

## Checklist


client/src/nv_ingest_client/cli/util/processing.py
```diff
@@ -313,24 +539,246 @@ def create_job_specs_for_batch(files_batch: List[str], tasks: Dict, client: NvIn
...
+    specified directory. If the job fails, it raises a `RuntimeError` and returns a list of
+    retry job IDs.
+
+    See Also
+    --------
+    check_ingest_result : Function to validate the result of the job.
+    save_response_data : Function to save the result to a directory.
+    """
+
+    result, _ = future.result()[0]
```
jdye64: It seems like [client.py:_submit_job(...)](https://github.com/NVIDIA/nv-ingest/blob/main/client/src/nv_ingest_client/client/client.py#L407) can possibly raise an exception here. Since that code is ran in the ThreadPoolExecutor the exception would present itself here at the `.result()` call. I realize the JobState would already be set to failed and we could just let the event bubble on up but should we also add a try/except here too just to make the intent more clear?

jdye64 (changes requested):
Looks good. Lots more good function docs. Had a thought about exception handling other than that LGTM

client/src/nv_ingest_client/cli/util/processing.py
drobison00: I'll add a try catch that logs the error and re-raises.

PR #64, opened 2024-09-09
Barebones workflow to trigger pre-commit github action

jdye64:
## Description
Barebones Github Action configuration to install and run `pre-commit` on the nv-ingest codebase

Developers will need to now make sure that `pre-commit` is installed in their development environment. This is included in the `util-requirements.txt` file now or you can simply run `pip install pre-commit`. 

Then you must install the pre-commit hooks with `pre-commit install`

If you find yourself working on a PR prior to this commit that does not have pre-commit you can skip the pre-commit checks and commit time by running `git commit -m "commit without checking" --no-verify` 

This checks for 
- Trailing whitespaces
- Fixes end of lines to include a single blank line
- check that no files larger than 1,500kb are committed. Maybe we adjust later?
- check that python files are valid
- format and lint with black to PEP8 standard

## Checklist


.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,19 @@
...
+    push:
+      branches: [main]
+
+jobs:
+  Explore-GitHub-Actions:
+    runs-on: ubuntu-latest
+    steps:
+      - run: echo "🎉 The job was automatically triggered by a ${{ github.event_name }} event."
+      - run: echo "🐧 This job is now running on a ${{ runner.os }} server hosted by GitHub!"
+      - run: echo "🔎 The name of your branch is ${{ github.ref }} and your repository is ${{ github.repository }}."
```
jarmak-nv: We should remove this line; [it's vulnerable to script injection](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#example-of-a-script-injection-attack) if someone opens a PR and names their repo in a malicious manner.

(I selected the wrong part - sorry! I mean L14 since it echos the repo name, but tbh we probably should just remove them all.)

jarmak-nv: (Any user modifiable field is at risk of this, repo is probably the least likely but still something we shouldn't open ourselves to.)

jarmak-nv (changes requested):

.github/workflows/pre-commit.yml
jdye64: Agreed. Honestly I just copied this example from a silly little boilerplate example in their docs. I'll make some meaningful changes here shortly.

jperez999 (approved):

edknv (approved):

Issue #32, opened 2024-09-04
[DOC]: Add docker login step for nvcr.io


randerzander:
### How would you describe the priority of this documentation request

None

### Please provide a link or source to the relevant docs

github.com/nvidia/nv-ingest

### Describe the problems in the documentation

The user needs to `docker login nvcr.io` before they'll be able to pull requisite container images. The README needs to make this clear.

randerzander:
Closed by #67

PR #67, opened 2024-09-09
Simplify the NGC API key environment variable

### LINKED ISSUE #30: [FEA]: Simplify the required API key environment variables

New Feature

Significant improvement

Step 1 in our README shows the creation of environment variables, specifically the NGC API key. However, it uses multiple variable names for the same API key. This is confusing and for consistency,  we should use the same variable, that is, NGC_CLI_API_KEY.

we should use a single env var NGC_CLI_API_KEY, if possible

### Additional context

_No response_

edknv:
## Description
Closes #30 #32 

- I think multiple variable names for the API key came about because 1. there are two versions of personal keys: the old style ("Generate API Key") and the new style ("Generate Personal Key"), 2. the new-style key can be generated in multiple ways: auto-generated via `build.nvidia.com` and https://org.ngc.nvidia.com/setup/personal-keys, and 3. when creating the API key on https://org.ngc.nvidia.com/setup/personal-keys, you can choose which services to include.
- The personal key that is auto-generated on `build.nvidia.com` does not include all the services that nv-ingest needs. Accordingly, this PR proposes that we recommend to users that they manually generate the keys and include all the necessary services.
- With one single key that has access to AI Foundation Models and Endpoints, Private Registry, and NGC Catalog, I was able to pull all the containers, all the containers were able to download the models, and hit the `build.nvidia.com` endpoints.
- `nv-embedqa-e5-v5:1.0.0` had a bug and it only worked with old-style personal keys. `1.0.1` fixes this bug and it works with the same new-style personal keys that other NIMs use (as long as appropriate services are included).

## Checklist


docker-compose.yaml
```diff
@@ -25,7 +25,7 @@ services:
     environment:
       - NIM_HTTP_API_PORT=8000
       - NIM_TRITON_LOG_VERBOSE=1
-      - NGC_API_KEY=$NIM_NGC_API_KEY
+      - NGC_API_KEY=${NIM_NGC_API_KEY:-${NGC_API_KEY:-ngcapikey}}
```
jdye64, line 28: I appreciate the defaulting to the older ENV var here. Thanks for doing that!

PR #70, opened 2024-09-11
Resolves #69 join_cached_and_deplot_output function is not adding all fields returned from cached.

### LINKED ISSUE #69: [BUG]: join_cached_and_deplot_output function is not adding all fields returned from cached.
### Version

24.09

_No response_

### Describe the bug.

Function is currently only returning 'chart_title' but not the other captured fields.

_No response_

### Relevant log output

_No response_

### Other/Misc.

_No response_

drobison00:
Resolves #69 

## Checklist

### DIFF
--- a/src/nv_ingest/util/image_processing/table_and_chart.py
+++ b/src/nv_ingest/util/image_processing/table_and_chart.py
@@ -45,6 +45,10 @@ def join_cached_and_deplot_output(cached_text, deplot_text):
         try:
             cached_text_dict = json.loads(cached_text)
             chart_content += cached_text_dict.get("chart_title", "")
+
+            if (deplot_text is not None):
+                chart_content += f" {deplot_text}"
+
             chart_content += " " + cached_text_dict.get("caption", "")
             chart_content += " " + cached_text_dict.get("info_deplot", "")
             chart_content += " " + cached_text_dict.get("x_title", "")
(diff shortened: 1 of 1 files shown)


src/nv_ingest/util/image_processing/table_and_chart.py
```diff
@@ -38,13 +41,26 @@ def join_cached_and_deplot_output(cached_text, deplot_text):
     """
     chart_content = ""
 
-    if (cached_text is not None) and (deplot_text is not None):
+    if (cached_text is not None):
```
boliu61, line 44: The order should strictly be: chart_title from cached, then deplot output, then other cached fields. We have experimented with different ordering, and found this to be the best. i.e, deplot output should not be at the very end.

Also, `caption` is not a cached field. It was from the captioning model (no longer used).

So it should be something like 
```
    if (cached_text is not None):
        cached_text_dict = json.loads(cached_text)

    chart_content += cached_text_dict.get("chart_title", "")

    if (deplot_text is not None):
        chart_content += f" {deplot_text}"

    chart_content += " " + cached_text_dict.get("x_title", "")
    chart_content += " " + cached_text_dict.get("xlabel", "")
    chart_content += " " + cached_text_dict.get("y_title", "")
    chart_content += " " + cached_text_dict.get("ylabel", "")
    chart_content += " " + cached_text_dict.get("legend_label", "")
    chart_content += " " + cached_text_dict.get("legend_title", "")
    chart_content += " " + cached_text_dict.get("mark_label", "")
    chart_content += " " + cached_text_dict.get("value_label", "")
    chart_content += " " + cached_text_dict.get("other", "")
```
why do we need `except json.JSONDecodeError:`? Have you seen this error?

drobison00: > why do we need except json.JSONDecodeError:? Have you seen this error?

That was in the original code that this was created from.I took it as an indication that sometimes we would not get JSON back.

PR #82, opened 2024-09-19
Simplify Python SDK example and offer a convenience function that transparently handles fetch retries

jdye64:
## Description
Simplify Python SDK example and offer a convenience function that transparently handles fetch retries. This approach takes the existing `fetch_job_result` and renames it to `fetch_job_result_cli`. That function is then only used exclusively from the CLI codebase. The reasoning for this is because the timeout approaches and how they are handled are different between the CLI and how users wish to have them transparently handled for them.

This approach was chosen since there are several examples and notebooks using the `fetch_job_result` function and this helps ensure those existing examples need not be changed.

## Checklist


randerzander (approved):
Approving from the end user API perspective. I was able to confirm the example Python API snippet works as expected.

Need @ChrisJar or @edknv to review the client code changes

edknv (approved):

PR #87, opened 2024-09-20
Adjust all instances of port 6379 that needed to be converted to 7670

jdye64:
## Description
The REST endpoint is now the default. This means that all lingering examples where port 6379 were used to reference the Redis endpoint they need to be updated to use port 7670 which is the REST endpoint port.

There are still places where port 6379 is used and is valid in the pipeline code. This PR only removes code from client side and examples.

## Checklist


randerzander (approved):
Wow, thanks for finding all the remaining outdated port numbers! Great catches

PR #109, opened 2024-10-01
Trace image model inference time in pdf extraction

edknv:
## Description

This PR adds the latency of image model inference during PDF extraction to the OTEL trace.

The model inference steps are treated as subtasks of the PDF extraction process. For example, in Zipkin, there will be an expand button next to the PDF extraction span:

span_expand_button

Clicking the expand button will show a more detailed breakdown of the latency.

span_image_models

Our tracing mechanism was originally designed to work with stages, where each stage passes control messages and uses timestamps from those messages. However, the extraction workflows use an adapter pattern, where each stage calls a separate function that does not use control messages. To handle this, this PR passes down a dictionary to the adapter level, allow the adapters to modify it, and then send it back up to the stage level. Passing around mutable objects is probably not best, but I think other approaches have their own drawbacks (e.g., we could bypass the control message by having a separate object for storing timestamps, but that could lead to inconsistencies and would require more code changes.

## Checklist


jdye64 (approved):
Looks good @edknv. You had mentioned not liking passing the dict into the called functions but I think you made the correct choice and that is about as good as we can hope for without making things too overly complex. I also really appreciate the decorator that we can use to capture future traces.

jdye64:
.... and most importantly it works =)

PR #139, opened 2024-10-03
Add extract_charts option for disabling chart extraction

edknv:
## Description
This PR introduces an optional `extract_charts` parameter to both the CLI and Python client. When `extract_charts` is set to false, the process will skip the deplot and cached steps. If the parameter is not specified, the default behavior remains unchanged, ensuring backward compatibility by continuing to run both table and chart extraction as before.

## Checklist

### DIFF
--- a/src/nv_ingest/extraction_workflows/pdf/pdfium_helper.py
+++ b/src/nv_ingest/extraction_workflows/pdf/pdfium_helper.py
@@ -188,11 +188,11 @@ def extract_tables_and_charts_using_image_ensemble(
         logger.error(f"Error during table/chart extraction: {str(e)}")
         raise
     finally:
-        if extract_tables and isinstance(paddle_client, grpcclient.InferenceServerClient):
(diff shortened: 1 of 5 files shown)


src/nv_ingest/extraction_workflows/pdf/pdfium_helper.py
```diff
@@ -176,11 +188,11 @@ def extract_tables_and_charts_using_image_ensemble(
         logger.error(f"Error during table/chart extraction: {str(e)}")
         raise
     finally:
-        if isinstance(paddle_client, grpcclient.InferenceServerClient):
+        if extract_tables and isinstance(paddle_client, grpcclient.InferenceServerClient):
```
drobison00: I don't think you need these extra checks. If extract tables/charts aren't set, the corresponding clients should just be 'None'

drobison00 (approved):

PR #154, opened 2024-10-10
Move PaddleOCR preprocessing logic to nv-ingest

edknv:
## Description
- PaddleOCR NIM moved preprocessing logic out of Triton code and moved it into NIM code. We move the same preprocessing logic into nv-ingest since we are not yet confident with performance with HTTP endpoints. 
- Remove cache directory mounts in docker compose since NIMs are not expected to be run a second time.
- Add all grpc and http endpoints to `docker-compose.yaml` but add a switch so we can easily switch between grpc and http.

## Checklist


docker-compose.yaml
```diff
@@ -138,15 +130,16 @@ services:
       - sys_nice
     environment:
       - CACHED_GRPC_ENDPOINT=cached:8001
-      - CACHED_HEALTH_ENDPOINT=cached:8000
```
drobison00, line 141: Is removing the Health endpoint intentional?

drobison00: If so, can you explain.

src/nv_ingest/util/nim/helpers.py
```diff
@@ -258,3 +389,44 @@ def is_ready(http_endpoint, ready_endpoint) -> bool:
...
+
+    if not metadata_endpoint.startswith("/") and not url.endswith("/"):
+        metadata_endpoint = "/" + metadata_endpoint
+
+    url = url + metadata_endpoint
+
+    # Call the metadata endpoint of the NIM
+    try:
+        # Use a short timeout to prevent long hanging calls. 5 seconds seems resonable
+        resp = requests.get(url, timeout=5)
```
drobison00, line 411: I think we may want to think about some number of retries here, and even a standard for when we consider a NIM request to be completely failed. It definitely seems possible that a NIM could be down or unresponsive for a short time and then come back. Do we really want to fail the job if this happens?

edknv: Thanks for catching this. A standard for when we consider a NIM to have completely failed sounds like a good idea, but honestly I'm not sure exactly how that looks like. For now, I just added a simple retry on the function in https://github.com/NVIDIA/nv-ingest/pull/154/commits/caa4de888936239b2b3ce8e2c0066ddf41ae2fa8.

docker-compose.yaml
edknv: Yes, I wanted to simplify and consolidate `*_HTTP_ENDPIONT` and `*_HEALTH_ENDPOINT` variables into one. @jdye64 can correct me, but I think the original intention of the readiness check was to use the `*_HTTP_ENDPOINT` variables, parse them, and append `/v1/ready` to get the healthcheck endpionts. But we ended up adding `*_HEALTH_ENDPOINT` variables because we were not ready to make the switch to using http endpoints and moved back to having only grpc variables. Since this PR adds the http endpoint variables for all the NIMs, we can get the readiness endpoints by parsing the http endpoint variables.

drobison00 (approved):

PR #157, opened 2024-10-10
Add notebooks that use LlamaIndex to calculate recall scores for NV-Ingest chart and table extraction

ChrisJar:
## Checklist

## TODO/Roadblocks:
- [x] The vdb_upload task doesn't support configuring different collections so everything ends up in the same vector store. This means I need flush vdb to avoid muddling results. It'd be convenient if I could keep the chart embeddings and table embeddings separate somehow
- [x] I'm currently using the cli client to run NV-Ingest as it's better for handling multiple files and it directly outputs errors rather than returning an empty result so it can be easier to debug, but this should be switched to python eventually
- [x] LlamaIndex doesn't include some of the metadata when querying the milvus vdb so I lose important information like pdf id and page number which is used to determine if a retrieved result is the correct result. If I could do this it would remove a lot of the complexity around loading results into a dataframe (as seen in the first few cells of the notebook)
- [x] Add text recall evaluation section
- [ ] Add back rerankers (broke for some reason and need to debug but my NV-Ingest service is in use)
- [x] Add bo767 text recall 
- [ ] Figure out how to share text question answering datasets


examples/recall.ipynb
```diff
@@ -0,0 +1,288 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "3f7aa138-00ca-4a31-b0d0-63aa03902cf2",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "nv-ingest-cli \\\n",
```
randerzander: please update to use #158 instead of the nv-ingest-cli

PR #166, opened 2024-10-15
add in gpu milvus

jperez999:
## Description
This PR changes the milvus index to the GPU CAGRA index and it uses the parameters that created the best recall for a randomized index (it is possible to tweak this parameter to get better results but that is entirely based on the dataset) which is a good starting point. It also changes the method of identifying the GPUs used in docker-compose file to specify IDs instead of counts and leveraging the CUDA_VISIBLE_DEVICES env variable. I left the variable in place, but changed it to look for GPU ID 0, which is essentially a no op. If we like this new methodology we can remove that env var completely.

## Checklist
- [ ] The documentation is up to date with these changes. <- Dont know what documentation I would want to change to describe these changes.


docker-compose.yaml
```diff
@@ -266,32 +267,39 @@ services:
...
-#      - "19530:19530"
-#      - "9091:9091"
-#    depends_on:
-#      - "etcd"
-#      - "minio"
+  # milvus:
+  #   # Turn on to leverage the `vdb_upload` task
+  #   restart: always
+  #   container_name: milvus-standalone
+  #   image: milvusdb/milvus:v2.4.9-gpu
```
jperez999, line 266: This is the new version of milvus that we would need to use for GPU indexing to work.

docker-compose.yaml
```diff
@@ -266,32 +267,39 @@ services:
...
+  #   healthcheck:
+  #     test: ["CMD", "curl", "-f", "http://localhost:9091/healthz"]
+  #     interval: 30s
+  #     start_period: 90s
+  #     timeout: 20s
+  #     retries: 3
+  #   ports:
+  #     - "19530:19530"
+  #     - "9091:9091"
+  #   deploy:
```
jperez999, line 285: Added GPU access to the container for functionality support.

src/nv_ingest/schemas/vdb_task_sink_schema.py
```diff
@@ -37,10 +37,11 @@ def build_default_milvus_config(embedding_size: int = 1024) -> typing.Dict[str,
         "index_conf": {
             "field_name": "vector",
             "metric_type": "L2",
-            "index_type": "HNSW",
+            "index_type": "GPU_CAGRA",
```
jperez999, line 40: I am thinking it is ok to expect GPU support by default, given that the milvus image is also GPU specific. Let me know if we need to be more malleable in this regard.

randerzander (approved):
Thanks, @jperez999 !

I'm eager to play w/ this in the context of forthcoming bulk indexing

jdye64 (approved):
LGTM

mpenn:
@jdye64 can you give this a review/merge?

PR #175, opened 2024-10-17
Install curl for healthcheck

edknv:
## Description
Installs curl for [healthcheck](https://github.com/NVIDIA/nv-ingest/blob/main/docker-compose.yaml#L161).

```
$ docker ps
CONTAINER ID   IMAGE                                                                      COMMAND                  CREATED          STATUS                      PORTS                                                                                                                                                                                                                                                                                NAMES
d0d78f932ffe   nvcr.io/ohlfw0olaadg/ea-participants/nv-ingest:24.08                       "/opt/conda/envs/nv_…"   38 seconds ago   Up 38 seconds (healthy)     0.0.0.0:7670->7670/tcp, :::7670->7670/tcp                                                                                                                                                                                                                                            nv-ingest-nv-ingest-ms-runtime-1
```

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,6 +18,7 @@ RUN apt-get update && apt-get install -y \
       wget \
       bzip2 \
       ca-certificates \
+      curl \
     && apt-get clean
 
 # Install miniconda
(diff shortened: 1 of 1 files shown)


enwaiax:
For better user experience, would suggest use HEALTHCHECK in Dockerfile to check the healthy status like openzipkin/zipkin does.

https://github.com/openzipkin/zipkin/blob/master/docker/Dockerfile#L70

edknv:
> would suggest use HEALTHCHECK in Dockerfile

That's certainly an option, but I think we made the right choice with having it in docker-compose because the nv-ingest container is pretty useless as a standalone without other NIMs and having it in docker-compose makes it easier for users to make changes. @jdye64 Thoughts?

drobison00 (approved):

Issue #164, opened 2024-10-15
[FEA]: Add bulk ingest to vdb task


jperez999:
### Is this a new feature, an improvement, or a change to existing functionality?

New Feature

### How would you describe the priority of this feature request

Currently preventing usage

### Please provide a clear description of problem this feature solves

 This feature will allow for bulk ingest to milvus using the vdb task when adding a large corpus of data to an index. This will decrease the time it takes to ingest large amounts of data into an index, as compared to the currently used streaming process. This will also require a minio sink addition where we save the embeddings to minio to be uploaded during bulk ingest. 

### Describe the feature, and optionally a solution or implementation and any alternatives

We will add a boolean to the VDBTask that will activate the use of bulk ingest if the users want to deploy using this strategy. Otherwise the current ingestion strategy will continue to be used.

i.e. 
VDBTask(bulk=True)

PR #179, opened 2024-10-21
Minio Embedding Upload

jperez999:
## Description
This PR creates a task that allows the user to save embeddings into a minio bucket. This can be used in conjunction with the upcoming feature of bulk ingest with milvus. 
This does not close but supports #164 

## Checklist


client/src/nv_ingest_client/primitives/tasks/store.py
```diff
@@ -49,7 +64,8 @@ def __init__(
         structured: bool = True,
         images: bool = False,
         store_method: _Type_Store_Method = None,
-        **extra_params,
+        extra_params: dict = None,
```
jperez999, line 67: this change is critical, unknown keyword arguments are not passed to the task when created via the client. So when you try to use the minio credentials to access a private bucket by getting them from the params dictionary you end up not finding anything so you default to `none`. This change makes extra_params an explicit argument and therefore carries over when the task is recreated via the service. It also leaves params available because that is used by the CLI tool to blanket pass information to tasks.

client/src/nv_ingest_client/primitives/tasks/store.py
```diff
@@ -72,17 +89,76 @@ def __str__(self) -> str:
...
+
+    _Type_Content_Type = Literal["embedding",]
+
+    _Type_Store_Method = Literal["minio",]
+
+    def __init__(
+        self,
+        embedding: bool = True,
+        store_method: _Type_Store_Method = None,
+        extra_params: dict = None,
```
jperez999, line 125: See above comment.

requirements.txt
```diff
@@ -33,3 +33,4 @@ fastapi==0.109.1
 uvicorn==0.24.0-post.1
 gunicorn==22.0.0
 python-multipart
+s3fs
```
jperez999, line 36: required for pandas to be able to directly write the file to minio storage.

src/nv_ingest/modules/storages/embedding_storage.py
```diff
@@ -0,0 +1,161 @@
...
+
+EmbeddingStorageLoaderFactory = ModuleLoaderFactory(MODULE_NAME, MODULE_NAMESPACE, EmbeddingStorageModuleSchema)
+
+def upload_embeddings(df: pd.DataFrame, params: Dict[str, Any]) -> pd.DataFrame:
+    """
+    Identify contents (e.g., images) within a dataframe and uploads the data to MinIO.
+    The image metadata in the metadata column is updated with the URL of the uploaded data.
+    """
+    
+    access_key = params.get("access_key", None)
```
jperez999, line 51: without the previous extra_params change this check was coming back empty because these key/value pairs were not getting propagated.

client/src/nv_ingest_client/primitives/tasks/store.py
jperez999: Also we can use keyword arguments directly because when we go to instantiate using the Ingest schema it complains that were are passing keyword arguments that are not expressly defined in the schema because we explicitly forbid that in the ingest schemas. Therefore, I decided to keep that standard in place and use the extra_params dictionary to capture all necessary connection information, this coincides with how other libraries also handle connection parameters (i.e. pandas, s3fs).

src/nv_ingest/modules/storages/embedding_storage.py
```diff
@@ -0,0 +1,161 @@
...
+        "key":access_key,
+        "secret": secret_key,
+        "client_kwargs": {
+        "endpoint_url": _DEFAULT_READ_ADDRESS
+        }
+    }
+
+    meta = df["metadata"]
+    emb_df_list = []  
+    file_uuid = uuid.uuid4().hex
```
jperez999, line 85: I opted to use a UUID here because an individual file could correspond to multiple objects from different files if batch is used.

client/src/nv_ingest_client/primitives/tasks/store.py
```diff
@@ -72,17 +89,76 @@ def __str__(self) -> str:
...
             "images": self._images,
-            "params": self._extra_params,
+            "extra_params": self._extra_params,
+            **self._params,
         }
 
         return {"type": "store", "task_properties": task_properties}
+
+
+class StoreEmbedTask(Task):
```
jperez999, line 112: I added this here, because the behavior is most similar to the upload images, where we are writing bytes to a bucket. If it should go somewhere else or as something else, let me know and we can adjust accordingly.

src/nv_ingest/schemas/ingest_job_schema.py
```diff
@@ -43,6 +43,7 @@ class TaskTypeEnum(str, Enum):
     filter = "filter"
     split = "split"
     store = "store"
```
jperez999, line 45: in an effort to make the minimal number of changes, I opted to leave the store_images behavior under the `store` tag. Although now that multiple store actions exist, that is probably something we might want to update to more accurate depict the action completed by a task.

tests/nv_ingest_client/primitives/tasks/test_store.py
```diff
@@ -29,7 +32,14 @@ def test_store_task_initialization():
 
 
 def test_store_task_str_representation():
-    task = StoreTask(structured=True, images=True, store_method="minio", endpoint="minio:9000")
+    task = StoreTask(
+        structured=True, 
+        images=True, 
+        store_method="minio", 
+        extra_params={
```
jperez999, line 39: these test edits are related to the previously `extra_params` changes in this review.

client/client_examples/examples/python_client_usage.ipynb
```diff
@@ -27,14 +27,14 @@
...
-   "id": "daf6a66e-1e8f-4d7c-ac70-91b2c79a9951",
+   "id": "e7727953",
    "metadata": {},
    "outputs": [],
    "source": [
     "import os\n",
     "\n",
     "# client config\n",
-    "HTTP_HOST = os.environ.get('HTTP_HOST', \"localhost\")\n",
+    "HTTP_HOST = os.environ.get('HTTP_HOST', \"nv-ingest-ms-runtime\")\n",
```
edknv: Should we switch this back?

client/client_examples/examples/python_client_usage.ipynb
```diff
@@ -46,12 +46,12 @@
     "DEFAULT_JOB_TIMEOUT = 90\n",
     "\n",
     "# sample input file and output directory\n",
-    "SAMPLE_PDF = \"/workspace/nv-ingest/data/multimodal_test.pdf\""
+    "SAMPLE_PDF = \"/raid/workspace/nv-ingest/data/multimodal_test.pdf\""
```
edknv: Should we switch this back?

client/src/nv_ingest_client/cli/util/processing.py
```diff
@@ -522,6 +522,9 @@ def create_job_specs_for_batch(files_batch: List[str], tasks: Dict[str, Any], cl
         if "store" in tasks:
             job_spec.add_task(tasks["store"])
 
+        if "store_embedding" in tasks:
```
edknv: I think there will be merge conflict here when you merge main. If `store_embedding`  was added in `TaskTypeEnum`, we can simply get rid of this and not worry about adding this in the client.

src/nv_ingest/schemas/ingest_job_schema.py
edknv: Sounds like a good idea to rename this.

client/src/nv_ingest_client/cli/util/processing.py
jperez999: I merged in the changes, so no more conflicts.

src/nv_ingest/schemas/ingest_job_schema.py
jperez999: maybe we can do that in a subsequent PR so that this does not get more bloated?

edknv: Yes.

PR #190, opened 2024-10-24
Add milvus bulk

### LINKED ISSUE #164: [FEA]: Add bulk ingest to vdb task

jperez999:
## Description
This PR adds in bulk ingest functionality, that can be used in a chained set of tasks or as a solo task after the fact. There were some changes made to the schema, because JSON dtypes are not compatible with parquet files. With #179 this closes #164. A large portion of the code reflected here is actually from #179 because it builds and requires that implementation.

## Checklist


src/nv_ingest/modules/storages/embedding_storage.py
```diff
@@ -0,0 +1,159 @@
...
+                "uploaded_embedding_url"
+            ] = write_path
+        # TODO: validate metadata before putting it back in.
+        # cm = str(metadata["content_metadata"])
+        text = metadata["content"]
+        # source_meta = str(metadata["source_metadata"])
+        emb = metadata["embedding"]
+        df.at[idx, "metadata"] = metadata
+
+        emb_df_list.append([emb, text])
```
jperez999: the portions of the record that are saved to this parquet are dependent on the schema of the collection in Milvus. We either need to make that schema static (as it is now) or we need to be able to consume that in this task when creating it.

src/nv_ingest/modules/sinks/vdb_task_sink.py
```diff
@@ -34,6 +39,61 @@
...
+        access_key=access_key,
+        secret_key=secret_key,
+        session_token=extra_params.get("session_token", None),
+        secure=extra_params.get("secure", False),
+        region=extra_params.get("region", None),
+    )
+    bucket_found = client.bucket_exists(bucket_name)
+    if not bucket_found:
+        raise ValueError(f"Could not find bucket {bucket_name}")
+    batch_files = [[f"{file.object_name}"] for file in client.list_objects(bucket_name, prefix=bulk_ingest_path)]
```
jperez999: given a specific path in the minio bucket, we will ingest all the files in that bucket path.

src/nv_ingest/schemas/vdb_task_sink_schema.py
```diff
@@ -63,16 +63,16 @@ def build_default_milvus_config(embedding_size: int = 1024) -> typing.Dict[str,
...
-                    name="source",
-                    dtype=pymilvus.DataType.JSON,
-                    description="Source document and raw data extracted content",
-                ).to_dict(),
-                pymilvus.FieldSchema(
-                    name="content_metadata",
-                    dtype=pymilvus.DataType.JSON,
-                    description="Content metadata",
-                ).to_dict(),
+                # pymilvus.FieldSchema(
```
jperez999: removed these fields because they are JSON is not compatible with parquet files. If we were to keep this data, when we go to upload, the ingestion blows up because milvus expect json data but actually gets structured data from arrow.

drobison00:
@mpenn If you have time could you take a look at this.

client/src/nv_ingest_client/cli/util/click.py
```diff
@@ -18,15 +18,15 @@
 from nv_ingest_client.primitives.tasks import ExtractTask
 from nv_ingest_client.primitives.tasks import FilterTask
 from nv_ingest_client.primitives.tasks import SplitTask
-from nv_ingest_client.primitives.tasks import StoreTask
+from nv_ingest_client.primitives.tasks import StoreTask, StoreEmbedTask
```
drobison00:
```suggestion
from nv_ingest_client.primitives.tasks import StoreTask
from nv_ingest_client.primitives.tasks import StoreEmbedTask
```

client/src/nv_ingest_client/cli/util/click.py
```diff
@@ -18,15 +18,15 @@
...
 from nv_ingest_client.primitives.tasks import VdbUploadTask
 from nv_ingest_client.primitives.tasks.caption import CaptionTaskSchema
 from nv_ingest_client.primitives.tasks.dedup import DedupTaskSchema
 from nv_ingest_client.primitives.tasks.embed import EmbedTaskSchema
 from nv_ingest_client.primitives.tasks.extract import ExtractTaskSchema
 from nv_ingest_client.primitives.tasks.filter import FilterTaskSchema
 from nv_ingest_client.primitives.tasks.split import SplitTaskSchema
-from nv_ingest_client.primitives.tasks.store import StoreTaskSchema
+from nv_ingest_client.primitives.tasks.store import StoreTaskSchema, StoreEmbedTaskSchema
```
drobison00:
```suggestion
from nv_ingest_client.primitives.tasks.store import StoreTaskSchema
from nv_ingest_client.primitives.tasks.store import StoreEmbedTaskSchema
```

client/src/nv_ingest_client/cli/util/click.py
```diff
@@ -113,6 +113,10 @@ def click_validate_task(ctx, param, value):
                 task_options = check_schema(StoreTaskSchema, options, task_id, json_options)
                 new_task_id = f"{task_id}"
                 new_task = StoreTask(**task_options.dict())
+            elif task_id == "store_embedding":
```
drobison00, line 131: Could you elaborate a bit on the differences between `store_embedding` and `embed` + `vdb_upload`

client/src/nv_ingest_client/primitives/tasks/__init__.py
```diff
@@ -8,7 +8,7 @@
 from .extract import ExtractTask
 from .filter import FilterTask
 from .split import SplitTask
-from .store import StoreTask
+from .store import StoreTask, StoreEmbedTask
```
drobison00:
```suggestion
from .store import StoreTask
from .store import StoreEmbedTask
```

client/src/nv_ingest_client/cli/util/click.py
drobison00: I want to be clear on whether or not adding a task vs adding configuration to something like vdb_upload is the right approach here.

client/src/nv_ingest_client/primitives/tasks/task_factory.py
```diff
@@ -14,7 +14,7 @@
 from .extract import ExtractTask
 from .filter import FilterTask
 from .split import SplitTask
-from .store import StoreTask
+from .store import StoreTask, StoreEmbedTask
```
drobison00:
```suggestion
from .store import StoreTask
from .store import StoreEmbedTask
```

src/nv_ingest/modules/sinks/vdb_task_sink.py
```diff
@@ -26,6 +28,9 @@
 from nv_ingest.util.flow_control import filter_by_task
 from nv_ingest.util.modules.config_validator import fetch_and_validate_module_config
 from nv_ingest.util.tracing import traceable
+from pymilvus import BulkInsertState, Collection, connections, utility
```
drobison00, line 31:
```suggestion
from pymilvus import BulkInsertState
from pymilvus import Collection
from pymilvus import connections
from pymilvus import utility
```

src/nv_ingest/modules/sinks/vdb_task_sink.py
```diff
@@ -34,6 +39,61 @@
 
 VDBTaskSinkLoaderFactory = ModuleLoaderFactory(MODULE_NAME, MODULE_NAMESPACE, VdbTaskSinkSchema)
 
+_DEFAULT_ENDPOINT = os.environ.get("MINIO_INTERNAL_ADDRESS", "minio:9000")
```
drobison00, line 42: I know we may not be fully consistent on this front yet, but I would like to avoid having environment values pulled in at the stage level, because it has the potential to obfuscate where the configuration is obtained. 

Please put default detection into `pipeline.py`, or going forward, `src/util/pipeline/stage_builders.py`
See: https://github.com/NVIDIA/nv-ingest/blob/main/src/pipeline.py#L59

src/nv_ingest/modules/sinks/vdb_task_sink.py
```diff
@@ -34,6 +39,61 @@
...
+                collection_name: str = None, 
+                bucket_name: str = None, 
+                bulk_ingest_path: str = None, 
+                extra_params: dict = None):
+
+    endpoint = extra_params.get("endpoint", _DEFAULT_ENDPOINT)
+    access_key = extra_params.get("access_key", None)
+    secret_key = extra_params.get("secret_key", None)
+
+    client = Minio(
```
drobison00, line 56: I know that Minio is the only supported option at the moment, but it is still a configuration parameter. Please add checking w.r.t which client is created based on the `store_method`

src/nv_ingest/modules/sinks/vdb_task_sink.py
```diff
@@ -34,6 +39,61 @@
...
+    uri_parsed = urlparse(milvus_uri)
+    conn = connections.connect(host=uri_parsed.hostname, port=uri_parsed.port)
+    collection = Collection(name=collection_name)
+
+    task_ids = []
+    for file in batch_files:
+        task_id = utility.do_bulk_insert(collection_name=collection_name, files=file)
+        task_ids.append(task_id)
+
+    while len(task_ids) > 0:
```
drobison00, line 78: This has the potential to get into an infinite loop if something goes wrong. We will want to add some kind of maximum threshold before the task is marked as failed.

src/nv_ingest/modules/sinks/vdb_task_sink.py
```diff
@@ -34,6 +39,61 @@
...
+        for id in task_ids:
+            state = utility.get_bulk_insert_state(task_id=id)
+            if state.state == BulkInsertState.ImportFailed or state.state == BulkInsertState.ImportFailedAndCleaned:
+                logger.error(f"The task {state.task_id} failed, reason: {state.failed_reason}")
+                task_ids.remove(id)
+            elif state.state == BulkInsertState.ImportCompleted:
+                logger.info(f"The task {state.task_id} completed")
+                task_ids.remove(id)
+
+    while True:
```
drobison00, line 90: Same issue here, we will want to make sure this cannot go on indefinitely; if we go more than X seconds without progress or Y overall seconds the job should be failed.

src/nv_ingest/modules/storages/embedding_storage.py
```diff
@@ -0,0 +1,159 @@
...
+MODULE_NAMESPACE = "nv_ingest"
+
+# TODO: Move these into pipeline.py to populate the stage and validate them using the pydantic schema on startup.
+_DEFAULT_ENDPOINT = os.environ.get("MINIO_INTERNAL_ADDRESS", "minio:9000")
+_DEFAULT_READ_ADDRESS = os.environ.get("MINIO_PUBLIC_ADDRESS", "http://minio:9000")
+_DEFAULT_BUCKET_NAME = os.environ.get("MINIO_BUCKET", "nv-ingest")
+
+EmbeddingStorageLoaderFactory = ModuleLoaderFactory(MODULE_NAME, MODULE_NAMESPACE, EmbeddingStorageModuleSchema)
+
+def upload_embeddings(df: pd.DataFrame, params: Dict[str, Any]) -> pd.DataFrame:
```
drobison00: We will definitely want to turn this into a multiprocessing stage. Anything doing uploads is likely to slow down the pipeline significantly.

src/nv_ingest/schemas/vdb_task_sink_schema.py
drobison00: @mpenn I'm not familiar enough with the lineage here, could you comment?

drobison00 (changes requested):
Some comments/concerns lets discuss.

client/src/nv_ingest_client/cli/util/click.py
jperez999: Store embedding is the action of saving an embedding or a set of embeddings to minio storage. Where vdb_upload (bulk ingestion) is so that all the written files can be ingested at once by milvus. While embed + vdb_upload (we can call it vanilla) is for streaming purposes, meaning that one embedding comes in and that same embedding is loaded in the milvus as they come through the pipe.

src/nv_ingest/modules/sinks/vdb_task_sink.py
jperez999: That is going to be difficult because the amount of time an index can take to complete creation is entirely dependent on the number of entries ingested. I can add a count for the timeframe between updates but other than that we wont know how much time this will actually take. It will be very different for 1k records and 1 million records or 1 billion records.  Depending on the number of records this can take hours, and it might seem long but still orders of magnitude faster than CPU based upload.

src/nv_ingest/modules/storages/embedding_storage.py
jperez999: Can you link an example to what you mean? Like async? Or do you mean spawn multiple processes to do this. Is there a maximum amount of data a single control message can hold ( I thought it would be limited by network packet limitations)? (still learning about capabilities here)

src/nv_ingest/modules/sinks/vdb_task_sink.py
jperez999: If something goes wrong with a task here, it is removed just the same as if it completed successfully. This can take a undeterminable amount of time because it is dependent on the amount of data stored in the minio embedding directory that needs to be ingested.

src/nv_ingest/schemas/vdb_task_sink_schema.py
mpenn: The lineage for this was to provide content in an intuitive structure to support the RAG workflow behind the [multimodal-pdf-data extraction blueprint](https://build.nvidia.com/nvidia/multimodal-pdf-data-extraction-for-enterprise-rag/blueprintcard). If we decide to move from this structure, it will likely require some blueprint re-work. Not impossible but would require some re-work.

client/src/nv_ingest_client/primitives/tasks/store.py
```diff
@@ -20,6 +20,12 @@
 _DEFAULT_STORE_METHOD = "minio"
 
 
+class StoreEmbedTaskSchema(BaseModel):
```
jperez999, line 23: This is here to have a place to add arguments if they become necessary.

client/src/nv_ingest_client/primitives/tasks/store.py
```diff
@@ -72,17 +80,66 @@ def __str__(self) -> str:
...
         """
         Convert to a dict for submission to redis (fixme)
         """
+
         task_properties = {
             "method": self._store_method,
             "structured": self._structured,
             "images": self._images,
-            "params": self._extra_params,
+            "extra_params": self._extra_params,
```
jperez999: This fixes the params issue where you cant connect to the private bucket.

src/nv_ingest/modules/sinks/vdb_task_sink.py
jperez999: I think we solved this, as leaving it in place, because Milvus only works with minio. And milvus is the only Vector Database that supports GPU indexes right now.

src/nv_ingest/modules/sinks/vdb_task_sink.py
jperez999: There was never a store method for the VDB upload task.

src/nv_ingest/modules/sinks/vdb_task_sink.py
```diff
@@ -328,6 +394,7 @@ def on_data(ctrl_msg: ControlMessage):
     node = builder.make_node(
         WRITE_TO_VECTOR_DB, ops.map(on_data), ops.filter(lambda val: val is not None), ops.on_completed(on_completed)
     )
+    node.launch_options.engines_per_pe = validated_config.progress_engines
```
jperez999, line 397: @mpenn helped me out with adding progress engines here.

src/nv_ingest/modules/storages/image_storage.py
```diff
@@ -135,7 +135,7 @@ def on_data(ctrl_msg: ControlMessage):
             if store_images:
                 content_types[ContentTypeEnum.IMAGE] = store_images
 
-            params = task_props.get("params", {})
+            params = task_props.get("extra_params", {})
```
jperez999: part of the fix for the params not coming all the way through from the client.

src/nv_ingest/schemas/vdb_task_sink_schema.py
jperez999: Still need further guidance on how we want to handle this schema change for VDB entries.

src/nv_ingest/stages/storages/embedding_storage_stage.py
```diff
@@ -0,0 +1,148 @@
...
+
+    Returns
+    -------
+    MultiProcessingBaseStage
+        A Morpheus stage with applied worker function.
+    """
+    validated_config = EmbeddingStorageModuleSchema()
+    _wrapped_process_fn = functools.partial(_store_embeddings, validated_config=validated_config)
+
+    return MultiProcessingBaseStage(
```
jperez999, line 158: Here is the multi processing conversion for the embedding storage task. @drobison00

src/nv_ingest/modules/sinks/vdb_task_sink.py
drobison00: That's fine, the issue is really that I want to avoid gathering environment variables in stages vs having them passed in to the stage as configuration parameters that can be validated.

src/nv_ingest/schemas/vdb_task_sink_schema.py
drobison00: Waiting on a decision from @randerzander on this; would definitely prefer a work around if it is feasible, as this is almost certainly a breaking change for some workflows.

drobison00 (approved):

PR #195, opened 2024-10-25
Add EA notes for API key generation.

reliseinv:
The API Generation steps are general but the developer needs keys with specific permission to pull docker resources for this repo. This PR adds detail for which team / org to use when creating the API key.


jdye64 (approved):
Awesome. Thanks @reliseinv !

PR #197, opened 2024-10-28
Introduce an improved client library API with chainable verb methods

### LINKED ISSUE #186: [FEA]: Improve Client Library API

Improvement

Significant improvement

Currently, the library uses something akin to a builder pattern for creating a JobSpec. While is is perfectly functional, its not as intuitive as it could be. To improve the overall experience, this feature request would add a new `JobOrchestration` class (or similar) that supports chainable verb methods (see below), and orchestrates JobSpec creation and submission.

_Current process_

```python
job_spec = JobSpec(
    document_type=file_type,
    payload=file_content,
    source_id=SAMPLE_PDF,
    source_name=SAMPLE_PDF,
    extended_options={
        "tracing_options": {
            "trace": True,
            "ts_send": time.time_ns(),
        }
    },
)

extract_task = ExtractTask(
    document_type=file_type,
    extract_text=True,
    extract_images=True,
    extract_tables=True,
    text_depth="document",
    extract_tables_method="yolox",
)

dedup_task = DedupTask(
    content_type="image",
    filter=True,
)

filter_task = FilterTask(
    content_type="image",
    min_size=128,
    max_aspect_ratio=5.0,
    min_aspect_ratio=0.2,
    filter=True,
)

job_spec.add_task(extract_task)
job_spec.add_task(dedup_task)
job_spec.add_task(filter_task)
```

_With updated API_
```python
client = NvIngestClient(
    message_client_allocator=RestClient,
    message_client_hostname=HTTP_HOST,
    message_client_port=HTTP_PORT,
    message_client_kwargs=None,
    msg_counter_id="nv-ingest-message-id",
    worker_pool_size=1,
)

job_orchestrator = IngestJobOrchestrator([documents_0, ...], client=client) # attempt to infer file type information when possible

job_orchestrator.extract(**kwargs).dedup(**kwargs).filter(**kwargs).trace() # Some complexity here if certain options aren't supported by a given document type, should be best effort to coerce into a sensible configuration.

job_orchestrator.run() # blocking submission of all jobs

futures = job_orchestrator.run_async() # Return a single future that indicates when all jobs have completed.

job_orchestrator.running_jobs() # how many jobs are currently in flight
job_orchestrator.remaining_jobs() # how many jobs are still unfinished
job_orchestrator.failed_jobs() # how many jobs have failed

...
```

### Additional context

Requirements:

- [ ] High level Job orchestration object that
  - [ ] Simplifies job creation from a file or files
  - [ ] Supports function chaining via verb's which correspond to
  - [ ] Supports running the defined job against the supplied client
  - [ ] Supports checking the current status of a running job set

edknv:
## Description
Closes #186.

```python
from nv_ingest_client.client import Ingestor

ingestor = (
    Ingestor(message_client_hostname="nv-ingest-ms-runtime")
    .files("data/bo20/*.pdf")
    .extract(
        extract_text=True,
        extract_tables=True,
        extract_images=True,
    )
)

results = ingestor.run()

print(f"Got {len(results)} results")
```

## Checklist


client/src/nv_ingest_client/client/job_manager.py
```diff
@@ -0,0 +1,348 @@
...
+    documents : List[str]
+        List of document paths to be processed.
+    client : Optional[NvIngestClient], optional
+        An instance of NvIngestClient. If not provided, a client is created.
+    job_queue_id : str, optional
+        The ID of the job queue for job submission, default is "morpheus_task_queue".
+    """
+
+    def __init__(
+        self,
```
drobison00: In the interest of easy of use, lets add the ability for the Manager to create a client in addition to taking one in.

I know you already do it in _create_client, but we can also take in the same constructor parameters.

client/src/nv_ingest_client/client/job_manager.py
```diff
@@ -0,0 +1,348 @@
...
+        documents: List[str],
+        client: Optional[NvIngestClient] = None,
+        job_queue_id: str = DEFAULT_JOB_QUEUE_ID,
+    ):
+        self._documents = documents
```
drobison00: I think we're going to want to pull in a library like fsspec, so we can handle remote paths, like s3://...., etc...

client/src/nv_ingest_client/client/job_manager.py
```diff
@@ -0,0 +1,348 @@
...
+            completed_futures.add(future)
+            future_results.append(result)
+            if completed_futures == submitted_futures:
+                combined_future.set_result(future_results)
+
+        for future in future_to_job_id:
+            future.add_done_callback(_done_callback)
+
+        return combined_future
+
```
drobison00: Thinking more about this, we're going to want an interface verb for something like `load()`, that checks to see if all the files specified exist locally, or need to be retrieved from some external source. It should be optional, because there will be some use cases where we want to let the ingest service itself do the load. 

We also are going to want a method like `files([....])`, that lets someone create the job_manager without selecting the file set. This will make the class re-usable, and a bit more flexible.

```python

job_manager.files(my_files)
result = job_managet.load().extract().run()

File: client/src/nv_ingest_client/client/job_manager.py [hunk 4]
```diff
@@ -0,0 +1,348 @@
...
+        ----------
+        kwargs : dict
+            Parameters specific to the ExtractTask.
+
+        Returns
+        -------
+        NvIngestJobManager
+            Returns self for chaining.
+        """
+        for document_type in self._job_specs.file_types:
```
drobison00: Please update this to add `table_data_extract` and `chart_data_extract` tasks if extract is specified with `extract_tables` and/or `extract_charts`.

drobison00: See: https://github.com/NVIDIA/nv-ingest/blob/main/client/src/nv_ingest_client/cli/util/click.py#L112

client/src/nv_ingest_client/client/__init__.py
```diff
@@ -2,6 +2,7 @@
 # All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-from .client import NvIngestClient
+from nv_ingest_client.client.client import NvIngestClient
+from nv_ingest_client.client.job_manager import NvIngestJobManager
 
-__all__ = ["NvIngestClient"]
+__all__ = ["NvIngestClient", "NvIngestJobManager"]
```
drobison00: I'm not sure I'm on board with NvIngestJobManager, because I don't think it entirely captures what the intention of this class is.

We almost want to have it be `NvIngestClient`, but `NvIngestInterface` or something similar might work if we don't want to make breaking changes.

@randerzander

drobison00 (changes requested):

client/src/nv_ingest_client/client/__init__.py
edknv: I initially had `NvInestPipeline` but later changed it to `NvIngestJobManager`. If this class is supposed to be *the* entrypoint, then another option is just `NvIngest`.

drobison00: Since we have `nv_ingest` and `nv_ingest_client`, having a class called `NvIngest` in the `nv_ingest_client` library feels confusing.

edknv: Ah, good point. They weren't kidding with naming things being one of the two hardest problems in computer science :) I'm also okay with `NvIngestInterface`.

randerzander: pipeline was more intuitive to me than jobmanager

but at the end of the day, we're ingesting (or extracting, in the stop at json results case). What about Ingestor?

edknv: Going with `Ingestor`. https://github.com/NVIDIA/nv-ingest/pull/197/commits/2995be3b652ad602f884cb423e372ee3f96f7a8e

client/src/nv_ingest_client/client/job_manager.py
edknv: Added the tasks in https://github.com/NVIDIA/nv-ingest/pull/197/commits/5e14aabc65a819cef29fa8e1f985e78a7a594cf8.

client/src/nv_ingest_client/client/job_manager.py
edknv: Done in https://github.com/NVIDIA/nv-ingest/pull/197/commits/d12287ab9d7415ca39bd0314985cacccebcc4c79

drobison00 (approved):

PR #215, opened 2024-11-05
Add the ability to the CLI to post process returned metadata and extract image to a local media file

### LINKED ISSUE #214: [FEA]: Add the ability for the CLI and client library to extract image content from metadata to disk and replace with URL in metadata file

New Feature

Would be nice

This is a quality of life update that can help with making metadata files more human readable when required.

Requires modifying the CLI and client library so that it supports an option to extract images and retain a link to them on disk.

### Additional context

_No response_

drobison00:
Resolves #214 

## Checklist

### DIFF
--- a/client/src/nv_ingest_client/cli/util/processing.py
+++ b/client/src/nv_ingest_client/cli/util/processing.py
@@ -2,12 +2,18 @@
 # All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+import base64
+import io
+import json
 import logging
+import os
 import re
 import time
 import traceback
+
 from collections import defaultdict
 from concurrent.futures import as_completed
+from PIL import Image
 from statistics import mean
 from statistics import median
 from typing import Any
@@ -15,11 +21,6 @@ from typing import Dict
 from typing import List
 from typing import Tuple
 from typing import Type
-import os
-import json
-import base64
-from PIL import Image
-import io
 
 from click import style
 from pydantic import BaseModel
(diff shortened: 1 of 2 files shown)


jdye64:
ok to test

client/src/nv_ingest_client/cli/util/processing.py
```diff
@@ -417,22 +395,56 @@ def save_response_data(response, output_directory):
...
+                        image_output_path = os.path.join(doc_type_path, "media", image_filename)
+
+                        # Ensure the media directory exists
+                        os.makedirs(os.path.dirname(image_output_path), exist_ok=True)
+
+                        # Save the image to disk
+                        image.save(image_output_path, format=image_ext.upper())
+
+                        # Update the metadata content with the image path
+                        meta["content"] = os.path.realpath(image_output_path)
```
jdye64: This is the only thing I question. I do like we make it very clear the image has been decoded and saved elsewhere but then wonder if we want to just leave the base64 encoded bytes here and add another metadata field for the locaiton the file was saved?

Honestly, I can't decide on which way so I'm fine with this.

jdye64 (approved):
Had a single comment but LGTM

client/src/nv_ingest_client/cli/util/processing.py
drobison00: One of the comments here is that the base64 encoded bytes make the file bad for human readability because it exceeds normal buffer sizes and doesn't scroll nicely.

zredeaux07:
Updates and typo fix look good!

PR #141, opened 2024-10-03
Use NVIDIA endpoints in example notebooks and demonstrate using milvus as a vector database

ChrisJar:
## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/nv-ingest/blob/main/CONTRIBUTING.md).
- [ ] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

## TODO:
- [x] Demonstrate add example vdbupload and store tasks (still working out how to hook this into llamaindex and langchain)
- [x] Demonstrate storing with minio and fetching and displaying some of the images

closes #122 
closes #124

PR #222, opened 2024-11-12
Update langchain_multimodal_rag.ipynb to support recept API updates

drobison00:
Small update set that brings the langchain_multimodal_rag example into line with recent API updates.

## Checklist


ChrisJar (approved):
#141 consolidates the embedding and upload to milvus into the NV-Ingest job rather than doing it with Langchain, so that'll need to be updated when #141 goes in, but other than that LGTM!

jdye64:
Completed as part of other PRs. No longer needed.

PR #230, opened 2024-11-14
Add Nemoretriever-parse (aka Eclair) NIM

edknv:
## Description
Part of https://github.com/NVIDIA/nv-ingest-private/issues/52
Also closes https://github.com/NVIDIA/nv-ingest-private/issues/49.

The NIM has its own docker-compose profile `nemoretriever-parse` so you have to specify `--profile` to start the NIM service, e.g.,
```
docker compose --profile nemoretriever-parse up --build
```

To submit a job via nv-ingest-client, specify `extract_method`.

CLI:
```
--task 'extract:{..., "extract_method": "nemoretriever_parse", ...}
```
Ingestor:
```
    .extract(
        ...
        extract_method="nemoretriever_parse",
        ...
    )
```

## Checklist


src/nv_ingest/stages/nim/chart_extraction.py
```diff
@@ -62,7 +64,8 @@ def _update_metadata(row: pd.Series, cached_client: Any, deplot_client: Any, tra
     # Only modify if content type is structured and subtype is 'chart' and chart_metadata exists
     if ((content_metadata.get("type") != "structured") or
             (content_metadata.get("subtype") != "chart") or
-            (chart_metadata is None)):
+            (chart_metadata is None) or
+            (chart_metadata.get("table_format") != TableFormatEnum.IMAGE)):
```
drobison00: I don't think this is correct. Even if the table_format is image, we can still extract the content in the chart extractor. Am I thinking about this wrong?

src/nv_ingest/stages/nim/table_extraction.py
```diff
@@ -63,7 +67,8 @@ def _update_metadata(row: pd.Series, paddle_client: Any, paddle_version: Any, tr
     # Only modify if content type is structured and subtype is 'table' and table_metadata exists
     if ((content_metadata.get("type") != "structured") or
             (content_metadata.get("subtype") != "table") or
-            (table_metadata is None)):
+            (table_metadata is None) or
+            (table_metadata.get("table_format") != TableFormatEnum.IMAGE)):
```
drobison00: Same question here.

src/nv_ingest/util/nim/helpers.py
```diff
@@ -107,29 +110,43 @@ def call_image_inference_model(client, model_name: str, image_data):
...
+        result = " ".join([output[0].decode("utf-8") for output in result.as_numpy("output")])
+    elif model_name == "doughnut":
+        result = [output.decode("utf-8") for output in result.as_numpy("output")]
+
+    return result
 
-def _call_image_inference_http_client(client, model_name: str, image_data):
-    base64_img = numpy_to_base64(image_data)
 
+def _call_image_inference_http_client(client, model_name: str, image_data: np.ndarray):
```
drobison00: I think we're getting a bit too specific in terms of data handling here. We probably want to standardize on a data format for each model and process it into that before we get to this function.

Small nitpick: please use base64_img(s) or base64_image(s), rather than base64_img and base64_images

drobison00 (changes requested):

src/nv_ingest/stages/nim/chart_extraction.py
edknv: I added this so that the tables extracted from the Doughnut model don't go through the table/chart extraction stages. Doughnut tables will already have text (as LaTex), so they don't need to go through the table/chart extraction stages. YOLOX tables need text extraction in table/chart extraction stages, and they are tagged as IMAGE tables so they do get processed in these stages. I'm not sure if that made sense, but I needed a way to skip table/chart extraction for tables identified by Doughnut, and thought TableFormat could be useful here to distinguish between yolox (== TableFormatEnum.IMAGE) and doughnut (==TableFormatEnum.LATEX).

src/nv_ingest/util/nim/helpers.py
edknv: I cleaned it up a bit in [c5b5d3e](https://github.com/NVIDIA/nv-ingest/pull/230/commits/c5b5d3e0f8b80a78d9fb68efca7f1cfeac834612) by basically grouping doughnut together with deplot and removing batching. Similarly to deplot, request-level batching is currently not supported in doughnut, and they have the same output format (for now). Note: The input/output format is not finalized and will likely change and actually be closer to the yolox output format so there will be more PRs to follow.

edknv:
[321052f](https://github.com/NVIDIA/nv-ingest/pull/230/commits/321052fd3d142ae2c80e63d16b069cea1cd3c9d0) adds support for preserving the text bounding boxes in the metadata (in the `hierarchy` field) and addresses issue https://github.com/NVIDIA/nv-ingest-private/issues/49. It also changes the way text blocks are concatenated to concatenate text blocks with `\n\n` as requested by research team.

PR #238, opened 2024-11-22
Add SimpleMessageBroker class to support a fully self contained pipeline.

### LINKED ISSUE #226: [FEA]: Add new SimpleMessageBroker and supporting elements to NV-ingest

New Feature

Currently preventing usage

**Overview:**
Currently, the ingest pipeline relies on a message broker (Redis by default) to feed data, requiring the deployment of a message broker container and front-end REST service. For testing or proof of concept scenarios, it would be beneficial to have a more streamlined option that eliminates these dependencies.

**Solution:**
Implement a simple inline message broker that can be used within the pipeline and create a corresponding client interface. This will allow the ingest service to run independently, without requiring local dependencies on an external message broker or REST service.

Introduce socket_task_source and socket_task_sink components to the nv_ingest service. These will be configurable to listen on a specified source, allowing jobs to be accepted and results returned over sockets. Additionally, update nv_ingest_client with options to submit jobs and fetch job results from the nv_ingest service via these socket connections.

### Additional context

_No response_

drobison00:
Resolves #226 

#### SimpleBroker and SimpleClient
- Introduced a `SimpleBroker` class that mimics Redis/Kafka but operates as a single-threaded process.
- Added a `SimpleClient` that adheres to the same `client_base` interface as `RedisClient`, making them interchangeable.

#### CLI Updates
- Enhanced CLI to support `--client_type=simple|rest`.
- **Note**: Code duplication for clients between `ingest` and `ingest-client` persists. A shared `ingest-common` module is recommended for future work.

#### SimpleBroker Features
- Lightly `NV-ingest` aware to facilitate `job_id` creation.
- Supports `push/pop` operations with a simple 3-way handshake protocol for transactional consistency:
  - **Failed Push**: Discards data.
  - **Failed Pop**: Re-inserts data, preserving original order.

#### ResponseSchema
- Updated both clients to return a standardized `ResponseSchema` object.
- Formalizes response structure, ensuring consistent behavior and transparency across client implementations.

#### NimClient and ModelInterface
- Added `NimClient` object and an abstract `ModelInterface` class to unify interaction with NIMs (HTTP/gRPC).
- Introduced specialized `ModelInterface` implementations for:
  - `deplot`
  - `cached`
  - `paddle`
  - `yolox`
  - Each handles specific pre/post-processing requirements.

#### Broker Abstraction
- Renamed `RedisTaskSource` and `RedisTaskSink` to `MessageBrokerTaskSource` and `MessageBrokerTaskSink`.
- Source and sink stages now accept a configurable `broker` parameter to dynamically construct the appropriate client object.

## Checklist


client/src/nv_ingest_client/message_clients/rest/rest_client.py
```diff
@@ -157,38 +160,42 @@ def fetch_message(self, job_id: str, timeout: float = 10) -> Optional[str]:
...
+                        response=result.text
+                    )
                 else:
                     # If the result contains a 200 then return the raw JSON string response
                     if response_code == 200:
-                        return result.text
+                        return ResponseSchema(
+                            response_code=0,
+                            response_reason="OK",
+                            response=result.text,
```
jdye64, line 188: I have a PR that was also returning the HTTP response header ['x-trace-id'] here as well. It isn't a problem but as a heads up I will likely modify this new `ResponseSchema` object to contain a `trace_id` attribute as well after this PR merges.

client/src/nv_ingest_client/schemas/response_schema.py
```diff
@@ -0,0 +1,19 @@
...
+
+from typing import Optional, Union
+from pydantic import BaseModel
+
+
+class ResponseSchema(BaseModel):
+    response_code: int
+    response_reason: Optional[str] = "OK"
+    response: Union[str, dict, None] = None
+    trace_id: Optional[str] = None  # Unique trace ID
```
jdye64, line 18: Ignore previous comment I see you already have the `trace_id` attribute here.

docker-compose.yaml
```diff
@@ -125,38 +126,59 @@ services:
...
     environment:
-      - CACHED_GRPC_ENDPOINT=cached:8001
-      - CACHED_HTTP_ENDPOINT=http://cached:8000/v1/infer
-      - CACHED_INFER_PROTOCOL=grpc
+      # Self-hosted cached endpoints.
+      #- CACHED_GRPC_ENDPOINT=cached:8001
+      #- CACHED_HTTP_ENDPOINT=http://cached:8000/v1/infer
+      #- CACHED_INFER_PROTOCOL=grpc
+      # build.nvidia.com hosted cached endpoints.
+      - CACHED_HTTP_ENDPOINT=https://ai.api.nvidia.com/v1/cv/university-at-buffalo/cached
```
jdye64: I wonder if there is a way to make this a configurable flag some local NIM deployment (as in the docker deployment itself) vs using the hosted endpoints. I can think of a few places where having to manually modify the docker-compose file to use local would be painful.

brev.dev notebook deployments for example

src/nv_ingest/schemas/message_broker_source_schema.py
```diff
@@ -6,11 +6,13 @@
...
+from nv_ingest.schemas.message_broker_client_schema import MessageBrokerClientSchema
 
 
-class RedisTaskSourceSchema(BaseModel):
-    redis_client: RedisClientSchema = RedisClientSchema()
+class MessageBrokerTaskSourceSchema(BaseModel):
+    broker_client: MessageBrokerClientSchema = MessageBrokerClientSchema()
+
     task_queue: str = "morpheus_task_queue"
+    raise_on_failure: bool = False
```
jdye64, line 16: Curious why we would never need to raise a failure?

src/nv_ingest/stages/nim/table_extraction.py
```diff
@@ -117,28 +121,51 @@ def _extract_table_data(df: pd.DataFrame, task_props: Dict[str, Any],
...
         logger.debug("No trace_info provided. Initialized empty trace_info dictionary.")
 
+    paddle_infer_protocol = stage_config.paddle_infer_protocol.lower()
+
+    if paddle_infer_protocol == 'grpc':
+        # Obtain paddle_version
+        # Assuming that the grpc endpoint is at index 0
+        paddle_endpoint = stage_config.paddle_endpoints[0]
+        try:
+            paddle_version = get_version(paddle_endpoint)
```
edknv: `get_version` takes an http endpoint so if the grpc endpoint endpoint is at index 0,  `paddle_endpoint = stage_config.paddle_endpoints[1]` (not `[0]`).

client/src/nv_ingest_client/nv_ingest_cli.py
```diff
@@ -67,6 +68,8 @@
 @click.option("--client_host", default="localhost", help="DNS name or URL for the endpoint.")
 @click.option("--client_port", default=6397, type=int, help="Port for the client endpoint.")
 @click.option("--client_kwargs", help="Additional arguments to pass to the client.", default="{}")
+@click.option("--client_type", default="rest", type=click.Choice(["rest", "simple"], case_sensitive=False),
+              help="Client type used to connect to the ingest service.")
```
edknv, line 72: I think I heard the client type on the client side should match the `MESSAGE_CLIENT_TYPE` on the server side. Is that correct? If so, should we change the default here to `simple` to match docker-compose.yaml, or vice versa?

docker-compose.yaml
```diff
@@ -125,38 +126,59 @@ services:
     volumes:
       - ${DATASET_ROOT:-./data}:/workspace/data
     ports:
+      # HTTP API
       - "7670:7670"
+      # Simple Broker API
+      - "7671:63790"
```
edknv: Is there a reason we want to avoid mapping 7671 to 7671? I had trouble with 63790 in my environment and had to not use 63790 and just use `MESSAGE_CLIENT_PORT=7671` to get the simple message broker to work.

src/nv_ingest/util/nim/yolox.py
```diff
@@ -2,15 +2,189 @@
...
+
+            return input_array
+
+        elif protocol == 'http':
+            logger.debug("Formatting input for HTTP Yolox model")
+            # Convert images to base64 and construct payload
+            images_base64 = []
+            for image in data['resized_images']:
+                # Convert numpy array to PIL Image
+                from PIL import Image
```
edknv: Can we move this to the top of the module?

src/nv_ingest/util/nim/paddle.py
```diff
@@ -0,0 +1,80 @@
...
+            if isinstance(response, np.ndarray):
+                if response.dtype.type is np.object_ or response.dtype.type is np.bytes_:
+                    output_strings = [
+                        output[0].decode('utf-8') if isinstance(output, bytes) else str(output) for output in response
+                    ]
+                    return " ".join(output_strings)
+                else:
+                    output_bytes = response.tobytes()
+                    output_str = output_bytes.decode('utf-8')
+                    return output_str
```
edknv, line 137: Why is this logic with data types necessary? It is because of the difference between NIM and Preview API?

docs/deployment.md
```diff
@@ -45,3 +45,18 @@ CONTAINER ID   IMAGE                                        COMMAND
...
+To run the nv-ingest service locally, we will require [Conda (Mamba) to be installed](https://mamba.readthedocs.io/en/latest/installation/mamba-installation.html).
+
+From the root of the repository, run the following commands to create a new Conda environment and install the required dependencies:
+```bash
+mamba env create --file ./docker/environments/nv_ingest_environment.yml --name nv_ingest_runtime
+
+pip install ./
+pip install ./client
+
+conda activate nv_ingest_runtime
```
jdye64: Maybe I'm misunderstanding this but shouldn't this activation be before the `pip install's`? That way the pip installs will go into the conda environment created?

client/src/nv_ingest_client/message_clients/simple/simple_client.py
```diff
@@ -0,0 +1,243 @@
...
+    def _handle_push(self, queue_name: str, message: str, timeout: Optional[float],
+                     for_nv_ingest: bool) -> ResponseSchema:
+        """Push a message to the queue, respecting the specified timeout."""
+        if (not queue_name or not isinstance(queue_name, str)):
+            return ResponseSchema(response_code=1, response_reason="Invalid queue name.")
+        if (not message or not isinstance(message, str)):
+            return ResponseSchema(response_code=1, response_reason="Invalid message.")
+
+        if (for_nv_ingest):
+            command = {"command": "PUSH_FOR_NV_INGEST", "queue_name": queue_name, "message": message}
```
jdye64, line 176: I think it would be good to have an ENUM somewhere that defines all of these `command` types instead of using hardcoded strings

client/src/nv_ingest_client/message_clients/simple/simple_client.py
```diff
@@ -0,0 +1,243 @@
...
+                with socket.create_connection((self._host, self._port), timeout=self._connection_timeout) as sock:
+                    self._send(sock, json.dumps(command).encode('utf-8'))
+                    # Receive initial response with transaction ID
+                    response_data = self._recv(sock)
+                    response = json.loads(response_data)
+
+                    if (response.get('response_code') != 0):
+                        if (response.get('response_reason') == "Queue is full" or
+                                response.get('response_reason') == "Queue is not available"):
+                            time.sleep(0.5)
```
jdye64, line 200: Is the idea here that the memory backing the queue is full and that would be the cause of a full queue here or is there an imposed maximum size?

src/nv_ingest/modules/sinks/message_broker_task_sink.py
```diff
@@ -48,7 +49,7 @@ def extract_data_frame(message: ControlMessage) -> Tuple[Any, Dict[str, Any]]:
     """
     try:
         with message.payload().mutable_dataframe() as mdf:
```
jdye64, line 51: Only because I am curious. How is `message.payload()` transferred to this code? Is it serialized behind the into (JSON, something else) and then transferred via Morpheus here? Some sort of shared memory? Something else?

src/nv_ingest/modules/sources/message_broker_task_source.py
```diff
@@ -0,0 +1,212 @@
...
+            db=broker_params.get("db", 0),
+            max_retries=validated_config.broker_client.max_retries,
+            max_backoff=validated_config.broker_client.max_backoff,
+            connection_timeout=validated_config.broker_client.connection_timeout,
+            use_ssl=broker_params.get("use_ssl", False),
+        )
+    elif client_type == "simple":
+        # Start or retrieve the singleton SimpleMessageBroker server
+        # TODO(Devin) add config param for max_queue_size
+        max_queue_size = broker_params.get("max_queue_size", 10000)
```
jdye64, line 212: I think this answers my earlier comment about the queue being full. Seems like a max imposed size of 10000

jdye64 (changes requested):
Left a few comments. Most were just clarification questions. I have listed the things here that I think would be good to address before merging. Big PR and good work!

Address:
- Seems like the the Conda activation should occur before the `pip install` steps in the docs/deployment.md. I might be missing some intended reason for that however?
- I think we have reached the point where a completely new docker compose file for this type of deployment is warranted. Maybe something like `docker-compose-lite.yaml`? Note: I'm working on adding us a `Makefile` setup so we can have targets invoking each of those perspective compose files making deployments more clear.
- I think an Enum, or some other centralized config, for the SimpleBroker messages would be good instead of using hardcoded string values.

General:
- I think the PR is already large enough but maybe we can have a followup PR that shows more about how to use the "library mode" with some docs or a notebook or something. Don't think we need that here just wanted to point it out.

Thanks again!

client/src/nv_ingest_client/nv_ingest_cli.py
drobison00: This is a bit confusing. The issue is that from the CLI's perspective, its using either a SimpleClient or a RESTClient. The CLI has no way of submitting Redis native requests to a Redis broker at the moment.

Historically, this was because we'd originally planning to standardize only on the REST interface and leave the choice of broker to the REST provider.

In contrast, the docker-compose configuration has Redis/Simple, because that dictates how the pipeline source is configured.

docker-compose.yaml
drobison00: No; I'll update so we do that.

src/nv_ingest/util/nim/paddle.py
drobison00: I think this had something to do with differences I was seeing between the local NIM and the one on BUILD. I'll double check to make sure its still needed -- if so I'll add a comment on why.

client/src/nv_ingest_client/message_clients/rest/rest_client.py
drobison00: The ResponseSchema has entries for the trace_id, and the transaction_id (see below). I think the reason they don't show up here is because this is the 'fetch' side, which doesn't do any tracing.  The `submit_message` path below this does add the trace/trasaction ids.

```python
class ResponseSchema(BaseModel):
    response_code: int
    response_reason: Optional[str] = "OK"
    response: Union[str, dict, None] = None
    trace_id: Optional[str] = None  # Unique trace ID
    transaction_id: Optional[str] = None  # Unique transaction ID
```

docker-compose.yaml
drobison00: Yes, I think we're getting to the point where we need a higher level of abstraction for deployment configuration. Probably something we should discuss.

client/src/nv_ingest_client/message_clients/simple/simple_client.py
drobison00: Sounds reasonable to me.

client/src/nv_ingest_client/message_clients/simple/simple_client.py
drobison00: There is an imposed maximum size, per queue. Currently this defaults to 10,000 where its used.

src/nv_ingest/modules/sinks/message_broker_task_sink.py
drobison00: Payload is a MessageMeta object which, generally, wraps a cuDF dataframe object. Morpheus has some specialized hooks into cuDF for managing it on the c++ side.

docker-compose.yaml
jdye64: Agreed. Nothing needed for this PR however just wanted to point it out. I'll resolve this.

jdye64 (approved):

PR #248, opened 2024-12-03
Convert table content to pseudo-markdown format

edknv:
## Description
Resolves https://github.com/NVIDIA/nv-ingest-private/issues/63.

## Checklist


docker-compose.yaml
```diff
@@ -133,22 +133,21 @@ services:
       - sys_nice
     environment:
       # Self-hosted cached endpoints.
-      #- CACHED_GRPC_ENDPOINT=cached:8001
+      - CACHED_GRPC_ENDPOINT=cached:8001
```
jdye64, line 136: Only because I am curious why are we switching back to grpc here? Is something not working with the http version?

docker-compose.yaml
```diff
@@ -133,22 +133,21 @@ services:
...
-      #- DEPLOT_HEALTH_ENDPOINT=deplot:8000
-      #- DEPLOT_HTTP_ENDPOINT=http://deplot:8000/v1/chat/completions
+      - DEPLOT_HTTP_ENDPOINT=http://deplot:8000/v1/chat/completions
       # build.nvidia.com hosted deplot
+      #- DEPLOT_HTTP_ENDPOINT=https://ai.api.nvidia.com/v1/vlm/google/deplot
       - DEPLOT_INFER_PROTOCOL=http
-      - DEPLOT_HTTP_ENDPOINT=https://ai.api.nvidia.com/v1/nvdev/vlm/google/deplot
       - DOUGHNUT_GRPC_TRITON=triton-doughnut:8001
-      - INGEST_LOG_LEVEL=DEBUG
+      - INGEST_LOG_LEVEL=DEFAULT
```
jdye64, line 150: Good catch

docker-compose.yaml
```diff
@@ -176,7 +175,7 @@ services:
       - YOLOX_HTTP_ENDPOINT=http://yolox:8000/v1/infer
       - YOLOX_INFER_PROTOCOL=grpc
       # build.nvidia.com hosted yolox endpoints.
-      #- YOLOX_HTTP_ENDPOINT=https://ai.api.nvidia.com/v1/nvdev/cv/nvidia/nv-yolox-page-elements-v1
+      #- YOLOX_HTTP_ENDPOINT=https://ai.api.nvidia.com/v1/cv/nvidia/nv-yolox-page-elements-v1
```
jdye64, line 178: Agree this is good to remove the `nvdev` since most folks will not have the appropriate key to access those endpoints.

jdye64 (approved):
Had a question about changing NIM endpoints from HTTP back to GRPC but other than that looks good

tests/nv_ingest/util/nim/test_paddle.py
```diff
@@ -0,0 +1,177 @@
...
+
+        data = {"base64_image": "mock_base64_string"}
+        result = legacy_paddle_ocr_model.prepare_data_for_inference(data)
+
+    mock_paddle_grpc_response = np.array([[b"mock_text"]])
+
+    result = legacy_paddle_ocr_model.parse_output(
+        mock_paddle_grpc_response, protocol="grpc", table_content_format="foo"
+    )
+    assert result[0] == "mock_text"
```
drobison00, line 176: This is great. Thanks for the new unit tests on the Nim.

drobison00 (approved):

docker-compose.yaml
edknv: It's mostly to align cached with yolox and paddle, which are on grpc. Besides, my understanding is that there are ongoing discussions on the performance of the http layers in the NIMs.

PR #254, opened 2024-12-05
Brev.dev Nv-Ingest PDF Launchable Notebook

jdye64:
## Description
Brev.dev offers the concept of a "launchable". The idea is a quick bootstrapped environment for testing out products using notebooks. This PR introduces a notebook that can be used in conjunction with brev.dev to demonstrate the capabilities of nv-ingest.

Notes:
- Otel prometheus port was changed from 8888->9988 because that conflicts with the jupyter notebook running in brev.dev. I think Jupyter notebook is common enough we should change that port regardless.
- A new docker-compose profile of `--profile retrieval` was added which bulk enables `etcd, minio, milvus, and attu`

## Checklist


docker-compose.yaml
```diff
@@ -218,90 +218,98 @@ services:
     ports:
       - "3000:3000"
 
-#  etcd:
```
drobison00, line 221: Do we really want to default these to deploy?

drobison00 (approved):

docker-compose.yaml
randerzander: note the added
profiles:
  - retrieval

they will only deploy if you pass the `--profile retrieval` arg to docker compose up

jdye64: They won't deploy by default. The user must pass in the `--profile retrieval` argument for them to deploy. Like

`docker compose up --profile retrieval` if they leave that off they will not be deployed.

PR #261, opened 2024-12-09
Simplify Dockerfile and begin transition to standardize on conda environments

drobison00:
## Checklist


jdye64 (approved):
I like having all of the versions strictly pinned down personally. It does make things slightly more difficult when/if we add or update dependencies but I think the juice is worth the squeeze there.

Might you add the freeze command you used so we can reference it later? My assumption given the detail of the conda files is you generated them from a previously solved environment. I think having that process documented somewhere is critical for other devs to know so if they need to update a dep they can roughly ...

1. Create environment from this existing file
2. Modify that environment for their needs
3. Run your "freeze command", guessing contribution guide is best place for this?
4. And then last take that and update the conda .yml file with those changes for their PR

drobison00:
> I like having all of the versions strictly pinned down personally. It does make things slightly more difficult when/if we add or update dependencies but I think the juice is worth the squeeze there.
> 
> Might you add the freeze command you used so we can reference it later? My assumption given the detail of the conda files is you generated them from a previously solved environment. I think having that process documented somewhere is critical for other devs to know so if they need to update a dep they can roughly ...
> 
> 1. Create environment from this existing file
> 2. Modify that environment for their needs
> 3. Run your "freeze command", guessing contribution guide is best place for this?
> 4. And then last take that and update the conda .yml file with those changes for their PR

Makes sense.

You're right, it was a bit of a bootstrapped process, but, I think going forward it will just consist of updating the created yaml file.

PR #263, opened 2024-12-09
add requirements and final abstraction

jperez999:
## Description
This PR fixes the requirements to support the use of the milvus tools. It also creates another layer of abstraction to make it easier to create a nv-ingest collection, write records to that collection and search against those records. Supports both hybrid and dense search.

## Checklist


jdye64 (approved):
LGTM - Note we added the `azure-storage-blob` dependency but that is fine since it is MIT licensed

PR #275, opened 2024-12-12
add streaming milvus capability, output_fields param, and pin pacakges

jperez999:
## Description
This PR adds support for milvus lite, it checks for the version of milvus and if you are using the embedded version, ensures you use the correct index params (no GPU CAGRA) and it uses streaming insert. Also added output_fields as a param for searches. Pinned versions of packages to ensure pypi only downloads one version of each package instead of download all of them.

## Checklist


jdye64 (approved):
LGTM

PR #278, opened 2024-12-12
Fix sparse embedding logic

jperez999:
## Description
This PR makes sure that if a sparse embedding ends up with no embedding values, we add a placeholder value. Otherwise milvus will not accept the record.

## Checklist


ChrisJar (approved):
LGTM

PR #281, opened 2024-12-12
Prop dim param

jperez999:
## Description
This PR changes the default dense dimension size to 2048 to support the new model. It also propagates the parameter to the top level extraction so it can be used to change the schema as the user requires. 

## Checklist


ChrisJar (approved):
LGTM

PR #283, opened 2024-12-13
Switch split task to token based splitting

ChrisJar:
## Checklist


src/nv_ingest/modules/transforms/nemo_doc_splitter.py
drobison00, line 1: Since we're basically changing the whole thing, lets rename it from nemo_doc_splitter to something more generic.

src/nv_ingest/modules/transforms/nemo_doc_splitter.py
```diff
@@ -167,16 +124,11 @@ def split_and_forward(message: ControlMessage):
...
             # Override parameters if set
-            split_by = task_props.get("split_by", validated_config.split_by)
-            split_length = task_props.get("split_length", validated_config.split_length)
-            split_overlap = task_props.get("split_overlap", validated_config.split_overlap)
-            max_character_length = task_props.get("max_character_length", validated_config.max_character_length)
-            sentence_window_size = task_props.get("sentence_window_size", validated_config.sentence_window_size)
+            tokenizer  = task_props.get("tokenizer", validated_config.tokenizer)
+            chunk_size = task_props.get("chunk_size", validated_config.chunk_size)
 
             logger.info(
```
drobison00: I know its not part of your PR changes, but could you swap this to a debug print.

src/nv_ingest/modules/transforms/nemo_doc_splitter.py
```diff
@@ -188,14 +140,11 @@ def split_and_forward(message: ControlMessage):
...
-                units = _split_into_units(content, split_by)
-                text_splits = _concatenate_units(
-                    units,
-                    split_length,
-                    split_overlap,
-                    max_character_length=max_character_length,
-                )
-                split_docs.extend(_build_split_documents(row, text_splits, sentence_window_size=sentence_window_size))
+                os.environ['TOKENIZERS_PARALLELISM'] = "False"
+                tokenizer_model = AutoTokenizer.from_pretrained(tokenizer)
```
drobison00: This has the potential to download new weights at runtime. We should try and pre-cache our supported Tokenizers at container build time, and/or allow the stage configuration to take an 'allow_downloads' parameter.

src/nv_ingest/modules/transforms/nemo_doc_splitter.py
```diff
@@ -188,14 +140,11 @@ def split_and_forward(message: ControlMessage):
...
 
-                units = _split_into_units(content, split_by)
-                text_splits = _concatenate_units(
-                    units,
-                    split_length,
-                    split_overlap,
-                    max_character_length=max_character_length,
-                )
-                split_docs.extend(_build_split_documents(row, text_splits, sentence_window_size=sentence_window_size))
+                os.environ['TOKENIZERS_PARALLELISM'] = "False"
```
drobison00: This should probably only be done once, and it has potential ramifications across the pipeline, because at some point tokenizers may be used somewhere else. Lets either set this at stage construction, or maybe even at pipeline level construction.

src/nv_ingest/modules/transforms/nemo_doc_splitter.py
drobison00: As a follow-on here, we're also going to want to create an approved list of models, probably the same as the pre-cached set, and add an additional flag if an end user wants to use one that isn't approved.

src/nv_ingest/modules/transforms/nemo_doc_splitter.py
ChrisJar: I think this is actually safe to take out completely, forgot to remove it earlier

src/nv_ingest/modules/transforms/nemo_doc_splitter.py
ChrisJar: So basically predownloading the vocab files from a list of approved models at build time so that they can be used later without accessing the internet?

tests/nv_ingest/schemas/test_text_splitter_schema.py
```diff
@@ -0,0 +1,28 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024, NVIDIA CORPORATION & AFFILIATES.
```
drobison00, line 1: Please add at least one more test that actually sets all the parameters for the schema to ensure its working as expected, and tests that confirm that the parameter limits are being handled correctly (can't set tokens to negative numbers, etc...)

src/nv_ingest/modules/transforms/text_splitter.py
```diff
@@ -0,0 +1,171 @@
...
+            with message.payload().mutable_dataframe() as mdf:
+                df = mdf.to_pandas()
+
+            # Filter to text only
+            bool_index = df["document_type"] == ContentTypeEnum.TEXT
+            df_filtered = df.loc[bool_index]
+
+            if df_filtered.empty:
+                gdf = cudf.from_pandas(df)
+                message_meta = MessageMeta(df=gdf)
```
drobison00: It doesn't look like we've mutated the original message at all at this point. We can probably just return the `message` without rebuilding the payload.

src/nv_ingest/modules/transforms/text_splitter.py
```diff
@@ -0,0 +1,171 @@
...
+                f"chunk_overlap: {chunk_overlap}"
+            )
+
+            tokenizer_model = AutoTokenizer.from_pretrained(tokenizer)
+
+            split_docs = []
+            for _, row in df_filtered.iterrows():
+                content = row["metadata"]["content"]
+
+                if content is None:
```
drobison00: Does this necessarily need to be an error? Or could we just set it to an empty string and continue processing?

src/nv_ingest/schemas/text_splitter_schema.py
```diff
@@ -0,0 +1,15 @@
...
+# All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+
+from pydantic import Field, BaseModel
+
+from typing_extensions import Annotated
+
+
+class TextSplitterSchema(BaseModel):
```
drobison00, line 12: Probably want some additional validation so that chunk_overlap has to be < chunk_size.

src/nv_ingest/modules/transforms/text_splitter.py
```diff
@@ -0,0 +1,171 @@
...
+def _split_into_chunks(text, tokenizer, chunk_size=300, chunk_overlap=0):
+    # Tokenize the text into token IDs
+    encoding = tokenizer.encode_plus(text, add_special_tokens=False, return_offsets_mapping=True)
+
+    # Get the token IDs and offsets for splitting
+    tokens = encoding["input_ids"]
+    offsets = encoding["offset_mapping"]
+
+    # Split the tokens into chunks of the desired size
+    chunks = [tokens[i : i + chunk_size] for i in range(0, len(tokens), chunk_size - chunk_overlap)]
```
drobison00: Related to my comment in the schema, we will want to be sure that chunk_overlap is strictly less than chunk_size. Otherwise we could conceivably get a negative size here.

ChrisJar:
@drobison00 Thanks for the reviews! I have a couple questions: How do you think we should go about preloading the vocab files in the case that the user doesn't want to allow downloads and in the case they do, how should we go about passing along the huggingface token to access gated models? My thought was to pull it from an environment variable on the client side like we do with unstructured and adobe and pass it along as another parameter in the schema

src/nv_ingest/modules/transforms/nemo_doc_splitter.py
drobison00: We can ignore this for now.

ChrisJar:
Also I can't seem to reproduce the test failure locally

edknv:
> Also I can't seem to reproduce the test failure locally

We have a flaky test. I meant to look into fixing it, but for now you can go into Actions and rerun the test.

src/nv_ingest/modules/transforms/text_splitter.py
```diff
@@ -0,0 +1,175 @@
...
+                split_source_types
+            )
+            df_filtered = df_filtered.loc[bool_index]
+
+            if df_filtered.empty:
+                return message
+
+            tokenizer_model = AutoTokenizer.from_pretrained(tokenizer, token=hf_access_token)
+
+            split_docs = []
```
drobison00, line 134: Matching documents should be collected and processed all at once as opposed to iterating rows; and we should actually make this a multiprocessing stage so we're able to use the worker pool for CPU bound tasks. But we can hold of if this needs to go in sooner.

drobison00 (approved):

README.md
```diff
@@ -396,6 +396,10 @@ https://pypi.org/project/pdfservices-sdk/
     required if you want to use the Adobe extraction service for PDF decomposition. Please review the
     [license agreement](https://github.com/adobe/pdfservices-python-sdk?tab=License-1-ov-file) for the
     pdfservices-sdk before enabling this option.
+- **`Split`**:
```
randerzander:
```suggestion
- **`Built With Llama:`**:
```

randerzander (changes requested):

src/nv_ingest/schemas/text_splitter_schema.py
```diff
@@ -0,0 +1,20 @@
...
+
+from pydantic import Field, BaseModel, field_validator
+
+from typing_extensions import Annotated
+
+
+class TextSplitterSchema(BaseModel):
+    tokenizer: str = "meta-llama/Llama-3.2-1B"
+    chunk_size: Annotated[int, Field(gt=0)] = 1024
+    chunk_overlap: Annotated[int, Field(ge=0)] = 20
```
shan-nvidia: A good range of chunking overlap is 10%-20% of chunk size. We used 15% overlapping in our experiments, which translates to ~150 tokens chunk_overlap.

Dockerfile
```diff
@@ -124,6 +129,12 @@ RUN --mount=type=cache,target=/opt/conda/pkgs\
     && pip install ./api/dist/*.whl \
     && pip install ./client/dist/*.whl
 
+
+RUN --mount=type=cache,target=/opt/conda/pkgs \
+    --mount=type=cache,target=/root/.cache/pip \
```
drobison00:
```suggestion
RUN  --mount=type=cache,target=/root/.cache/pip \
```

drobison00 (approved):

PR #291, opened 2024-12-17
Upgrade to Pydantic2

jdye64:
## Description
This PR updates the nv-ingest codebase to use Pydantic 2 instead of Pydantic 1. We have seen challenges where other libraries being installed require Pydantic greater than 2 while we are still using Pydantic 1. This is causing dependency issues and therefore we are making the switch to Pydantic 2

## Checklist


drobison00 (changes requested):
Looks good; one small issue I found.

`conda/packages/**/meta.yaml` need their pydantic version updated as well.

jdye64:
> Looks good; one small issue I found.
> 
> `conda/packages/**/meta.yaml` need their pydantic version updated as well.

Good catch I just updated it. I'm trying to perform lots of different runs now. I have a feeling that there is certainly a few more things that need tweaked as well.

PR #299, opened 2024-12-18
Remove unused yolox functions and fix nim client shut down.

edknv:
## Description

Some small clean-up items:
- `process_inference_results` has been refactored and moved [here](https://github.com/NVIDIA/nv-ingest/blob/main/src/nv_ingest/util/nim/yolox.py#L251-L293).
- `yolox_client` etc are all instances of `NimClient` so there's no need for the isinstance check.

## Checklist


drobison00 (approved):
@edknv Is this all because its being handled in the NIM now?

PR #365, opened 2025-01-21
Replace Cached with Yolox-graphic-elements and drop Deplot

edknv:
## Checklist


docker-compose.yaml
```diff
@@ -99,7 +77,7 @@ services:
         reservations:
           devices:
             - driver: nvidia
-              device_ids: ["1"]
+              device_ids: ["0"]
```
randerzander, line 80:
```suggestion
              device_ids: ["1"]
```
shouldn't we keep it on #1, since that's where cached was? Moving to 0 I think is leading to compute competition with everything else already on 0

docker-compose.yaml
```diff
@@ -34,12 +34,12 @@ services:
...
               capabilities: [gpu]
     runtime: nvidia
 
-  deplot:
-    image: ${DEPLOT_IMAGE:-nvcr.io/nvidia/nemo-microservices/deplot}:${DEPLOT_TAG:-1.0.0}
+  yolox-graphic-elements:
```
randerzander, line 41: I might be misreading GitHub's review changes UI, but can we put this new NIM on device 1, since 0 is overloaded?

edknv: We used to have Deplot on device 0, but we can remove Deplot with this PR, so I moved everything to device 0 including this new NIM. Isn't that what we wanted to do by dropping Deplot, i.e., fit everything in 1 device and drop the minimum requirement to 1 GPU?

randerzander: that is one goal, but we should do that subject to more benchmarking. we don't yet have the ability to set the batch size, and I'd like to confirm we're not saturating compute capacity of a single device (for good portion of bo767 execution, both GPUs have >50% utilization).

edknv: ok, moved the new nim to device 1 in b5fd7fa.

randerzander: I did more benchmarking.. results are great, we can move everything back to a single 80GB GPU :)

docker-compose.yaml
randerzander, line 1: Let's also go ahead and make the switch to the llama embedder. This is what I was testing on the single GPU deployment.

EMBEDDING_IMAGE=nvcr.io/nim/nvidia/llama-3.2-nv-embedqa-1b-v2
EMBEDDING_TAG=latest
EMBEDDING_NIM_MODEL_NAME=nvidia/llama-3.2-nv-embedqa-1b-v2

edknv: Switched to llama in https://github.com/NVIDIA/nv-ingest/pull/365/commits/41f9e7be45d649b77b8971ae40925ffc527b38e4.

PR #369, opened 2025-01-23
Throughput improvements when number of PDFs is less than extract progress engines

drobison00:
## Checklist


src/nv_ingest/extraction_workflows/pdf/pdfium_helper.py
```diff
@@ -263,117 +330,121 @@ def pdfium_extractor(
...
-                        pass  # Pdfium failed to extract the image associated with this object - corrupt or missing.
-
-        # Table and chart collection
-        if extract_tables or extract_charts:
-            pages.append(page)
-
-    if extract_text and text_depth == TextTypeEnum.DOCUMENT and len(accumulated_text) > 0:
-        text_extraction = construct_text_metadata(
+        # Now wait for all futures to complete
+        for fut in concurrent.futures.as_completed(futures):
```
jdye64, line 428: I actually don't know this ... if say 10 futures are in this list. One of them fails, what prevents this from continuing on with only partial results? Or is the exception purposely not caught so the entire batch fails?

drobison00: The entire batch fails. Basically, any failure in document processing results in a failure of the entire document, so any failure in a batch is basically a failure in the entire document.

Generally speaking I don't know how we'd give someone enough information back for them to decide if a partial failure was acceptable or not, so its always been a failure. That said, the relaxed retry cases in the Nim Clients should remove many of the spurrious failures we'd seen previously.

PR #384, opened 2025-01-28
Filter out whitespace-only strings before embedding

edknv:
## Description
Embedding NIM services return errors if the input text is empty or all whitespace. This PR updates the content filtering logic to ensure strings with only whitespace are excluded from the embedding process.

## Checklist


jdye64 (approved):
LGTM. Sensible workaround for the current situation

drobison00 (approved):

Issue #353, opened 2025-01-20
[FEA]: Add ability to detect bounding boxes for compound images during extraction


drobison00:
### Is this a new feature, an improvement, or a change to existing functionality?

New Feature

### How would you describe the priority of this feature request

Significant improvement

## Description

During the extraction phase for PDFs, PPTx, and Docx, we often encounter situations where we have a number of small images that collectively make up a compound image. Currently, we treat all of these images independently, returning each as its own primitive in the extraction results.

We want to improve this behavior and make it configurable, so that it is possible to preserve the existing behavior, or to instruct the extraction phase to attempt to identify a single bounding box for a set of connected images.

**Note**

- We do **not** know a-priori if a collection of images is part of a single image.
- There may be **multiple distinct collections** on a page.

### Describe the feature, and optionally a solution or implementation and any alternatives

This issue aims to define an approach to:
1. Identify clusters of bounding boxes that belong to the same compound image.
2. Compute the overall bounding box for each identified compound image.

## Required Behavior

- [ ] Group bounding boxes that belong to the same compound image.
- [ ] Handle scenarios where bounding boxes may be close but not necessarily overlapping.
- [ ] Compute a minimal bounding box that encapsulates all grouped bounding boxes.
- [ ] Support configurable proximity thresholds for clustering.
- [ ] Ensure that detected table/chart images are not included in connected component bounding boxes.
- [ ] Output detected compound image bounding boxes.
- [ ] Provide visualization for debugging grouped bounding boxes.
- [ ] Efficient performance on large collections of bounding boxes.

### Steps:

1. **Initialization:**
   - Identify all bounding boxes on a page.
   - Exclude table/chart bounding boxes from processing.

2. **Expansion:**
   - Expand each bounding box by a configurable margin.
   - Merge overlapping or adjacent bounding boxes iteratively.

3. **Refinement:**
   - Apply post-processing to fine-tune the final bounding box.
   - Remove potential over-grouping by applying size and aspect ratio constraints.

4. **Output:**
   - Store and visualize final compound bounding boxes.

### Scenario 1: Adjacent Boxes Forming a Compound Image

```mermaid
graph TD
    A[Box 1] -->|Close to| B[Box 2]
    B -->|Close to| C[Box 3]
    D[Box 4] -- No Connection --> E[Box 5]
```

Expected output:
- Compound Image 1: { Box 1, Box 2, Box 3 }
- Compound Image 2: { Box 4, Box 5 }

---

[...]

PR #386, opened 2025-01-29
Add ability to group collections of image bounding boxes

edknv:
## Description

Partially addresses #353. This PR is focused on PDFs. Other document types will be addressed in future PRs.

This PR adds support for a new `extract_images_method` option in the PDFium-based extraction workflow, allowing users to choose between a simple “as-is” extraction of image objects (`simple`) or a “grouping” approach (`group`) that groups related bounding boxes together. It introduces a new parameter, `extract_images_params`, for configuring thresholds and additional clustering details.

In the previous approach (`simple`), we only extracted `PdfImageObject`s, but PDFium provides several additional graphics objects in addition to PdfImageObject. For our purposes, we now also extract:
- The ~~[PdfPathObject](https://pdfium.patagames.com/help/html/T_Patagames_Pdf_Net_PdfPathObject.htm)~~ [FPDF_PAGEOBJ_PATH](https://pdfium.googlesource.com/pdfium/+/refs/heads/main/public/fpdf_edit.h#40) The PdfPathObject is an arbitrary shape made up of straight lines, rectangles, and cubic Bézier curves. A path may intersect itself and may have disconnected sections and holes.
The ~~[PdfFormObject](https://pdfium.patagames.com/help/html/T_Patagames_Pdf_Net_PdfFormObject.htm)~~ [FPDF_PAGEOBJ_FORM](https://pdfium.googlesource.com/pdfium/+/refs/heads/main/public/fpdf_edit.h#43) is used to group graphical elements together as a unit for various purposes.

As an example, the following has a PdfFormObject. The object inside the red box is a form object that contains several nested component images. If we use the bounding box of the PdfFormObject (instead of naively extracting all the images with depth > 1), we can correctly group together all the components.
image

In the following example, we have what looks like an image, but it's actually made of many "shapes" (PdfPathObject). We can group them together as one single object. 
image

## Checklist


src/nv_ingest/extraction_workflows/pdf/pdfium_helper.py
```diff
@@ -26,24 +26,23 @@
...
-from nv_ingest.util.pdf.metadata_aggregators import Base64Image
 from nv_ingest.util.pdf.metadata_aggregators import CroppedImageWithContent
 from nv_ingest.util.pdf.metadata_aggregators import construct_image_metadata_from_pdf_image
 from nv_ingest.util.pdf.metadata_aggregators import construct_table_and_chart_metadata
 from nv_ingest.util.pdf.metadata_aggregators import construct_text_metadata
 from nv_ingest.util.pdf.metadata_aggregators import extract_pdf_metadata
-from nv_ingest.util.pdf.pdfium import PDFIUM_PAGEOBJ_MAPPING
 from nv_ingest.util.pdf.pdfium import pdfium_pages_to_numpy
-from nv_ingest.util.pdf.pdfium import pdfium_try_get_bitmap_as_numpy
+from nv_ingest.util.pdf.pdfium import extract_nested_simple_images_from_pdfium_page
```
drobison00, line 46: Unless there is a specific reason not to, can we generalize this idea to extracting from an image vs a pdfium page.

We will also need to propagate the generalized logic for bounding box grouping into pptx, docx, and the image extractors.

src/nv_ingest/util/pdf/pdfium.py
```diff
@@ -188,3 +195,354 @@ def pdfium_pages_to_numpy(
...
+        path_bbox = convert_pdfium_position(obj.get_pos(), page_width, page_height)
+        path_bboxes.append(path_bbox)
+
+    # If merging is disabled or no bounding boxes were found, return them as-is
+    if (not merge) or (not path_bboxes):
+        return path_bboxes
+
+    merged_bboxes = []
+    path_groups = group_bounding_boxes(path_bboxes)
+    path_bboxes = combine_groups_into_bboxes(path_bboxes, path_groups, min_num_components=3)
```
drobison00: Why 3? May want to make this configurable if its something that could be tinkered with for accuracy.

src/nv_ingest/util/pdf/pdfium.py
```diff
@@ -188,3 +195,354 @@ def pdfium_pages_to_numpy(
...
+    (xmin1, ymin1, xmax1, ymax1) = b1
+    (xmin2, ymin2, xmax2, ymax2) = b2
+
+    # Check if they overlap horizontally
+    overlap_x = not (xmax1 < xmin2 or xmax2 < xmin1)
+    # Check if they overlap vertically
+    overlap_y = not (ymax1 < ymin2 or ymax2 < ymin1)
+
+    # If there's an actual overlap area, that's enough
+    if overlap_x and overlap_y:
```
drobison00: Wondering if we need two checks here vs just doing a single check on the expanded boxes. Is there a performance reason?

src/nv_ingest/util/pdf/pdfium.py
```diff
@@ -188,3 +195,354 @@ def pdfium_pages_to_numpy(
...
+
+    x0 = max(0, x0)
+    y0 = max(0, y0)
+    x1 = min(page_width, x1)
+    y1 = min(page_height, y1)
+
+    return [int(x0), int(y0), int(x1), int(y1)]
+
+
+def boxes_are_close_or_overlap(b1, b2, threshold=10.0):
```
drobison00: In line with my comment on the imports. I think these functions should be generalized for images, moved into a common utility library, and utilized in all the other extraction stages. Any specific logic related to pdfium pages can stay here.

src/nv_ingest/util/pdf/pdfium.py
```diff
@@ -188,3 +195,354 @@ def pdfium_pages_to_numpy(
...
+    """
+    Given a list of bounding boxes,
+    returns a list of groups (lists) of bounding box indices.
+    """
+    n = len(boxes)
+    visited = [False] * n
+    adjacency_list = [[] for _ in range(n)]
+
+    # Build adjacency by checking closeness/overlap
+    for i in range(n):
```
drobison00: This is probably fine for small numbers of boxes; but its still O(n^2), which could cause problems on some of the pathological PDFs we've seen in the past. Some of the US-GS topo pdfs have layouts where there is effectively 1 image per pixel, and 10s of 1000s of images on the page.

I may be missing if you're doing this somewhere, but I think we should use a different approach (maybe a quadtree), skip grouping, or emit some warnings if the number of bounding boxes on a page exceeds some threshold.

drobison00 (changes requested):

src/nv_ingest/util/pdf/pdfium.py
edknv: Made it configurable in [64748f0](https://github.com/NVIDIA/nv-ingest/pull/386/commits/64748f0cd94f12eafe7c64dbaa12da5180159a56).

src/nv_ingest/util/pdf/pdfium.py
edknv: You are right. It was a left over from previous experiments. Removed the redundant check.

src/nv_ingest/util/pdf/pdfium.py
edknv: Moved these functions to `util/image_processing`.

src/nv_ingest/util/pdf/pdfium.py
edknv: I think it makes sense to skip grouping and emit a warning if the number of bounding boxes exceed the threshold. From my experiments, most of the gain come from using the bounding box of PdfFormObject and grouping PdfPathObject. These PDFs with an unreasonable number of objects don't gain much by grouping them because they seem to be just badly constructed PDFs to begin with. In my opinion, they would be better solved by OCR models.

src/nv_ingest/extraction_workflows/pdf/pdfium_helper.py
edknv: I aimed to address only PDFs in this PR. I'm not sure yet if the approach in this PR can be generalized to other document types, but I'll be looking at pptx/docx and images next. But for now, I think these functions will be PDF specific. I'll revisit and refactor depending on how it works with other document types.

mara004:
Note that the docs linked above are for a commercial .NET SDK and not applicable to pypdfium2/pdfium.
The official docs are here:
https://pypdfium2.readthedocs.io/
https://pdfium.googlesource.com/pdfium/+/refs/heads/main/public/ (see the comments in the header files)

src/nv_ingest/util/pdf/pdfium.py
```diff
@@ -188,3 +198,219 @@ def pdfium_pages_to_numpy(
         images.append(img_arr)
 
     return images, padding_offsets
+
+
+def convert_pdfium_position(pos, page_width, page_height):
+    """
+    Convert a PDFium bounding box (which typically has an origin at the bottom-left)
+    to a more standard bounding-box format with y=0 at the top.
+    """
```
mara004, line 227: I'm not sure this method is generally correct, as the PDF spec allows for a pretty arbitrary coordinate system.
Indeed, the origin typically is the bottom left, but IIRC the coordinate system can be laid out between any opposite corners in principle. Also, the coordinate system does not necessarily have to start at 0, it could be any arbitrary value.

Related: https://github.com/pypdfium2-team/pypdfium2/discussions/284

drobison00 (approved):

PR #387, opened 2025-01-29
BugFix: Abort on certain response codes and retry on others

### LINKED ISSUE #382: [BUG]:
### Version

24.10

_No response_

### Describe the bug.

When using the REST Client to fetch a message from FastAPI, if the job ID does not exist, the Rest proxy will immediately return a '404', and the client will immediately re-issue the fetch request. This results in 'spamming' the log outputs, with something similar to the snippet below. 

This is incorrect behaviorally: we should engage in some kind of exponential backoff to avoid flooding the user, and recognize that there is a failure if the job ID does not exist after some grace period.

This is also incorrect technically: the 404 isn't a timeout, its an HTTP request failure.

```bash
DEBUG:nv_ingest_client.util.processing:Error processing future result: Timeout: No response within 10 seconds for job ID 0
DEBUG:nv_ingest_client.message_clients.rest.rest_client:Invoking fetch_message http endpoint @ 'http://localhost:7670/v1/fetch_job/'
DEBUG:urllib3.connectionpool:Starting new HTTP connection (1): localhost:7670
```

```shell
No direct reproducer, this behavior was discovered when diagnosing another issue where a submit_message call to the FastAPI service was failing and silently failing job creation.
```

### Relevant log output

```shell

```

### Other/Misc.

_No response_

jdye64:
## Description
The previous logic grouped all possible HTTP response codes into the same "bucket". This meant that any response received from the server would result in a retry. Some HTTP response codes like 404 & 500 are unlikely to remediate themselves with a retry with the extra same input and conditions and therefore spam the client with nearly unlimited retries. This PR adjust that logic to instead through RuntimeErrors in those conditions and offer a placeholder where other conditions of that manner can later be added as well.

This closes #382 

## Checklist

### DIFF
--- a/client/src/nv_ingest_client/client/client.py
+++ b/client/src/nv_ingest_client/client/client.py
@@ -321,7 +321,19 @@ class NvIngestClient:
                     # Only pop once we know we've successfully decoded the response or errored out
                     _ = self._pop_job_state(job_index)
             else:
-                raise TimeoutError(f"Timeout: No response within {timeout} seconds for job ID {job_index}")
+                # There are a plethora of codes that can be thrown. Some offer specific insights while
+                # others can be grouped into a general failure category. We check for specific codes here
+                # and then generally error on the others.
+                if response.response_code == 404:
+                    # job_id not found on serverside. This condition will not alleviate itself with a retry
+                    raise RuntimeError(f"JobId: {job_state.job_id} not found - Reason: {response.response_reason}")
+                elif response.response_code == 500:
+                    # properly propagated server side error
+                    raise RuntimeError(f"Response: {response.response} - Reason: {response.response_reason}")
+                else:
+                    # Generalized errors group. These errors are ones that could potentially be resolved and
+                    # therefore should be retried. Existing logic works based on TimeoutErrors so we raise here.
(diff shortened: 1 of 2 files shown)


client/src/nv_ingest_client/client/client.py
```diff
@@ -321,7 +321,19 @@ def _fetch_job_result(self, job_index: str, timeout: float = 100, data_only: boo
                     # Only pop once we know we've successfully decoded the response or errored out
                     _ = self._pop_job_state(job_index)
             else:
-                raise TimeoutError(f"Timeout: No response within {timeout} seconds for job ID {job_index}")
+                # There are a plethora of codes that can be thrown. Some offer specific insights while
+                # others can be grouped into a general failure category. We check for specific codes here
+                # and then generally error on the others.
+                if response.response_code == 404:
```
drobison00, line 327: I don't know if we want to make the generic client aware of HTTP response codes. It's also used for Redis/SimpleBroker, and whatever else we may need to support in the future.

Maybe we just have recoverable vs non-recoverable response codes that it knows about and translate errors from the different client instances into those?

drobison00 (changes requested):

jdye64:
Resolution moved into PR https://github.com/NVIDIA/nv-ingest/pull/485

PR #409, opened 2025-02-06
add bounding box to content metadata on milvus

edknv:
## Description
This PR adds the bounding boxes of images/tables/charts to the metadata stored on milvus. To avoid inflating the milvus usage unnecessarily, we only extract the necessary bounding box values and insert them into the existing `content_metadata` field.

## Checklist

### DIFF
--- a/client/src/nv_ingest_client/util/milvus.py
+++ b/client/src/nv_ingest_client/util/milvus.py
@@ -370,6 +370,33 @@ def _pull_text(element, enable_text: bool, enable_charts: bool, enable_tables: b
     return text
 
 
+def _insert_location_into_content_metadata(element, enable_charts: bool, enable_tables: bool, enable_images: bool):
+    location = max_dimensions = None
+    if element["document_type"] == "structured":
+        location = element["metadata"]["table_metadata"]["table_location"]
+        max_dimensions = element["metadata"]["table_metadata"]["table_location_max_dimensions"]
+        if element["metadata"]["content_metadata"]["subtype"] == "chart" and not enable_charts:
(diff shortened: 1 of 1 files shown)


client/src/nv_ingest_client/util/milvus.py
```diff
@@ -370,6 +370,33 @@ def _pull_text(element, enable_text: bool, enable_charts: bool, enable_tables: b
...
+            location = max_dimensions = None
+    elif element["document_type"] == "image" and enable_images:
+        location = element["metadata"]["image_metadata"]["image_location"]
+        max_dimensions = element["metadata"]["image_metadata"]["image_location_max_dimensions"]
+    verify_emb = verify_embedding(element)
+    if (not location) or (not verify_emb):
+        source_name = element["metadata"]["source_metadata"]["source_name"]
+        pg_num = element["metadata"]["content_metadata"]["page_number"]
+        doc_type = element["document_type"]
+        if not verify_emb:
```
jperez999: can we remove the embedding check here. Since it already happens in the _pull text call. I would keep the location check for sure, that way we can tell when something does not have a location.

jperez999 (changes requested):

client/src/nv_ingest_client/util/milvus.py
edknv: Removed embedding check in [efed575](https://github.com/NVIDIA/nv-ingest/pull/409/commits/efed5754be97450a9ceba8fb5956f288acef3028)

jperez999 (approved):

PR #414, opened 2025-02-07
close pdfium document before submitting to thread pool

edknv:
## Checklist


drobison00:
This is a major performance regression, and its not clear why this fixes anything (it doesn't look like we're sending any pdfium objects to the executor). I think we need to better understand whats happening.

edknv:
This was only for demo purposes and wasn't meant to suggest a permanent fix.

PR #423, opened 2025-02-10
Improve exception handling and information content, and unify exception styles / NIM handling.

drobison00:
## Checklist


src/nv_ingest/modules/transforms/embed_extractions.py
```diff
@@ -40,7 +33,7 @@
 EmbedExtractionsLoaderFactory = ModuleLoaderFactory(MODULE_NAME, MODULE_NAMESPACE, EmbedExtractionsSchema)
 
 
-async def _make_async_request(
+def _make_async_request(
```
jdye64, line 36: I'm sure there is a good reason but why are the OpenAI async calls being changed? Is it to get more details on the exception encountered or something else?

jdye64 (approved):
A fan of improved exception handling

src/nv_ingest/modules/transforms/embed_extractions.py
drobison00: Its to bring it in line with what we do everywhere else. The codebase itself isn't fully Async compatible, so the process of using Async functions with a local process loop isn't really much different from using a thread pool, but its also a pattern we don't use anywhere else in the code base.

Its meant to be a consistency change rather than a functional one.

PR #433, opened 2025-02-11
add config to pull env vars into params

jperez999:
## Description
This PR adds a ClientConfigSchema that pulls parameters from env vars. These can then be used where ever they may be required. Currently they are only used in the nv-ingest-retrieval call. 

## Checklist


drobison00 (approved):

client/src/nv_ingest_client/util/util.py
```diff
@@ -31,6 +31,16 @@
 logger = logging.getLogger(__name__)
 
 
+class ClientConfigSchema:
```
drobison00, line 34: Could you add unit tests for this schema.

drobison00 (changes requested):

client/src/nv_ingest_client/util/util.py
jperez999: added unit test

PR #440, opened 2025-02-13
Add configuration options for changing embedding model name and NIM endpoint url at runtime

ChrisJar:
## Description

This PR adds model name, and NIM endpoint url configuration options to the embed task so that users can configure different embedding NIMs/models at runtime. If these configuration options aren't set it falls back to the current behavior of pulling the model name from an env variable when starting the container.

## Checklist


drobison00:
Looks good; sorry, the recent unification updates introduced some differences that break some of these issues. Happy to sync on it and help merge.

ChrisJar:
Implemented with #510

PR #449, opened 2025-02-14
Add retrieval profile to docker compose up command

nkmcalli:
Add retrieval profile to docker compose up command


sosahi (approved):

nkmcalli:
From slack: "That check is expected to fail currently and it should be fine to merge doc-only changes."

PR #483, opened 2025-02-21
Introduce infographic

edknv:
## Description

- `infographic` is added as a subtype of the `structured` type.
- Similarly to charts and tables, there is a stage that filters images tagged with the infographics subtype and calls PaddleOCR to extract text from the images.
- The page-elements version is detected from the metadata, and different hyperparameters are used depending on the version.
- To use infographics in the eval script, add the v2 image name and tag in `.env`, and add
```python
create_nvingest_collection("infographics", f"http://{milvus_hostname}:19530", sparse=sparse, gpu_search=True)

infographic_results = [[element for element in results if element['metadata']['content_metadata']['subtype'] == 'infographic'] for results in results]

write_to_nvingest_collection(infographic_results, "infographics", sparse=sparse, milvus_uri=f"http://{milvus_hostname}:19530", minio_endpoint="minio:9000")
...
```

## Checklist


client/src/nv_ingest_client/client/interface.py
```diff
@@ -392,10 +392,15 @@ def extract(self, **kwargs: Any) -> "Ingestor":
         """
         extract_tables = kwargs.pop("extract_tables", True)
         extract_charts = kwargs.pop("extract_charts", True)
+        extract_infographics = kwargs.pop("extract_infographics", True)
```
edknv: Open question: `extract_infographics` is set to True by default, but a recent test shows that there may be a performance hit with infographics (~3 pages/sec drop from ~12 pages/sec to ~9 pages/sec). One option is we could make the default `extract_infographics=False` to avoid the throughput drop in the default path, and let the user explicitly enable the inforgraphics mode if they want it.

src/nv_ingest/extraction_workflows/pdf/pdfium_helper.py
```diff
@@ -148,45 +155,59 @@ def extract_table_and_chart_images(
...
 
-            table_data = CroppedImageWithContent(
+            bbox_in_orig_coord = (
+                int(w1) - pad_width,
+                int(h1) - pad_height,
+                int(w2) - pad_width,
+                int(h2) - pad_height,
+            )
+            max_width = orig_width - 2 * pad_width
+            max_height = orig_height - 2 * pad_height
```
edknv, line 203: This fixes a bug in bounding boxes in metadata. The cropped images are correct in `cropped = crop_image(original_image, (int(w1), int(h1), int(w2), int(h2)))` but `CroppedImageWithContent(bbox=...)` was incorrect. Some users might want to use this bbox coordinates to extract the images directly from the pdf instead of using the base64 image in `table_content`. The coordinates can be used in something like:
```python
        w_orig = page.rect.width
        h_orig = page.rect.height
        loc = md["metadata"]["table_metadata"]["table_location"]        
        w_max, h_max = md["metadata"]["table_metadata"]["table_location_max_dimensions"]
        bbox = [loc[0] / w_max * w_orig, loc[1] / h_max * h_orig, loc[2] / w_max * w_orig, loc[3] / h_max * h_orig]
        pix = page.get_pixmap(clip=bbox)
```

src/nv_ingest/extraction_workflows/pdf/pdfium_helper.py
```diff
@@ -389,16 +412,19 @@ def pdfium_extractor(
...
 
             # If we want tables or charts, rasterize the page and store it
-            if extract_tables or extract_charts:
-                image, _ = pdfium_pages_to_numpy(
-                    [page], scale_tuple=(YOLOX_MAX_WIDTH, YOLOX_MAX_HEIGHT), trace_info=trace_info
+            if extract_tables or extract_charts or extract_infographics:
+                image, padding_offsets = pdfium_pages_to_numpy(
+                    [page],
+                    scale_tuple=(YOLOX_PAGE_IMAGE_PREPROC_WIDTH, YOLOX_PAGE_IMAGE_PREPROC_HEIGHT),
+                    padding_tuple=(YOLOX_PAGE_IMAGE_PREPROC_WIDTH, YOLOX_PAGE_IMAGE_PREPROC_HEIGHT),
```
edknv, line 430: Padding is necessary here to align with the above comment regarding the bounding boxes in metadata: https://github.com/NVIDIA/nv-ingest/pull/483/files#r1970386230

client/src/nv_ingest_client/client/interface.py
drobison00: @randerzander This seems like it should be opt-in functionality with the given perf hit. Thoughts?

src/nv_ingest/extraction_workflows/pdf/pdfium_helper.py
```diff
@@ -449,3 +475,24 @@ def pdfium_extractor(
         extracted_data.append(doc_text_meta)
 
     return extracted_data
+
+
+def _get_yolox_model_name(config, default_model_name="nv-yolox-page-elements-v1"):
```
drobison00: I don't think this should live in pdfium_helper. Maybe yolox.py, or something where its available to anyone else using it.

src/nv_ingest/stages/nim/infographic_extraction.py
```diff
@@ -0,0 +1,254 @@
...
+
+logger = logging.getLogger(__name__)
+
+PADDLE_MIN_WIDTH = 32
+PADDLE_MIN_HEIGHT = 32
+
+
+def _update_metadata(
+    base64_images: List[str],
+    paddle_client: NimClient,
```
drobison00, line 32: Is this supposed to be paddle_client?

src/nv_ingest/stages/nim/infographic_extraction.py
```diff
@@ -0,0 +1,254 @@
...
+            valid_indices.append(i)
+        else:
+            # Image is too small; mark as skipped.
+            results[i] = (img, None, None)
+
+    # Prepare data payloads for both clients.
+    data_paddle = {"base64_images": valid_images}
+
+    _ = worker_pool_size
+    with ThreadPoolExecutor(max_workers=2) as executor:
```
drobison00: Same question about paddle.

Also, we don't need a ThreadPoolExecutor here. The only reason we use it in places like the chart extractor is so we can run two models in parallel.

src/nv_ingest/stages/nim/infographic_extraction.py
edknv: Yes, we only need Paddle to extract test from the images. No need for another yolox model unlike table or chart extraction.

src/nv_ingest/stages/nim/infographic_extraction.py
edknv: Removed in [8bc3c7d](https://github.com/NVIDIA/nv-ingest/pull/483/commits/8bc3c7d5263d659a580db16073357115b22db0b1)

src/nv_ingest/extraction_workflows/pdf/pdfium_helper.py
edknv: Moved to yolox.py.

client/src/nv_ingest_client/client/interface.py
edknv: I went ahead and made it an opt-in in [6413b8d](https://github.com/NVIDIA/nv-ingest/pull/483/commits/6413b8d4663a650f32534b3b81e9099a544e6cdf). `extract_infographics` is False by default and users will have to explicitly turn it on for now.

PR #486, opened 2025-02-24
Adding --profile vlm for self hosted VLM NIM

randerzander:
## Checklist

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -149,6 +149,31 @@ services:
     profiles:
       - nemoretriever-parse
 
+  vlm:
+    image: ${VLM_IMAGE:-nvcr.io/nim/meta/llama-3.2-11b-vision-instruct}:${VLM_TAG:-latest}
+    ports:
+      - "8018:8000"
+    user: root
+    environment:
+      - NIM_HTTP_API_PORT=8000
+      - NIM_TRITON_LOG_VERBOSE=1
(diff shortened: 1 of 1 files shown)


randerzander:
@jdye64 helm charts will need this added as an option

@drobison00 - do internal default endpoints for build hosted NIMs need an udpate given I pointed the nv-ingest container's env var to a local NIM?

docker-compose.yaml
```diff
@@ -149,6 +149,31 @@ services:
...
+    user: root
+    environment:
+      - NIM_HTTP_API_PORT=8000
+      - NIM_TRITON_LOG_VERBOSE=1
+      - NGC_API_KEY=${NIM_NGC_API_KEY:-${NGC_API_KEY:-ngcapikey}}
+      - CUDA_VISIBLE_DEVICES=0
+      # VLM will use all available VRAM on device
+      # For more info
+      # https://docs.nvidia.com/nim/vision-language-models/latest/configuration.html
+      #- NIM_KVCACHE_PERCENT=.25
```
jdye64, line 165: Shouldn't this be uncommented for default setup?

randerzander: I've put this on a second GPU, which should have more VRAM available, since the first is full of the rest of the NIMs.

I leave that comment there as a reference for how to control VRAM usage if desired.

It'll be similar for Nemoretriever-Parse and RIVA

We'll have to explain this approach in detail in the doc pages

PR #502, opened 2025-03-03
no create when already exists

jperez999:
## Description
This PR fixes an issue with collections when you restart milvus, they seem to have some metadata attached that associates the collections with the instance of milvus that created them. So even though you have the index parameters and the same schema if you try to use the create collection api it will still say that you are trying to create a collection with a different set of parameters, instead of returning the already existing collection. This PR gives us the desired behavior where it uses the supplied collection. Now we no longer create a collection if the collection already exists (when recreate is false).

## Checklist


ChrisJar (approved):
lgtm!

PR #510, opened 2025-03-05
Add embedding nim endpoint and model name env variables to client

ChrisJar:
## Description

This introduces the `EMBEDDING_NIM_ENDPOINT` env variable to the client which will default to https://integrate.api.nvidia.com/v1 if unset or empty string, and will convert urls that don’t start with http:// or https:// to the format `http://<EMBEDDING_NIM_ENDPOINT>/v1`

## Checklist

### DIFF
--- a/client/src/nv_ingest_client/primitives/tasks/embed.py
+++ b/client/src/nv_ingest_client/primitives/tasks/embed.py
@@ -63,7 +63,7 @@ class EmbedTask(Task):
         embedding_nim_endpoint = os.getenv("EMBEDDING_NIM_ENDPOINT", "https://integrate.api.nvidia.com/v1")
         if embedding_nim_endpoint == "":
             self._embedding_nim_endpoint = "https://integrate.api.nvidia.com/v1"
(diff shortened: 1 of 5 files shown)


client/src/nv_ingest_client/primitives/tasks/embed.py
```diff
@@ -58,6 +60,17 @@ def __init__(self, text: bool = None, tables: bool = None, filter_errors: bool =
                 "'tables' parameter is deprecated and will be ignored. Future versions will remove this argument."
             )
 
+        embedding_nim_endpoint = os.getenv("EMBEDDING_NIM_ENDPOINT", "https://integrate.api.nvidia.com/v1")
+        if embedding_nim_endpoint == "":
+            self._embedding_nim_endpoint = "https://integrate.api.nvidia.com/v1"
+        elif embedding_nim_endpoint[:7] != "http://" and embedding_nim_endpoint[:8] != "https://":
```
jdye64: would prefer to see python str `.startswith('http://')` and `.startswith('https://')'` used here

jdye64 (changes requested):
Small nit for checking http and https pattern. Other that that looks good

PR #511, opened 2025-03-05
enable text embedding extraction and vdb upload for audio

edknv:
## Description
This PR adds support for text embedding extraction and vector database upload in the audio transcript extraction workflow. This enables downstream tasks to leverage embeddings for retrieval.

## Checklist

### DIFF
--- a/src/nv_ingest/stages/nim/audio_extraction.py
+++ b/src/nv_ingest/stages/nim/audio_extraction.py
@@ -2,21 +2,24 @@
 # All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-import logging
 import functools
+import logging
 import traceback
-
-import pandas as pd
 from typing import Any
 from typing import Dict
 from typing import Optional
 from typing import Tuple
 
+import pandas as pd
 from morpheus.config import Config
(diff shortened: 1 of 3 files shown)


src/nv_ingest/stages/nim/audio_extraction.py
```diff
@@ -57,7 +60,7 @@ def _update_metadata(row: pd.Series, audio_client: Any, trace_info: Dict) -> Dic
     content_metadata = metadata.get("content_metadata", {})
 
     # Only modify if content type is audio
-    if (content_metadata.get("type") != "audio") and (base64_audio in (None, "")):
+    if (content_metadata.get("type") != ContentTypeEnum.AUDIO) and (base64_audio in (None, "")):
```
drobison00: Should this be != AUDIO **or** base64_audio in (None,"")

drobison00 (approved):
Approved, although i think the and -> or change makes sense.

src/nv_ingest/stages/nim/audio_extraction.py
edknv: Thanks for catching this.

PR #527, opened 2025-03-07
set NIM_TRITON_MAX_BATCH_SIZE & NIM_TRITON_MODEL_BATCH_SIZE to defaul…

jdye64:
…ts with override option for all supported NIMs

## Checklist


docker-compose.yaml
```diff
@@ -20,6 +20,8 @@ services:
       - NIM_TRITON_LOG_VERBOSE=1
       - NGC_API_KEY=${NIM_NGC_API_KEY:-${NGC_API_KEY:-ngcapikey}}
       - CUDA_VISIBLE_DEVICES=0
+      - NIM_TRITON_MAX_BATCH_SIZE=${PAGE_ELEMENTS_BATCH_SIZE:-1}
```
randerzander: confusing I know..

only the yolox NIMs use: - NIM_TRITON_MODEL_BATCH_SIZE=1

and only the embedder NIM uses: - NIM_TRITON_MAX_BATCH_SIZE=1

jdye64:
@randerzander how about that? Look good now?

PR #528, opened 2025-03-07
Update notebook to remove store() call and use only vdb_upload() now

guilherme-pombo:
## Description
The store() call no longer works and results in error:

```
Error while processing job ID 0: /nv-ingest/data/multimodal_test.pdf
[]: failed
Failed to process the message.
Event that caused this failure: annotation::612eed31-ec4b-45d9-b370-6f34240350d4 -> Error in on_data: Failed to store extracted objects: 'DataFrame' object has no attribute 'mutable_dataframe'
```

Since it's the user's responsibility to preserve access to original content (i.e. we only keep text representations in milvus) we no longer need store() and use only vdb_upload() now.

## Checklist
- [X] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/nv-ingest/blob/main/CONTRIBUTING.md).
- [X] New or existing tests cover these changes.
- [X] The documentation is up to date with these changes.
- [X] If adjusting docker-compose.yaml environment variables have you ensured those are mimicked in the Helm values.yaml file.


jdye64 (approved):
Thanks for noting and removing this

PR #537, opened 2025-03-10
Conda client build fix

jdye64:
## Description
Adjusts the `api` and `client` conda recipes to pull from the src/version.py file rather than attempt to pull from setup.py which no longer exists for the respective projects after moving to pyproject.toml file instead.

## Checklist


conda/environments/nv_ingest_environment.yml
```diff
@@ -47,7 +48,8 @@ dependencies:
   - pip
   - pip:
       - llama-index-embeddings-nvidia
-      - nvidia-riva-client
       - opencv-python # For some reason conda cant solve our req set with py-opencv so we need to use pip
       - pymilvus>=2.5.0
       - pymilvus[bulk_writer, model]
+      - ffmpeg>=1.1
```
edknv: The package nv-ingest uses is actually `ffmpeg-python` which is already listed on line 13.

conda/packages/nv_ingest/meta.yaml
```diff
@@ -66,6 +66,9 @@ requirements:
     - unstructured-client>=0.25.9
     - uvicorn
     - wand>=0.6.10
+  pip:
+    - ffmpeg>=1.1
```
edknv: Similar comment as above, but it looks like `meta.yaml` is missing `ffmpeg-python>=0.2.0`.

PR #546, opened 2025-03-11
Add support for bmp files and treat html files as text

ChrisJar:
## Checklist


drobison00 (changes requested):
Please add a testing .bmp file to ./data so we're able to validate. You can probably just convert multimodal_test.png to multimodal_test.bmp

PR #553, opened 2025-03-12
25.03 Updates including release notes and library mode

nkmcalli:
25.03 Updates including release notes and library mode


docs/docs/extraction/images/overview-retriever.png
sosahi, line 1: can we make the diagram larger? also in dark mode it becomes pretty ineligible 
Screenshot 2025-03-11 at 11 19 51 PM

docs/docs/extraction/ngc-api-key.md
sosahi, line 1: need to update the image and remove the "AI Foundation Models and Endpoints" from the text as well as the screenshot and add "Public API Endpoints". I took an screenshot to replace this with
Screenshot 2025-03-11 at 11 26 42 PM

docs/docs/extraction/images/overview-retriever.png
nkmcalli: I made it bigger.  If they can provide a diagram with a white background instead of a transparent background that would  also help.  I can swap that in any time they provide it.

docs/docs/extraction/ngc-api-key.md
nkmcalli: Updated

PR #605, opened 2025-03-18
Update README for 25.03 release

randerzander:
…ons to release conda package, wordsmithing

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ ingestor = (
 
 print("Starting ingestion..")
 t0 = time.time()
-results = ingestor.ingest()
+results = ingestor.ingest(show_progress=True)
 t1 = time.time()
 print(f"Time taken: {t1-t0} seconds")
 
--- a/README.md
+++ b/README.md
@@ -64,10 +64,10 @@ Library mode deployment of nv-ingest requires:
 Create a fresh conda environment in which to install nv-ingest and dependencies.
 
 ```shell
-conda create -y --name nvingest python=3.10
-conda activate nvingest
-conda install -y -c rapidsai -c conda-forge -c nvidia nv_ingest=25.3.0 nv_ingest_client=25.3.0 nv_ingest_api=25.3.0
-pip install opencv-python llama-index-embeddings-nvidia pymilvus 'pymilvus[bulk_writer, model]' milvus-lite nvidia-riva-client unstructured-client
+conda create -y --name nvingest python=3.10 && \
+    conda activate nvingest && \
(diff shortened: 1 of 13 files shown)


drobison00 (approved):

README.md
```diff
@@ -49,375 +45,258 @@ NV-Ingest supports the following file types:
...
-- **extract_images** — Uses [PDFium](https://github.com/pypdfium2-team/pypdfium2/) to extract images.
-- **extract_tables** — Uses [YOLOX](https://github.com/Megvii-BaseDetection/YOLOX) to find tables and charts. Uses [PaddleOCR](https://github.com/PaddlePaddle/PaddleOCR) for table extraction, and [Deplot](https://huggingface.co/google/deplot) and CACHED for chart extraction.
-- **extract_charts** — (Optional) Enables or disables Deplot and CACHED for chart extraction.
+\begin{tabular}{ccc}
+**Animal** & **Activity** & **Place**\\
+Giraffe & Driving a car & At the beach\\
+Lion & Putting on sunscreen & At the park\\
+Cat & Jumping onto a laptop & In a home office\\
+Dog & Chasing a squirrel & In the front yard\\
+\end{tabular}
```
edknv: Nit, this looks like Eclair's output (LaTex format). Was it intentional, or do we actually want Paddle's markdown format? It looks like the quickstart doc also has a couple in LaTeX format, so I was curious.

randerzander: good catch. I had an eclair run in my milvus.db file from a prior test run

randerzander: fixed

docker-compose.yaml
```diff
@@ -40,7 +40,7 @@ services:
     runtime: nvidia
 
   graphic-elements:
-    image: ${YOLOX_GRAPHIC_ELEMENTS_IMAGE:-nvcr.io/nvidia/nemo-microservices/nemoretriever-graphic-elements-v1}:${YOLOX_GRAPHIC_ELEMENTS_TAG:-1.1}
+    image: ${YOLOX_GRAPHIC_ELEMENTS_IMAGE:-nvcr.io/nvidia/nemo-microservices/nemoretriever-graphic-elements-v1}:${YOLOX_GRAPHIC_ELEMENTS_TAG:-latest}
```
jdye64: Latest is handy for sure but do we really want to have the extra ambiguity associated with using it?

jdye64 (changes requested):
Like the more concise readme for sure.

Some linter fixes are needed so those don't cause subsequent CI failures

One debatable question about using "latest". I can see it being handy but could also see it being problematic and ambiguous at times. I prefer using explicit versions but could be swayed

docker-compose.yaml
randerzander: I'd think for main it's fine, but for the release branch we'd want a specific version

PR #618, opened 2025-03-24
Add workflow for integration tests

edknv:
## Description

This workflow sets up the structure for testing nv-ingest in library mode. It builds conda packages and prepares to run integration tests to verify library usage, mirroring how users consume the project via conda packages.

The workflow is triggered on `push` to main and `pull_request_target` events, including when the `ok-to-test` label is added.

Note: This is a skeleton workflow only -- no actual tests will run yet. Because `pull_request_target` runs from the base branch, we need to merge this PR into `main` first before any tests can actually execute.

## Checklist
- ~~[ ] If adjusting docker-compose.yaml environment variables have you ensured those are mimicked in the Helm values.yaml file.~~

### DIFF
--- a/.github/workflows/test-library-mode.yml
+++ b/.github/workflows/test-library-mode.yml
@@ -1,0 +1,86 @@
+name: Test Nv-Ingest Library Mode
+
+on:
+  pull_request_target:
+    types:
+      - opened
+      - synchronize
+      - reopened
+      # Adding 'labeled' to the list of activity types that trigger this event
+      # (default: opened, synchronize, reopened) so that we can run this
+      # workflow when the 'ok-to-test' label is added.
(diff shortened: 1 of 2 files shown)


.github/workflows/test-library-mode.yml
```diff
@@ -0,0 +1,86 @@
...
+        uses: actions/download-artifact@v4
+
+      - uses: conda-incubator/setup-miniconda@v3
+        with:
+          auto-update-conda: true
+
+      - name: Install Conda
+        shell: bash -el {0}
+        run: |
+          conda create -y --name nvingest python=3.10
```
jdye64, line 66: I think it would be nice to have this "library mode setup" in an external script at some point but its fine for now. Since we have to merge to test this running even lets iron that out first and then can move it later.

PR #627, opened 2025-03-25
Add pdf and image extraction integration tests

edknv:
## Description
Adds a few integration tests for PDF and image extraction workflows.

TODO: More stages (split, embed, etc.) and other document formats (pptx, docx, audio, etc.) which will be followed up in separate PRs.

## Checklist

### DIFF
--- a/examples/launch_libmode_and_run_ingestor.py
+++ b/examples/launch_libmode_and_run_ingestor.py
@@ -3,13 +3,14 @@ import os
 import sys
 import time
 
-from nv_ingest.framework.orchestration.morpheus.util.pipeline.pipeline_runners import (
-    PipelineCreationSchema,
-    start_pipeline_subprocess,
-)
-from nv_ingest_api.util.message_brokers.simple_message_broker import SimpleClient
-from nv_ingest_client.client import Ingestor, NvIngestClient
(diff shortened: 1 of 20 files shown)


examples/launch_libmode_and_run_ingestor.py
```diff
@@ -70,11 +75,13 @@ def main():
         # Optionally, wait a bit before starting the ingestor to ensure the pipeline is ready
         time.sleep(10)
 
+        # If the pipeline is not running, exit immediately
+        if pipeline_process.poll() is not None:
+            raise RuntimeError("Error running pipeline subprocess.")
```
edknv, line 80: This block was added because the script hangs when the pipeline generation subprocess in `start_pipeline_subprocess` fails, .e.g, when the environment variables are incorrectly set.

examples/launch_libmode_and_run_ingestor.py
```diff
@@ -70,11 +75,13 @@ def main():
...
+
         # Run ingestion after starting the pipeline
         run_ingestor()
 
-        pipeline_process.wait()
```
edknv, line 76: This line is removed because the script is not terminated without explicitly killing the process even if the ingestion succeeds.

PR #631, opened 2025-03-26
Ray orchestration prototyping

drobison00:
## Checklist


api/src/nv_ingest_api/internal/primitives/nim/model_interface/yolox.py
```diff
@@ -709,7 +709,13 @@ def postprocess_results(
             raise ValueError(f"Error in postprocessing {result.shape} and {original_image_shape}: {e}")
 
         for box, score, label in zip(bboxes, scores, labels):
-            class_name = class_labels[int(label)]
+            # TODO(Devin): Sometimes we get back unexpected class labels?
```
jdye64, line 712: odd

src/nv_ingest/framework/orchestration/ray/primitives/progress_engine.py
jdye64, line 1: I rather like this concept. What do you envision on capturing here? Would this be throughput numbers of documents or something more like system profiling metrics?

jdye64 (commented):
Lots of changes so I wanted to focus on the highest level things for now to not get bogged down in all the changes. Pretty excited about this!

- It isn't clear to me what role Redis plays now as it appeared you were using Ray queues. Can you elaborate on that?
- I saw a few place that `mrc` was still being imported. I assume that was likely just oversight but is Morpheus required in any way (other than needed for existing Morpheus pipelines of course which would continue to exist as this PR is written)
- The Ray pipeline builder was a little confusing to me. There was lots of dynamic things happening, which I assume means some cool features, but any explanation for those would be great.
- I know you have made some runs with this already. Do you care to share a snippet here for others to poke around at it? I didn't see any dependency changes so maybe a conda env file and startup script?
- What are your general thoughts about exposing each stage a little more directly to the end user? I appreciate the directory structure but what about if we offered a path for end users to more directly interact with the Ray Actors like they would a standard Python library? When I saw the structure the first thing I wanted was to be able to fire up python, import the audio_extractor, and simply invoke something like `run()` on the Ray actor (or Task whichever) on a underlying directory with a batch of .wav files for example.  I understand a pipeline gives us a nice construct to tweak and tune steps and stages but then wonder if we might be hindering certain use cases where folks want to use just individual parts of those stages? Surely there is some middle ground there? Maybe even just adding some more convenience methods gets us there?

src/nv_ingest/framework/orchestration/ray/edges/async_queue_edge.py
```diff
@@ -0,0 +1,63 @@
...
+    async def write(self, item: Any) -> bool:
+        """
+        Write an item into the edge asynchronously.
+        """
+        if self.queue.full():
+            async with self.stats_lock:
+                self.stats["queue_full_count"] += 1
+            logger.info("Queue is full. Incrementing queue_full_count.")
+        logger.info("Attempting to put item into the queue.")
+        await self.queue.put(item)
```
jperez999, line 39: put in a queue will block automatically if the queue is full. What is the motivation behind this size increment?

src/nv_ingest/framework/orchestration/ray/examples/task_source_sink_harness.py
```diff
@@ -0,0 +1,94 @@
...
+
+    # Create an instance of the MessageBrokerTaskSourceStage actor.
+    source_actor = MessageBrokerTaskSourceStage.remote(source_config, 1)
+
+    # Create an instance of the MessageBrokerTaskSinkStage actor.
+    sink_actor = MessageBrokerTaskSinkStage.remote(sink_config, 1)
+
+    # Connect the stages:
+    # The source's output edge is the queue_edge.
+    ray.get(source_actor.set_output_edge.remote(queue_edge))
```
jperez999, line 67: Based on this example is the idea to use queues to pass data between actors/stages. Or is this specifically for the message broker Tasks?

src/nv_ingest/framework/orchestration/ray/primitives/progress_engine.py
jperez999: I would also like to understand how you plan to incorporate this ProgressEngine.

src/nv_ingest/framework/orchestration/ray/util/pipeline/pid_controller.py
```diff
@@ -0,0 +1,327 @@
...
+
+import logging
+import numpy as np
+from collections import deque
+from typing import Dict, Any, Deque, Optional
+
+logger = logging.getLogger(__name__)
+
+
+class PIDController:
```
jperez999, line 34: Is the PID Controller used to create copies of stages? Is it essentially a controller for what the ray workers are supposed to be doing?

src/nv_ingest/framework/orchestration/ray/util/pipeline/vis_tools.py
```diff
@@ -0,0 +1,166 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-25, NVIDIA CORPORATION & AFFILIATES.
+# All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from typing import List
+
+
+def render_stage_box(label: str, width: int = 20) -> List[str]:
```
jperez999: Is the purpose of this module to visualize the pipeline? There are packages that do this out of the box, like graphviz. If we use that decrease the size of the code base by removing this.

jperez999 (approved):
Left some comments, would be cool to get some answers to those. Otherwise looks like a good first step.

src/nv_ingest/framework/orchestration/ray/primitives/progress_engine.py
drobison00: This was an early prototype just to mimic what we we're doing in Morpheus. In morpheus the progress engine is responsible for pulling items off the edge/queue into the stage and running them.

Using a ray.util.queue as an edge between stages, and having each stage run a background thread that pulls from the queue accomplishes the same thing and ended up being simpler overall.

src/nv_ingest/framework/orchestration/ray/edges/async_queue_edge.py
drobison00: Originally, to maintain metrics for auto-scaling. This has been simplified quite a bit with the PID controller / memory resource scaling abstractions.

src/nv_ingest/framework/orchestration/ray/examples/task_source_sink_harness.py
drobison00: Yep, basically you have [Stage] -> edge_queue -> [Stage], which allows stages to be replicated easily and independently of the edges.

src/nv_ingest/framework/orchestration/ray/util/pipeline/pid_controller.py
drobison00: PID Controller is used to decided how/when to scale stages, but takes into account the immediate proportional change, integral long term change, and in theory differential change.

Its a nice abstraction that tries to prevent oscillations or jerky behavior in scaling.

src/nv_ingest/framework/orchestration/ray/util/pipeline/vis_tools.py
drobison00: Yeah, this is random junk I was using to visualize things early on, will be removed in the near future.

api/src/nv_ingest_api/internal/primitives/nim/model_interface/yolox.py
drobison00: > odd
Yeah, I think this may be a model difference or slightly different behavior depending on the platform. I was getting labels in [0,3] on an H100 system, and [0,2] on mine, set it aside for now, because we don't use the other label index.

1. Redis is the default broker the pipeline reads from. Its not used anywhere in the pipeline, but it allows job submission to scale, and throttle independently of the pipeline.

2. I think I was using a helper function in one place. Shouldn't be used anymore.

3. The RayPipeline supports dynamic up/down scaling, and throttling, based on queue occupancy and global memory thresholds. At a base level it supports construction of the pipeline, monitoring pipeline elements, and dynamic scaling.

4. You can run the test harness (which is basically a statically defined pipeline) as `python src/nv_ingest/framework/orchestration/ray/examples/pipeline_test_harness.py` (this will kick off the Ray pipeline and start a SimpleBroker). 

Once the pipeline is running you can do something like this:

`nv-ingest-cli --batch_size=32 --doc ../datasets/bo767/**/*.pdf --client_host=localhost --client_port=7671 --client_type=simple --output_directory=../datasets/processed/scratch_synch_embed --task='extract:{"document_type": "pdf", "extract_tables": "True", "extract_charts": "True", "extract_infographics": "True", "extract_images": "True", "extract_text": "True"}' --task='embed' --task='split'`

src/nv_ingest/framework/orchestration/ray/edges/async_queue_edge.py
jperez999: Doesnt ray handle this internally https://docs.ray.io/en/latest/cluster/vms/user-guides/configuring-autoscaling.html, or does the PID serve a different purpose?

src/nv_ingest/framework/orchestration/ray/examples/task_source_sink_harness.py
jperez999: Based on my understanding, we are not trying to use the DAG pattern here the Producer-Consumer pattern, correct?

src/nv_ingest/framework/orchestration/ray/util/pipeline/pid_controller.py
jperez999: Doesnt ray handle this internally https://docs.ray.io/en/latest/cluster/vms/user-guides/configuring-autoscaling.html, or does the PID serve a different purpose?

src/nv_ingest/framework/orchestration/ray/examples/task_source_sink_harness.py
drobison00: This could be extended to being more of a DAG, and I expect we will for cases where we have a heterogeneous set of input files that can each be independently extracted (docx, pdf, etc..). But yes, currently, its a linear pipeline with a producer -> consumer relationship between each stage.

src/nv_ingest/framework/orchestration/ray/util/pipeline/pid_controller.py
drobison00: I think the page you linked is about auto-scaling nodes (like VMs/hardware), not really Actor replicas. That's something we could eventually look into. But, in general I think simple replication of the ingest service itself is probably enough. The work is embarrassingly parallel at the job level, and we can leverage the message broker as a load balancing mechanism.

I believe ray-serve has some auto-scaling features; but, it has a fairly different focus and I don't really think we want to use it at the moment. Even then, the ray-serve autoscaler doesn't provide the ability (that I'm aware of) to do things like the edge wire up at creation time, or to start the background processing thread, it just creates a new actor instance, which doesn't give as much control as the PID, that uses instantaneous and historic edge occupancy stats, and the monitoring process which supplies the appropriate edge and starts the progress loop on the actor.

Another issue is that we have somewhat un-predictable memory usage (depends on the document content), and the current PID / Resource allocation mechanisms will throttle or down-scale based on observed memory usage patterns, which gives us more flexibility than we would have with just Ray's auto scaling.

PR #636, opened 2025-04-01
Add integration test for audio extraction

edknv:
## Description
This PR introduces an integration test for audio extraction that validates transcription output from a `.wav` file using the "audio" extract method. The test ensures that audio input is correctly processed and returns the expected transcript.

## Checklist

### DIFF
--- a/tests/integration/test_extract_audio.py
+++ b/tests/integration/test_extract_audio.py
@@ -8,7 +8,8 @@ from nv_ingest_client.client import Ingestor
 from nv_ingest_client.client import NvIngestClient
 
 
-def test_images_extract_only(
+@pytest.mark.integration
+def test_audio_extract_only(
     pipeline_process,
 ):
     client = NvIngestClient(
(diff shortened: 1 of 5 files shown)


tests/integration/test_extract_audio.py
```diff
@@ -0,0 +1,38 @@
...
+import sys
+import time
+
+import pytest
+from nv_ingest_api.util.message_brokers.simple_message_broker import SimpleClient
+from nv_ingest_client.client import Ingestor
+from nv_ingest_client.client import NvIngestClient
+
+
+def test_images_extract_only(
```
drobison00: For anything like this can you decorate with `@pytest.mark.integration`; so that it has to intentionally be run.

e.g. `pytest -m integration ./tests/integration`

drobison00 (changes requested):

tests/integration/test_extract_audio.py
edknv: Added in [74b081a](https://github.com/NVIDIA/nv-ingest/pull/636/commits/74b081af2f8a8862f1a244033141dc60fa17592b)

PR #654, opened 2025-04-11
Provide option to pass full url and headers to RestClient

jeffreyscarpenter:
## Description
This PR modifies the RestClient to allow the full URL of the nv-ingest endpoint to be passed, as an alternative to the hostname/port parameters. This is intended as a minimally intrusive, backward compatible change that can flexibly handle a variety of URL scheme, hostname, and port options. 

For example, for our application we intend to use a base URLs of the form: https://mysite.com/path. Given this base URL, the client would then be able to call https://mysite.com/path/v1/submit_job and https://mysite.com/path/v1/submit_job.

There is no way to get the current hostname/port parameters to correctly handle this type of URL, as any path after the base URL gets dropped. (It also doesn't seem to handle https:// URLs or non-default port numbers very well.) 

The PR also allows the user to specify HTTP headers such as an authentication token that are applied on calls to the backend nv-ingest service.

## Checklist

### DIFF
--- a/api/src/nv_ingest_api/util/service_clients/rest/rest_client.py
+++ b/api/src/nv_ingest_api/util/service_clients/rest/rest_client.py
@@ -215,11 +215,11 @@ class RestClient(MessageBrokerClientBase):
 
                 # Fetch using streaming response
                 with requests.get(
-                        url,
-                        timeout=(30, 600),
-                        stream=True,
-                        headers=headers,
-                        auth=self._auth,
+                    url,
+                    timeout=(30, 600),
+                    stream=True,
+                    headers=headers,
+                    auth=self._auth,
                 ) as result:
                     response_code = result.status_code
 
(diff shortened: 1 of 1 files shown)


jeffreyscarpenter:
Sample usage via Ingestor:

```
            ingestor = (
                Ingestor(
                    message_client_kwargs={
                        "base_url": self.base_url,
                        "headers": {"Authorization": f"Bearer {self.api_key}"},
                        "max_retries": 3,
                        "timeout": 60,
                    }
                )
```

api/src/nv_ingest_api/util/service_clients/rest/rest_client.py
```diff
@@ -109,6 +110,14 @@ def __init__(
         self._submit_endpoint = "/v1/submit_job"
         self._fetch_endpoint = "/v1/fetch_job"
 
+        if "base_url" in kwargs:
```
jdye64, line 113: Appreciate this log message

jdye64 (approved):
@jeffreyscarpenter thanks for the PR. This makes perfect sense to me.

jdye64:
@jeffreyscarpenter one nit before merging. Can you please resolve the linter issues?

jeffreyscarpenter:
> @jeffreyscarpenter one nit before merging. Can you please resolve the linter issues?

Ok, I just see the one issue, submitted PR. Let me know if you see anything else

jeffreyscarpenter:
@jdye64 apologies for churn, found a second linter issue. I managed to run the check locally so I think I've got everything now.

PR #658, opened 2025-04-14
Add metadata file support and example/documentation

jperez999:
## Description
Adds the ability to ingest directly from a supported file. File types that are support are: json, csv, parquet. If we need another file types supported, they can be added as they are requested. Currently, all file reads are done with default settings, if we find that users require the ability to send custom values to args and kwargs we will need to add that support in. 

## Checklist


ChrisJar (approved):
Lgtm!

PR #660, opened 2025-04-15
Set NIM_TRITON_MODEL_BATCH_SIZE = 1 for all NIMs by default

jdye64:
## Description
Set NIM_TRITON_MODEL_BATCH_SIZE = 1 for all NIMs by default. Values can be overridden at deploy time by using `--set {NIM_NAME}.env.NIM_TRITON_MODEL_BATCH_SIZE=X`

## Checklist
- [] New or existing tests cover these changes.

### DIFF
--- a/helm/values.yaml
+++ b/helm/values.yaml
@@ -473,6 +473,7 @@ milvus:
       storageClass: null
   minio:
     mode: standalone
+    bucketName: a-bucket
     persistence:
       size: 10Gi
       storageClass: null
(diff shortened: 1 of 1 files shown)


drobison00:
Whats the motivation here?

jdye64:
To match the logic we currently have in `docker-compose.yaml`

This is also required for time-slicing or any other device sharing approaches offered by the nvidia-gpu-operator. Without setting this the gpu-operator will attempt to schedule a pod per GPU requiring a users cluster to require at least 8 gpus

Users can always choose to override those values at deploy time if they so wish

jperez999 (approved):

PR #670, opened 2025-04-21
Update ulimit troubleshooting section

nkmcalli:
Update ulimit troubleshooting section

### DIFF
--- a/docs/docs/extraction/quickstart-guide.md
+++ b/docs/docs/extraction/quickstart-guide.md
@@ -405,8 +405,10 @@ You can specify multiple `--profile` options.
 
 ## Troubleshooting
 
-In rare cases, when you run a job you might an see an error similar to `max open file descriptor` or `can't start new thread`. 
-These errors occur when the open file descriptor limit for your service user account is too low.
+### Too Many Open Files Error
+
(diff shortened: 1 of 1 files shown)


docs/docs/extraction/quickstart-guide.md
```diff
@@ -405,9 +405,13 @@ You can specify multiple `--profile` options.
 
 ## Troubleshooting
 
-In rare cases, when you run a job you might an see an error similar to `max open file descriptor`. 
-This error is related to number of active jobs and cores available on your computer. 
-To resolve this issue, raise the maximum number of open file descriptors by using the following [ulimit](https://ss64.com/bash/ulimit.html) command.
+In rare cases, when you run a job you might an see an error similar to `max open file descriptor` or `can't start new thread`.
```
drobison00: These should be broken out into two separate suggestions.

`ulimit -n XXX` will address the open files limits
`ulimit -u YYY` will  address the max user processes

nkmcalli: Okay, updated!  The separate sections don't necessarily make sense now, since there is so much duplication, but eventually we might move this to a single dedicated troubleshooting page with other troubleshooting items, in which case the separate sections will be better.  Please check that what I have is correct though.

drobison00 (approved):

PR #672, opened 2025-04-22
Fix streaming and threshold

jperez999:
## Description
This PR changes how streaming is auto activated. It now checks the count of how many elements (chunks) are processed. The threshold has been increased to 50 given this change. And we also plumbed the stream parameter so you can activate streaming insert to milvus regardless of the number of elements.

## Checklist

### DIFF
--- a/client/src/nv_ingest_client/util/milvus.py
+++ b/client/src/nv_ingest_client/util/milvus.py
@@ -223,6 +223,7 @@ class MilvusOperator:
         meta_dataframe: Union[str, pd.DataFrame] = None,
         meta_source_field: str = None,
         meta_fields: list[str] = None,
+        stream: bool = False,
         **kwargs,
     ):
         """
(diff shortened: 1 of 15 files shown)


client/src/nv_ingest_client/util/milvus.py
```diff
@@ -912,10 +917,11 @@ def write_to_nvingest_collection(
     access_key: str = "minioadmin",
     secret_key: str = "minioadmin",
     bucket_name: str = "a-bucket",
-    threshold: int = 10,
+    threshold: int = 50,
```
randerzander:
```suggestion
    threshold: int = 1000,
```

when testing w/ bo20, there are 812 chunks. with stream=True, total runtime goes from 114 to 76s

Seems like the threshold should be higher, which gives a nice perf lift for smaller ingest jobs

randerzander (changes requested):

edknv (approved):
Changes look good to me. I think we can ignore the failing integration test; I don't see how this PR fails the integration test, and it's probably due to some unrelated issues with the build endpoint. However, there is one unit test that is failing and needs updating before merging.

PR #694, opened 2025-04-29
Grpc embed

jperez999:
## Description
This PR adds two methods for connecting to the NIM microservices. Http and GRPC. This allows for standardization in access to NIMs from the client. These tools should be used whereever connections to NIMs are required from within the client. 

## Checklist


api/src/nv_ingest_api/internal/primitives/nim/model_interface/text_embedding.py
```diff
@@ -20,25 +21,18 @@ def name(self) -> str:
         """
         return "Embedding"
 
-    def prepare_data_for_inference(self, data: Dict[str, Any]) -> Dict[str, Any]:
+    def prepare_data_for_inference(self, data: List[dict]) -> List[str]:
```
drobison00: Type hints here look off.

We're still treating data as a dictionary in the function body, so I don't think it should be a List[dict], and we're returning a dictionary, so List[str] doesn't look right either.

api/src/nv_ingest_api/internal/primitives/nim/model_interface/text_embedding.py
```diff
@@ -20,25 +21,18 @@ def name(self) -> str:
...
             raise KeyError("Input data must include 'prompts'.")
-        # Ensure the prompts are in list format.
         if not isinstance(data["prompts"], list):
             data["prompts"] = [data["prompts"]]
-        return data
+        return {"prompts": data["prompts"]}
 
     def format_input(
-        self, data: Dict[str, Any], protocol: str, max_batch_size: int, **kwargs
+        self, data: List[str], protocol: str, max_batch_size: int, **kwargs
```
drobison00: List[str] looks incorrect here as well, since the data being passed in here is basically what was returned from prepare_data_for_inference above.

api/src/nv_ingest_api/internal/primitives/nim/nim_client.py
```diff
@@ -238,16 +238,22 @@ def _grpc_infer(self, formatted_input: np.ndarray, model_name: str) -> np.ndarra
...
+        outputs = [grpcclient.InferRequestedOutput(output_name) for output_name in output_names]
+        response = self.client.infer(
+            model_name=model_name, parameters=parameters, inputs=[input_tensors], outputs=outputs
+        )
         logger.debug(f"gRPC inference response: {response}")
 
         # TODO(self.client.has_error(response)) => raise error
-
-        return response.as_numpy("output")
+        return [response.as_numpy(output.name()) for output in outputs]
```
drobison00, line 256: Do we need something like this to preserve backwards compatibility?

```suggestion
        if (len(outputs) == 1):
            return response.as_numpy(outputs[0].name())
        else:
            return [response.as_numpy(output.name()) for output in outputs]
```

jperez999: No this preserves the backward compatability because we encompass "output", the expected response in a list as well. https://github.com/NVIDIA/nv-ingest/pull/694/files#diff-1de66883d0f1ef734745dfde931907c390640a70e68b4e60684ebd69e707a6b3R242 So it handles the default use case.

api/src/nv_ingest_api/internal/primitives/nim/model_interface/text_embedding.py
jperez999: fixed

api/src/nv_ingest_api/internal/primitives/nim/model_interface/text_embedding.py
jperez999: fixed

drobison00 (approved):

PR #700, opened 2025-05-01
Indexes pymilvus update

jperez999:
## Description
This PR fixes pymilvus 2.5.7+, so that we are not dependent on _indexes parameter for IndexParams when not available. We also maintain backwards compatibility. Now we can continue to use the latest version of pymilvus in the client.

## Checklist


jdye64 (approved):

jdye64:
@jperez999 should this also be added to the 25.4.1 branch?

@sosahi what do you think?

If so @jperez999 can you please open another PR with the same changes against the 25.4.1 branch. That is the branch I will be basing 25.4.2 off of.

edknv (approved):

PR #707, opened 2025-05-04
set recreate default to false

FengJi2021:
## Description
Set recreate default to False 
Link to issue
https://github.com/NVIDIA/nv-ingest/issues/706

## Checklist
- [X] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/nv-ingest/blob/main/CONTRIBUTING.md).
- [X] New or existing tests cover these changes.
- [X] The documentation is up to date with these changes.
- [X] If adjusting docker-compose.yaml environment variables have you ensured those are mimicked in the Helm values.yaml file.


jperez999:
This may cause issues for people running this multiple times with the same collection name. Are we sure we want to do this? If we are, the PR lgtm.

jperez999 (approved):

PR #718, opened 2025-05-08
Fix README TOC Quickstart link

jacobtomlinson:
Small tweak to fix the README TOC. Looks like the Quickstart heading got renamed in #605 but the TOC didn't. This PR just gets the link working again.


jdye64 (approved):
thanks for catching this @jacobtomlinson

PR #727, opened 2025-05-12
Add test for NIM client single vs multiple response grpc

jperez999:
## Description
This PR provides a unit test for grpc, where if you have only one output, it is expected that the output is not in a list otherwise it should be a list.

## Checklist


edknv:
Possibly a duplicate of https://github.com/NVIDIA/nv-ingest/pull/726?

jperez999:
@edknv Yes, I swapped it to just adding a test. In case someone tries to make another change to the NIM Client in this regard.

PR #729, opened 2025-05-13
Removes lingering references to Morpheus after the switch to Ray

jdye64:
## Description
Removes lingering references to Morpheus after the switch to Ray.

## Checklist


randerzander:
run integration tests

Dockerfile
```diff
@@ -6,7 +6,7 @@
 ARG BASE_IMG=nvcr.io/nvidia/cuda
 ARG BASE_IMG_TAG=12.5.1-base-ubuntu22.04
 
-# Use NVIDIA Morpheus as the base image
+# Use NVIDIA cuda as the base image
```
drobison00, line 9: Just to note: we should probably change this at some point  too, if we build from a stripped down ubuntu22.04 we'll probably be a lot lighter than with the cuda container.

api/src/nv_ingest_api/internal/extract/image/chart_extractor.py
```diff
@@ -27,7 +27,7 @@
 PADDLE_MIN_WIDTH = 32
 PADDLE_MIN_HEIGHT = 32
 
-logger = logging.getLogger(f"morpheus.{__name__}")
+logger = logging.getLogger(f"ray.{__name__}")
```
drobison00, line 30: I think this snuck in.

```suggestion
logger = logging.getLogger(f"{__name__}")
```

src/nv_ingest/framework/orchestration/morpheus/util/pipeline/pipeline_runners.py
```diff
@@ -62,9 +62,6 @@ class PipelineCreationSchema(BaseModel):
     message_client_port: str = "7671"
```
drobison00: One overall comment on anything under `framework/orchestration/morpheus`

These shouldn't be updated to anything else, they're meant to be Morpheus specific. If we want to remove all of the associated code, we can prune the full orchestration branch.

tests/service_tests/modules/injectors/test_metadata_injection.py
```diff
@@ -6,6 +6,8 @@
 import pandas as pd
 import pytest
 
+pytest.importorskip("mrc")
+
 from nv_ingest.framework.orchestration.morpheus.modules.injectors.metadata_injector import on_data
```
drobison00, line 11: We should figure this out too. If we're keeping the old morpheus code, we should leave its unit tests in place. If we're going to strip all of it out, then we should remove tests for Morpheus specific construts.

drobison00 (changes requested):
Generally ok, except for the changes under orchestration/Morpheus. We should either remove all those files, or leave them as is along with their unit tests.

Dockerfile
jdye64: Big fan of this. Will do in a follow up PR.

src/nv_ingest/framework/orchestration/morpheus/util/pipeline/pipeline_runners.py
jdye64: I'm going to remove it all for now. Since morpheus dependency is now removed leaving it doesn't really buy us anything.

PR #747, opened 2025-05-15
Update readme intro to use docs intro with full name

nkmcalli:
Update readme intro to use docs intro with full name

PR #755, opened 2025-05-16
Update readme intro to use docs intro with full name

nkmcalli:
Update readme intro to use docs intro with full name

Duplicate changes from #747 (main) into release/25.4.2


drobison00 (approved):
Looks fine; the driver mismatch error on the integration tests is spurious.

Issue #766, opened 2025-05-20
[FEA]: Add support for creating lambda stages


drobison00:
### Is this a new feature, an improvement, or a change to existing functionality?

New Feature

### How would you describe the priority of this feature request

Significant improvement

### Please provide a clear description of problem this feature solves

Introduce a utility function that wraps a user-defined Callable[[ControlMessage], ControlMessage] as a RayActorStage. This enables ad hoc logic injection into the DAG without boilerplate subclassing.

### Describe the feature, and optionally a solution or implementation and any alternatives

```python
def wrap_callable_as_stage(fn: Callable[[ControlMessage], ControlMessage]) -> RayActorStage:
    class CallableWrapperStage(RayActorStage):
        def on_data(self, control_message: ControlMessage) -> ControlMessage:
            try:
                return fn(control_message)
            except Exception as e:
                self._logger.exception(f"Callable raised an error: {e}")
                return control_message  # or optionally None, based on design
    return CallableWrapperStage()
```

```python
def my_transform(cm: ControlMessage) -> ControlMessage:
    payload = cm.payload()
    updated_payload = do_work(payload)
   cm.payload(updated_payload)
    return cm

transform_stage = wrap_callable_as_stage(my_transform)
```

### Additional context

```mermaid
flowchart LR
    A[Input ControlMessage] --> B[wrap_callable_as_stage]
    B --> C[User Callable<br/>cm -> cm]
    C --> D[Transformed ControlMessage]
```

PR #768, opened 2025-05-20
Issue #766 -- Add lambda stage constructors

drobison00:
## Checklist


nkmcalli:
Does this need to go out for the current release?  If not, my suggestion is to move the doc to a separate PR from the code.  I'd like to do an edit pass and I don't want to block your code PR while we do that.  

My biggest suggestion is that we want to move the summary table and the pro / cons to the top of the page, before you explain how to do it.  First the reader uses the summary and the pros / cons to decide if they want it, then they keep reading to learn how to do it.  I can make that change in a PR or a commit if you want.

drobison00:
@nkmcalli It doesn't need to go in the current release. Happy to put the doc in a separate PR.

edknv (approved):

PR #696, opened 2025-04-30
Treat json, md, and sh filetypes as text

ChrisJar:
## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/nv-ingest/blob/main/CONTRIBUTING.md).
- [ ] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.
- [ ] If adjusting docker-compose.yaml environment variables have you ensured those are mimicked in the Helm values.yaml file.

PR #770, opened 2025-05-21
Add multiple more text-like filetypes

jarmak-nv:
## Description
Small PR to add other text-likes, similar to:
- #696  

### Enhancements to document type support:
* Expanded the `text_documents` fixture in `client/client_tests/client/test_interface.py` to include new file types: `.log`, `.xml`, `.yaml`, `.yml`, `.ini`, `.conf`, `.cfg`, `.tex`, `.rst`, and `.adoc`.
* Updated `_DEFAULT_EXTRACTOR_MAP` in `client/src/nv_ingest_client/primitives/tasks/extract.py` to map the new file types to appropriate extractors. For example, `.xml` is mapped to `lxml`, and `.yaml`/`.yml` are mapped to `txt`.
* Modified the `extract_method_must_be_valid` method in `client/src/nv_ingest_client/primitives/tasks/extract.py` to skip validation for the newly added text-like file types.

### New test files for added document types:
* Added test files for `.adoc`, `.cfg`, `.conf`, `.ini`, `.rst`, `.tex`, `.xml`, `.yaml`, and `.yml` in the `data/` directory to validate processing of these formats. Each file contains representative content for its respective format. [[1]](diffhunk://#diff-ec3413761473712fe4b6f41c1bfe016b05ef4bc35475d3113ffdf53a4c4719a2R1-R14) [[2]](diffhunk://#diff-dd59e6a73e265a5b80332ddd56c298d7cc1332d8f14657d581439e2add505074R1-R14) [[3]](diffhunk://#diff-6739a5e1d0f02f7b75c794dc2295ad23ef8a693c3723d07ad67caf1be8c9f986R1-R21) [[4]](diffhunk://#diff-5c1efe7de2984a0f6b1754b91193d8007a79d2be5a7573f10c3e00f4fa726546R1-R5) [[5]](diffhunk://#diff-e89f834b4744ac184ed19f7f4f031c8702ff0fc243a5c3d70351fd2d428a2575R1-R23) [[6]](diffhunk://#diff-69fd842f416e4936904cd29f38aca338c9acb5dbb18ed935bc4cf52f2aba6297R1-R10) [[7]](diffhunk://#diff-fbffdf5701b9269ce114bf08d905cd90f5ee6ab57137dfba38ad7b167d541d44R1-R21) [[8]](diffhunk://#diff-6463e9ee63e3417f6cfc475404191965edee7dd0085997e5f36bed1dcc8aeab5R1-R5) [[9]](diffhunk://#diff-6cb558fa2a419c86e53ea7e35cb9dd2f178a697556b70ec6bb56a9a6c5b4902bR1-R8)

## Checklist
- [x] ~If adjusting docker-compose.yaml environment variables have you ensured those are mimicked in the Helm values.yaml file.~

### DIFF
--- a/client/client_tests/client/test_interface.py
+++ b/client/client_tests/client/test_interface.py
@@ -47,7 +47,23 @@ def documents():
 
 @pytest.fixture
 def text_documents():
-    return ["data/test.txt", "data/test.html", "data/test.json", "data/test.md", "data/test.sh"]
+    return [
+        "data/test.txt",
+        "data/test.html",
+        "data/test.json",
+        "data/test.md",
+        "data/test.sh",
+        "data/test.log",
+        "data/test.xml",
+        "data/test.yaml",
+        "data/test.yml",
+        "data/test.ini",
+        "data/test.conf",
+        "data/test.cfg",
+        "data/test.tex",
+        "data/test.rst",
+        "data/test.adoc",
+    ]
 
 
 @pytest.fixture
--- a/client/src/nv_ingest_client/primitives/tasks/extract.py
+++ b/client/src/nv_ingest_client/primitives/tasks/extract.py
@@ -32,26 +32,41 @@ ADOBE_CLIENT_ID = os.environ.get("ADOBE_CLIENT_ID", None)
 ADOBE_CLIENT_SECRET = os.environ.get("ADOBE_CLIENT_SECRET", None)
 
 _DEFAULT_EXTRACTOR_MAP = {
+    # Image files
     "bmp": "image",
-    "csv": "pandas",
-    "docx": "python_docx",
-    "excel": "openpyxl",
-    "html": "txt",
     "jpeg": "image",
     "jpg": "image",
-    "parquet": "pandas",
-    "pdf": "pdfium",
     "png": "image",
-    "pptx": "python_pptx",
-    "text": "txt",
     "tiff": "image",
-    "txt": "txt",
-    "xml": "lxml",
-    "mp3": "audio",
-    "wav": "audio",
+    # Text files
+    "html": "txt",
     "json": "txt",
     "md": "txt",
     "sh": "txt",
+    "text": "txt",
+    "txt": "txt",
+    "log": "txt",
+    "yaml": "txt",
+    "yml": "txt",
+    "ini": "txt",
+    "conf": "txt",
+    "cfg": "txt",
+    "tex": "txt",
+    "rst": "txt",
+    "adoc": "txt",
+    # Data files
+    "csv": "pandas",
+    "parquet": "pandas",
+    "excel": "openpyxl",
+    # Document files
+    "docx": "python_docx",
+    "pdf": "pdfium",
+    "pptx": "python_pptx",
+    # Audio files
+    "mp3": "audio",
+    "wav": "audio",
+    # XML files
+    "xml": "lxml",
 }
 
 _Type_Extract_Method_PDF = Literal[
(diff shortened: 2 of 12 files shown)


client/src/nv_ingest_client/primitives/tasks/extract.py
```diff
@@ -32,26 +32,41 @@
...
+    "yml": "txt",
+    "ini": "txt",
+    "conf": "txt",
+    "cfg": "txt",
+    "tex": "txt",
+    "rst": "txt",
+    "adoc": "txt",
+    # Data files
+    "csv": "pandas",
+    "parquet": "pandas",
```
drobison00, line 59: Should probably discuss the intent with Parquet/excel, since those files have binary/structural data that we may not want to use directly. Maybe dump both of them to CSV first?

@edknv I know there is a follow on task to make all this cleaner and differentiate document types that do and do not need extractions. I think this starts to motivate that a bit more, but also maybe we need to consider some type of ConvertToText stage that acts as a default extractor, and lets us remove structural data from files like Parquet before embedding.

client/src/nv_ingest_client/primitives/tasks/extract.py
```diff
@@ -32,26 +32,41 @@
...
+    "excel": "openpyxl",
+    # Document files
+    "docx": "python_docx",
+    "pdf": "pdfium",
+    "pptx": "python_pptx",
+    # Audio files
+    "mp3": "audio",
```
drobison00, line 66: Have you tested with mp3s? I can't remember if we have an extract path for this in the audio extractor stage.

client/src/nv_ingest_client/primitives/tasks/extract.py
```diff
@@ -32,26 +32,41 @@
...
+    "ini": "txt",
+    "conf": "txt",
+    "cfg": "txt",
+    "tex": "txt",
+    "rst": "txt",
+    "adoc": "txt",
+    # Data files
+    "csv": "pandas",
+    "parquet": "pandas",
+    "excel": "openpyxl",
```
drobison00, line 60: We don't have an extraction stage for this right now, would need to be added before we make excel a supported doctype.

client/client_tests/client/test_interface.py
```diff
@@ -47,7 +47,23 @@ def documents():
 
 @pytest.fixture
 def text_documents():
-    return ["data/test.txt", "data/test.html", "data/test.json", "data/test.md", "data/test.sh"]
+    return [
+        "data/test.txt",
```
drobison00, line 51: Will want to add mp3, csv, excel, and parquet here for completeness.

client/src/nv_ingest_client/primitives/tasks/extract.py
```diff
@@ -32,26 +32,41 @@
...
+    "yaml": "txt",
+    "yml": "txt",
+    "ini": "txt",
+    "conf": "txt",
+    "cfg": "txt",
+    "tex": "txt",
+    "rst": "txt",
+    "adoc": "txt",
+    # Data files
+    "csv": "pandas",
```
drobison00, line 58: We don't have a pandas extraction stage for this currently, and its not 100% what form we would be extracting to. Should discuss the desired outcome.

drobison00 (changes requested):

client/src/nv_ingest_client/primitives/tasks/extract.py
jarmak-nv: Yeah - tbh; I don't have any intent around the items that are handled by `pandas`/`openpyxl` - they're currently in there on `main`:
https://github.com/NVIDIA/nv-ingest/blob/2a0918d618d9c7190351ac8e9c88267e719ab8ed/client/src/nv_ingest_client/primitives/tasks/extract.py#L36
https://github.com/NVIDIA/nv-ingest/blob/2a0918d618d9c7190351ac8e9c88267e719ab8ed/client/src/nv_ingest_client/primitives/tasks/extract.py#L38
https://github.com/NVIDIA/nv-ingest/blob/2a0918d618d9c7190351ac8e9c88267e719ab8ed/client/src/nv_ingest_client/primitives/tasks/extract.py#L42

 I was just cleaning up the ordering to make it easier to group/understand what fixtures apply etc. Agree that structured data is a special case - handling as text may be better than nothing, but also might imply a level of support for structured data that isn't there.

client/src/nv_ingest_client/primitives/tasks/extract.py
jarmak-nv: See comment above for links, they're currently there on `main` - how are they handled right now?

jarmak-nv:
Added a reply around the pandas/excel/mp3 stage but in general my comment is those aren't new from this PR.

Happy remove them in this PR though if they aren't supported yet.

PR #800, opened 2025-06-02
Add dataloader

jperez999:
## Description
Adds a dataloader, currently support audio. This DL will take a file or group of files and go through them and check if they are larger than the user defined `split_interval` and splits them to chunks of split_interval or smaller. Then it loads the chunks as bytes into a queue to be used by the dataloader for iterating.

## Checklist


api/src/nv_ingest_api/util/dataloader/dataloader.py
```diff
@@ -0,0 +1,177 @@
...
+
+        path_file = Path(input_path)
+        file_name = path_file.stem
+        suffix = path_file.suffix
+        output_dir = Path(output_dir)
+        output_dir.mkdir(parents=True, exist_ok=True)
+        output_pattern = output_dir / f"{file_name}_chunk_%04d{suffix}"
+        num_splits = 0
+        try:
+            with fsspec.open(input_path, "rb", encoding="utf-8") as f_in:
```
drobison00:
```suggestion
            with fsspec.open(input_path, "rb") as f_in:
```

api/src/nv_ingest_api/util/dataloader/dataloader.py
```diff
@@ -0,0 +1,177 @@
...
+        else:
+            files = [path]
+        for file in files:
+            if thread_stop:
+                return
+            with open(file, "rb") as f:
+                queue.put(f.read())
+        files_completed.extend(files)
+    queue.put(StopIteration)
+    return files_completed
```
drobison00: Having process_data return something here is a little confusing, since its assuming it already has a queue and will be in a thread, in which case any return values will be immediately discarded.

api/src/nv_ingest_api/util/dataloader/dataloader.py
```diff
@@ -0,0 +1,177 @@
...
+
+
+class LoaderInterface(ABC):
+
+    @abstractmethod
+    def split(self, input_path: str, output_dir: str, split_interval: int = 0):
+        pass
+
+
+def _probe(filename="pipe:", cmd="ffprobe", format=None, file_handle=None, timeout=None, **kwargs):
```
drobison00: Is it possible to do this without calling out to a subprocess? This opens us up to situations where we're more susceptible to deadlocks/hangs that are hard to diagnose, zombie processes, etc..

api/src/nv_ingest_api/util/dataloader/dataloader.py
```diff
@@ -0,0 +1,177 @@
...
+        self.thread_stop = False
+        self.queue = queue.Queue(size)
+        self.paths = [Path(path) for path in paths]
+        self.output_dir = output_dir
+        self.split_interval = split_interval
+        self.chunk_count = 0
+        self.interface = interface
+        self.files_completed = []
+
+    def __getitem__(self, index: int):
```
drobison00: It would be good to support indexing into a split file vs just iterating across it. This is a fairly unexpected behavior for `__getitem__`.

api/src/nv_ingest_api/util/dataloader/dataloader.py
```diff
@@ -0,0 +1,177 @@
...
+            args=(
+                self.queue,
+                self.interface,
+                self.paths,
+                self.output_dir,
+                self.thread_stop,
+                self.split_interval,
+                self.files_completed,
+            ),
+            daemon=True,
```
drobison00: This should be more RAII-like. Could make a best erffort cleanup in `__del__`  and/or `__exit__`  , but we shouldn't rely on the user to know they have to call stop or they'll leak a thread instance.

api/src/nv_ingest_api/util/dataloader/dataloader.py
```diff
@@ -0,0 +1,177 @@
...
+                (
+                    ffmpeg.input(
+                        "pipe:",
+                    )
+                    .output(str(output_pattern), f="segment", segment_time=segment_time, c="copy", map="0")
+                    .run(capture_stdout=True, capture_stderr=True, input=f_in.read())
+                )
+        except ffmpeg.Error as e:
+            logging.error(f"FFmpeg error for file {input_path}: {e.stderr.decode()}")
+        files = [str(output_dir / f"{file_name}_chunk_{i:04d}{suffix}") for i in range(int(num_splits))]
```
drobison00: Need some way to default, maybe optionally, clean up any files created by this process when the DataLoader goes out of scope, or we encounter some kind of unexpected processing error.

api/src/nv_ingest_api/util/dataloader/dataloader.py
```diff
@@ -0,0 +1,177 @@
...
+            if thread_stop:
+                return
+            with open(file, "rb") as f:
+                queue.put(f.read())
+        files_completed.extend(files)
+    queue.put(StopIteration)
+    return files_completed
+
+
+class DataLoader:
```
drobison00: One thing I would like for this, is to provide more information to someone using the object about what it's backed by. Being able to query things like runtime in seconds, frames, sample rate, etc.. and subsequently seek or chunk based on that information would give more flexibility for users to load and use their data. Like the `get_metadata()` item in https://github.com/NVIDIA/nv-ingest/issues/713

api/src/nv_ingest_api/util/dataloader/dataloader.py
```diff
@@ -0,0 +1,177 @@
...
+def process_data(
+    queue: queue.Queue,
+    interface: LoaderInterface,
+    paths: list[str],
+    output_dir: str,
+    thread_stop: threading.Event,
+    split_interval: int = 450,
+    files_completed: list[str] = None,
+):
+    for path in paths:
```
drobison00: We probably want to wrap this with a try/except block since any failures won't be propagated back to the main thread and it may deadlock waiting on the queue. My intuition tells me a single failure in any file should result in raising a something like a FileNotFound or RuntimeError back through the queue,

api/src/nv_ingest_api/util/dataloader/dataloader.py
```diff
@@ -0,0 +1,177 @@
...
+        communicate_kwargs["timeout"] = timeout
+    if file_handle:
+        communicate_kwargs["input"] = file_handle
+    out, err = p.communicate(**communicate_kwargs)
+    if p.returncode != 0:
+        raise ffmpeg._run.Error("ffprobe", out, err)
+    return json.loads(out.decode("utf-8"))
+
+
+class MediaInterface(LoaderInterface):
```
drobison00: One thing that would be nice here would be to be able to seek/split based on bytes, time, or frame offset; it makes chopping up the file more intutitive.

This would make it easier for end users who want to do something like splitting into 60 second chunks, or who want to process based on specific offsets in a file or based on other meta information.

drobison00 (changes requested):

api/src/nv_ingest_api/util/dataloader/dataloader.py
jperez999: No it is not, even, ffmpeg-python package calls out to a subprocess. I had to pull this code in to be able to use it on bytes instead of a file. But its basicially the same thing as the _probe from the package but with the ability to use the filebytes. https://github.com/kkroening/ffmpeg-python/blob/master/ffmpeg/_probe.py. This is the fastest library running both the probe and the split logic. Others take an order of magnitude longer to actually run through the testing.

api/src/nv_ingest_api/util/dataloader/dataloader.py
jperez999: The user shouldn't ever call stop. This is something we use to reset the iterator. We want to be able to continue using the same iterator (from where it stopped) if a user decides to break during an iteration and then still access the rest of the data using next(itr). But I can definitely add a del and exit that does the same.

api/src/nv_ingest_api/util/dataloader/dataloader.py
jperez999: In order to solve for this I am going to remove the __getitem__ method. I understand that the behavior is unexpected, however, the reason for it being this way is that we are using a queue to store our chunks. And not all chunks are available all the time. In order to preserve system memory we do load in all the files at the same time. And because the elements we have reside within a queue, we dont have the ability to quickly index into the queue and pick out the corresponding value. Also once the value is returned by the iterator it no longer exists in the iterator. Meaning that if a user decides to do something like itr[1], then itr[0] we will end up in a scenario where the user would not get the desired behavior. I mean we can make it to where the iterator dumps all data to a list (which I am strongly against) and then make it indexible but in doing so we run the risk of adversely occupying larger portions of system memory than expected which can manifest as problems when running the pipeline.

api/src/nv_ingest_api/util/dataloader/dataloader.py
jperez999: If we do this, we would have to redo the work of splitting the files. I will add a switch that allows for clean up, but can be turned off. I can see a user having a ton of files and going through the process of breaking them up, running an iteration of the pipeline and then maybe wanting to tweak something in the pipeline but run with the same files. It would be cool if they could just use the artifacts already created and point to that directory for ingestor files. Thoughts?

api/src/nv_ingest_api/util/dataloader/dataloader.py
jperez999: All of these are just different degrees of time. I think that for RAG we actually want to limit this, and users probably want to get as big of elements as possible. This ensures that they can take advantage of the larger context windows in the model. We have seen that larger contexts usually end up meaning better results. (I.e. PDFs getting split at page level, the less you split the better)

api/src/nv_ingest_api/util/dataloader/dataloader.py
jperez999: Also the daemon=True prevents the leaking of the thread. If the main process shuts down that boolean ensures the thread dies with the parent.

api/src/nv_ingest_api/util/dataloader/dataloader.py
```diff
@@ -0,0 +1,371 @@
...
+                raise ValueError(f"Invalid split type: {split_type}")
+
+        def _get_path_metadata(self):
+            """
+            Get the metadata for a path.
+            path: str, path to get the metadata for if None, get the metadata for all paths
+            """
+            return self.path_metadata
+
+    def load_data(queue: queue.Queue, paths: list[str], thread_stop: threading.Event):
```
drobison00, line 253: I think, that even though thread_stop is annotated as a threading.Event here, it is passed in as a boolean in all the use cases, not as a threading.Event. I believe this results in a copy of the bool being passed to the worker, not an actual event that will be observable across threads.

Its got to be initialized as a threading.Event, then checked as thread_stop.is_set()

PR #847, opened 2025-06-18
[REVIEW] PDF Extraction with Parallel Page Rendering

VibhuJawa:
This PR improves PDF extraction performance and configuration flexibility:

- Enabled parallel PDF page rendering using ProcessPoolExecutor for faster processing and natural page order
- Added _get_max_workers to dynamically calculate optimal worker count based on system resources


api/src/nv_ingest_api/internal/extract/pdf/engines/pdfium.py
```diff
@@ -377,6 +378,94 @@ def _extract_page_elements(
...
+    pdf_source: Union[str, bytes],
+    page_idx: int,
+    size: Tuple[int, int],
+    execution_trace_log: Optional[List] = None,
+) -> Tuple[int, np.ndarray, Tuple[int, int]]:
+    """
+    Opens `pdf_source`, renders `page_idx`, and returns
+    (page_idx, image_array, padding_offsets).
+    """
+    doc = libpdfium.PdfDocument(pdf_source)
```
jdye64: Am I understanding correctly that if a PDF file was say 1,000 pages this would open/close that same file 1,000 times to get each individual page here?

api/src/nv_ingest_api/internal/extract/pdf/engines/pdfium.py
```diff
@@ -377,6 +378,94 @@ def _extract_page_elements(
...
+
+    with ProcessPoolExecutor(max_workers=max_workers) as pool:
+        futs = [pool.submit(_render_page_from_file, pdf_source, idx, size) for idx in range(page_count)]
+        for fut in as_completed(futs):
+            idx, img, pad = fut.result()
+            images[idx] = (img, pad)
+            if execution_trace_log is not None:
+                execution_trace_log[f"trace::exit::pdf_extraction::pdfium_pages_to_numpy_{idx}"] = datetime.now()
+
+    return images
```
jdye64: Maybe not now but I could see this using a lot of memory holding all of the images in memory here. Maybe at some point we introduce a generator concept that writes each image after successful generation to disk or the next stage in the pipeline without having to accumulate them all here.

drobison00: This was actually how we did things in the original iteration probably last July, and yes especially depending on the render DPI, it substantially increases memory footprint. Another, somewhat pathological issue is that we are going to be doing very large memory allocations/de-allocations. Some PDFs are well into the 400 MB range, or more in some cases. Worst case you could be allocating that much memory across replica_count * worker count, and at the same time storing all rendered pages in memory.

api/src/nv_ingest_api/internal/schemas/extract/extract_pdf_schema.py
```diff
@@ -4,14 +4,54 @@
...
+    try:
+        # Check if the number of workers is set directly via an environment variable.
+        max_workers_env = os.environ.get("NV_INGEST_WORKERS_PER_PROGRESS_ENGINE")
+        if max_workers_env is not None:
+            return max(1, int(max_workers_env))
+
+        cpu_count = max(1, os.cpu_count())
+        # NV_INGEST_MAX_UTIL is the number of concurrent processes, defaulting to 1.
+        # Ensure nv_ingest_max_util is at least 1 to avoid division by zero.
+        nv_ingest_max_util = max(1, int(os.environ.get("NV_INGEST_MAX_UTIL", "1")))
```
drobison00, line 41: This will be much too aggressive under load, as it will basically be applied per replica, which could result in something like replica_count * MAX_UTIL processes just for extraction. The problem is that we're trying to set a global maximum parallelism count inside of a single worker, without coordinating across workers.

api/src/nv_ingest_api/internal/extract/pdf/engines/pdfium.py
VibhuJawa: I agree, we should dump out  these images to disk without a accumulation stage here as a follow up.

VibhuJawa: I am thinking more about this, I think once @edknv  changes with https://github.com/NVIDIA/nv-ingest/pull/799 land, we can also just produce and save this in memory/on disk using a jpeg/png like format (so each page just produces that) 

For context, the numpy array  of the pdf image we normally benchmark is   `2510.15625 kb`  and that in jpeg (at q=100) is `117.49  kb`, so can be a 20x reduction

drobison00: >I agree, we should dump out these images to disk without a accumulation stage here as a follow up.

This could be an option, but its not clear to me that it makes sense as a default. We put ourselves into a situation where we could have many workers doing large-ish sequantial writes to the same disk in parallel, which has the potential to cause problems for customers who are using slower storage, or perhaps do not even have significant storage on a process node.

Another thing to consider here is that there does not appear to be a specific reason for this one-page-per-process approach, it could be blocks of pages.

Pdfium has an `import_pages` call, which could potentially be used to separate a larger pdf into smaller batches which could be delegated to individual processes.

```python
            pdf = pdfium.PdfDocument(....)
            new_pdf = pdfium.PdfDocument.new()
            new_pdf.import_pages(pdf, list(range(start_page, end_page)))
```

api/src/nv_ingest_api/internal/extract/pdf/engines/pdfium.py
drobison00: Yeah, I agree with Jeremy here, this is inefficient, but also has resource / pathological behavior concerns. See below, but, at minimum it seems like we would want to extract multiple pages per process and find a way to avoid duplicating each PDF, in full, for each process.

api/src/nv_ingest_api/internal/extract/pdf/engines/pdfium.py
```diff
@@ -483,6 +573,17 @@ def pdfium_extractor(
     pages_for_tables = []  # Accumulate tuples of (page_idx, np_image)
     futures = []  # To track asynchronous table/chart extraction tasks
 
+    logger.debug(f"Using {pdfium_config.workers_per_progress_engine} workers for table/chart/infographic extraction.")
+    # Pre-render all pages in parallel if needed for table/chart/infographic extraction
+    rendered_imgs: Optional[List[Tuple[np.ndarray, Tuple[int, int]]]] = None
+    if extract_tables or extract_charts or extract_infographics:
+        rendered_imgs = render_single_pdf_parallel(
```
drobison00, line 618: I think we still need a more explicit root-cause to justify the changes here.

It is clear based on your testing, that there is slowdown when interleaving the render process with inference calls, but there does not seem to be a reasonable explanation as to why these two operations are resulting in contention with a completely separate service (say page-elements).

If the assertion is that there is CPU contention between the current, single process per pdf, approach to rendering and the the Triton service, then we need to understand that better, as there is no obvious reason why Triton would, consistently, be using the same CPU core as the extractor process. The alternative, is that there is some kind of GIL level contention between the render loop and the threads the NVIngest client uses to make inference requests to Triton. 

If it is a GIL related issue, I would prefer to first try and address that directly instead of resorting to rendering the entire file at once. If it is a CPU contention issue, then we need a clear picture of what is causing that contention.

drobison00 (changes requested):

api/src/nv_ingest_api/internal/extract/pdf/engines/pdfium.py
VibhuJawa: >Another thing to consider here is that there does not appear to be a specific reason for this one-page-per-process approach, it could be blocks of pages.

I like this idea, we can explore a block of pages and see how that performs and that should help elevate some of the concerns. 

I think b/w that and only keeping base64 strings around (instead of the whole images) , we can reduce memory  per process a bunch.

PR #863, opened 2025-06-30
Add podman support

araza008:
## Description
- Add podman compatible compose files
    - podman does not support profiles; hence, created separate collection of services
- Update quickstart markdown with instructions to run the compose file

## Checklist

### DIFF
--- a/docs/docs/extraction/quickstart-guide.md
+++ b/docs/docs/extraction/quickstart-guide.md
@@ -51,6 +51,11 @@ If you prefer, you can run on Kubernetes by using [our Helm chart](https://githu
 6. Start core services. This example uses the table-structure profile.  For more information about other profiles, see [Profile Information](#profile-information).
 
     `docker compose --profile retrieval --profile table-structure up`
+    ```shell
+    podman-compose -f podman-compose.yaml up -d
+    podman-compose -f podman-compose-retrieval.yaml up -d
+    podman-compose -f podman-compose-misc.yaml up table-structure -d
+    ```
 
     !!! tip
 
--- a/docs/docs/extraction/quickstart-guide.md
+++ b/docs/docs/extraction/quickstart-guide.md
@@ -26,7 +26,7 @@ If you prefer, you can run on Kubernetes by using [our Helm chart](https://githu
 
     ```shell
     # This is required to access pre-built containers and NIM microservices
-    $ docker login nvcr.io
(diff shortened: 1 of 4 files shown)


docs/docs/extraction/quickstart-guide.md
```diff
@@ -26,7 +26,7 @@ If you prefer, you can run on Kubernetes by using [our Helm chart](https://githu
 
     ```shell
     # This is required to access pre-built containers and NIM microservices
-    $ docker login nvcr.io
+    $ <docker OR podman> login nvcr.io
```
randerzander: Thanks for contributing documentation for podman compose. Happy to have some docs on how to use it with nv-ingest.

That said, generally, our quickstart guide is complex already. Instead of layering in podman here, why not move podman yaml files and docs into a separate, dedicated subfolder and docs page?

Also, instead of duplicating environment variables and values into these new yaml files, are you sure there's not a way to script parsing the docker-compose.yaml file into a compatible set of podman yamls? Without that, the configs will drift apart rapidly, which will be messy.

randerzander (changes requested):

docs/docs/extraction/quickstart-guide.md
randerzander: Additionally, in a podman README or docs, can you explain the motivation for it? I understand it has to do with WSL?

Can you please address when someone would want to use it instead of docker compose?

araza008: That said, generally, our quickstart guide is complex already. Instead of layering in podman here, why not move podman yaml files and docs into a separate, dedicated subfolder and docs page?
> Done! I was also able to converge the containers to single yaml file as podman-compose>=1.1.0 supports profile

Also, instead of duplicating environment variables and values into these new yaml files, are you sure there's not a way to script parsing the docker-compose.yaml file into a compatible set of podman yamls? Without that, the configs will drift apart rapidly, which will be messy.
> It is possible, but not all features supported on docker are available on podman. The script that I write today may not work if the yaml file adds something that is currently not supported on podman. Can we table this for a follow-up PR?

araza008: Additionally, in a podman README or docs, can you explain the motivation for it? I understand it has to do with WSL?

Can you please address when someone would want to use it instead of docker compose?

> Added a note to rtx/docs/quickstart-guide.md

PR #879, opened 2025-07-11
Remove dependencies from nv-ingest-client that are no longer needed

jdye64:
## Description
Remove dependencies from nv-ingest-client that are no longer needed

## Checklist


client/pyproject.toml
```diff
@@ -23,16 +23,9 @@ dependencies = [
     "click>=8.1.8",
     "fsspec>=2025.2.0",
     "httpx>=0.28.1",
-    "langchain-nvidia-ai-endpoints>=0.3.7",
-    "llama-index-embeddings-nvidia==0.1.5",
```
jperez999, line 27: Don't remove the `llama-index-embeddings-nvidia` we still use this in milvus.py

jperez999 (changes requested):
By removing these packages we may need to update the documentation in some places.

PR #889, opened 2025-07-15
[REVIEW] Update main to use YOLOX Ensemble NIMs

VibhuJawa:
## Description

This PR replaces https://github.com/NVIDIA/nv-ingest/pull/799 but deprecates standard YOLOX models in favor of YOLOX Emsemble models. 

Also removes torch dependency.


edknv (approved):
LGTM

api/src/nv_ingest_api/internal/extract/pdf/engines/pdfium.py
```diff
@@ -551,8 +536,6 @@ def pdfium_extractor(
             if extract_tables or extract_charts or extract_infographics:
                 image, padding_offsets = pdfium_pages_to_numpy(
                     [page],
-                    scale_tuple=(YOLOX_PAGE_IMAGE_PREPROC_WIDTH, YOLOX_PAGE_IMAGE_PREPROC_HEIGHT),
-                    padding_tuple=(YOLOX_PAGE_IMAGE_PREPROC_WIDTH, YOLOX_PAGE_IMAGE_PREPROC_HEIGHT),
```
edknv: We might need to keep this because some pages can be very large which we would want to downscale before sending it to NIM.

docker-compose.yaml
```diff
@@ -10,6 +10,7 @@ services:
 
   page-elements:
     image: ${YOLOX_IMAGE:-nvcr.io/nim/nvidia/nemoretriever-page-elements-v2}:${YOLOX_TAG:-latest}
+    shm_size: 16gb
```
randerzander, line 13: please add to graphic elements and table structure as well

api/src/nv_ingest_api/internal/extract/pdf/engines/pdfium.py
VibhuJawa: Thanks for this fix. This at least fixed some slow down I was seeing in ~b767~ bo20, still running b767  <3

docker-compose.yaml
VibhuJawa: Resolved by https://github.com/NVIDIA/nv-ingest/pull/889/commits/076ef1db69e5f31e92502264867e56e0cef5b4f7

PR #913, opened 2025-07-24
[REVIEW] Add stable ordering based on file sizes

VibhuJawa:
## Description
This PR adds stable ordering  of input files , (still testing/opening early for feedback)

### DIFF
--- a/client/client_tests/primitives/jobs/test_job_spec.py
+++ b/client/client_tests/primitives/jobs/test_job_spec.py
@@ -137,11 +137,14 @@ def test_init_with_job_specs(job_spec_fixture):
     assert job_spec_fixture in batch_job_spec._file_type_to_job_spec["pdf"]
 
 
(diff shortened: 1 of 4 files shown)


VibhuJawa:
Restarted CI because do not think this error is related this assertion at all (We may have some variability in the NIMs) so probably need to make the test more robust.  : 

```
>       assert multimodal_first_chart_xaxis in chart_contents
E       AssertionError: assert 'Hammer - Powerdrill - Bluetooth speaker - Minifridge - Premium desk fan' in 'Gadgets and their cost    Dollars $80.00 - $100.00 - $120.00 - $140.00 - $160.00 Bluetooth speaker  Frequency Ranges ... - Subwoofer  scale) Hertz (log 10 - 100 - 1000 - 10000 - 100000 Frequency Range Start (Hz) - Frequency Range End (Hz)'
```

PR #921, opened 2025-07-29
Updating Ray Wheel Env per Anyscale

jioffe502:
## Description
Updating Ray Wheel Env per Anyscale

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -144,6 +144,12 @@ RUN source activate nv_ingest_runtime && \
 
 RUN chmod +x /workspace/docker/entrypoint.sh
 
+
+COPY ray-3.0.0.dev0-cp311-cp311-manylinux2014_x86_64.whl /workspace/ray-3.0.0.dev0-cp311-cp311-manylinux2014_x86_64.whl
+
+RUN source activate nv_ingest_runtime && \
+    pip install /workspace/ray-3.0.0.dev0-cp311-cp311-manylinux2014_x86_64.whl
+
 # Set entrypoint to tini with a custom entrypoint script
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -144,6 +144,12 @@ RUN source activate nv_ingest_runtime && \
 
 RUN chmod +x /workspace/docker/entrypoint.sh
 
+
+COPY ray-3.0.0.dev0-cp311-cp311-manylinux2014_x86_64.whl /workspace/ray-3.0.0.dev0-cp311-cp311-manylinux2014_x86_64.whl
```
jdye64: Lets change the name here. I imagine we might get several more iterations of these .whl files. So lets make it something like `ray.whl` 

Reason being since we are not including the whl in the repo anyone can build and test a container locally and just update whatever new .whl file they have to be `ray.whl` and get those changes.

Make sense? Doesn't sound as good written down as it did in my head to be honest ...

Dockerfile-Ray
```diff
@@ -0,0 +1,184 @@
...
+
+# Remove graphviz and its dependencies to reduce image size
+RUN source activate nv_ingest_runtime && \
+    mamba remove graphviz python-graphviz --force -y && \
+    mamba uninstall gtk3 pango cairo fonts-conda-ecosystem -y
+
+RUN chmod +x /workspace/docker/entrypoint.sh
+
+
+COPY ray-3.0.0.dev0-cp311-cp311-manylinux2014_x86_64.whl /workspace/ray-3.0.0.dev0-cp311-cp311-manylinux2014_x86_64.whl
```
jdye64:
```suggestion
COPY ray-latest-patch.whl /workspace/ray-latest-patch.whl
```

PR #963, opened 2025-08-12
Copy objects from the source bucket to minio default bucket

edknv:
## Description
`pymilvus.utility.do_bulk_insert `only reads from the default MinIO bucket (`a-bucket`). This PR adds a temporary workaround: copy the objects from the source bucket to the default bucket before bulk insert

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,12 +172,17 @@ ingestor = (
 
 print("Starting ingestion..")
 t0 = time.time()
-results = ingestor.ingest(show_progress=True)
+# Return both successes and failures so you can inspect partial errors without stopping uploads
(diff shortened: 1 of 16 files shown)


jdye64 (changes requested):
Looks good. See diff for ast and copy import. Don't see what those are needed for with the change?

Should we also attempt to create the copy source bucket if it does not exist? That would happen before the uploading of the parquet files to minio. That would save users from having to implement routines to create those buckets themselves which seems error prone. Thoughts?

jdye64 (approved):
Makes sense about isort. I just didn't notice that at first. Thanks for the change on creating the bucket if it does not exist. LGTM

PR #965, opened 2025-08-13
Decompose nv-ingest-ms-runtime into nv-ingest-rest-api & nv-ingest-worker

jdye64:
## Description
This PR splits the monolithic nv-ingest-ms-runtime service into 2 separate, and new, services.

- nv-ingest-rest-api
- nv-ingest-worker

The driver for this change is several things.

- Scalability. The REST API and worker services can be independently scaled to allow for more fine grained control of traffic
- Canary releases where worker components can be individually upgraded in production while leaving REST API services live
- Ease reverse proxy setups in Helm

## Checklist

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -241,24 +241,19 @@ services:
     profiles:
       - audio
 
-  nv-ingest-ms-runtime:
-    image: nvcr.io/nvidia/nemo-microservices/nv-ingest:25.6.2
-    shm_size: 40gb # Should be at minimum 30% of assigned memory per Ray documentation
+  nv-ingest-rest-api:
+    image: nvcr.io/nvidia/nemo-microservices/nv-ingest-rest-api:latest
     build:
       context: ${NV_INGEST_ROOT:-.}
       dockerfile: "./Dockerfile"
-      target: runtime
+      target: rest-api
       args:
         DOWNLOAD_LLAMA_TOKENIZER: ${DOWNLOAD_LLAMA_TOKENIZER:-False}
         HF_ACCESS_TOKEN: ${HF_ACCESS_TOKEN:-hfaccesstoken}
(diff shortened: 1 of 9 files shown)


docker-compose.yaml
```diff
@@ -351,6 +346,91 @@ services:
...
+      target: worker
+      args:
+        DOWNLOAD_LLAMA_TOKENIZER: ${DOWNLOAD_LLAMA_TOKENIZER:-False}
+        HF_ACCESS_TOKEN: ${HF_ACCESS_TOKEN:-hfaccesstoken}
+        MODEL_PREDOWNLOAD_PATH: ${MODEL_PREDOWNLOAD_PATH:-/workspace/models/}
+    volumes:
+      - ${DATASET_ROOT:-./data}:/workspace/data
+    cap_add:
+      - sys_nice
+    environment:
```
jdye64, line 373: I don't like having 2 services that share basically the same environment variables simply because trying to keep things synced never ends well. Just leaving here for now and we can discuss approaches.

Notice nv-ingest-rest-api has nearly the same variables

It also seems cumbersome to try and say "well rest-api only needs these envs but then worker needs these other ones and some of the same" seems like blanket of all env variables and a single place to store them all would be preferred. Also easier for docs

docker-compose.yaml
```diff
@@ -351,6 +346,91 @@ services:
               device_ids: ["0"]
               capabilities: [gpu]
 
+
+  nv-ingest-worker:
```
drobison00, line 358: Can we call this nv-ingest-service, so it aligns with all the unit tests.

docker-compose.yaml
drobison00: Do we need any of the args above for the rest service? I didn't think we were doing any real work there.

PR #996, opened 2025-08-20
Mil stream batch

jperez999:
## Description
This PR optimizes streaming files to milvus by adding batching.

## Checklist


client/src/nv_ingest_client/util/vdb/milvus.py
```diff
@@ -899,9 +899,9 @@ def stream_insert_milvus(records, client: MilvusClient, collection_name: str):
         Milvus Collection to search against
     """
     count = 0
-    for element in records:
-        client.insert(collection_name=collection_name, data=[element])
-        count += 1
+    for idx in range(0, len(records), batch_size):
```
jperez999, line 903: Did not use a threadpool executor because when I did I was getting worse recall scores than just batching and sending sequentially.

drobison00: That seems a little scary; is this indicating accuracy is insert order dependent?

jperez999: I do not have enough information to explain what the issue is. Current line of thinking is that the the increased number of simultaneous writes is causing an issue. I am hoping it is not an index corruption error, but we will track it down.

jperez999: Figured it out, same as with the bulk upload, you need time to let the final complete index be constructed.

client/src/nv_ingest_client/util/vdb/milvus.py
```diff
@@ -899,12 +900,28 @@ def stream_insert_milvus(records, client: MilvusClient, collection_name: str):
...
+        client.insert(collection_name=collection_name, data=records[idx : idx + batch_size])
+        count += len(records[idx : idx + batch_size])
     logger.info(f"streamed {count} records")
 
 
+def wait_for_index(collection_name: str, num_elements: int, client: MilvusClient):
```
jperez999, line 909: added this function to handle waiting for all inserts to complete.

edknv (approved):

PR #1000, opened 2025-08-22
feat: Add store task benchmarking with image analysis utilities. 

jioffe502:
Implements comprehensive benchmarking for nv-ingest store task performance comparison and configurable image extraction to disk for visual analysis.

## Description
Adds comprehensive benchmarking tools to measure the performance impact of nv-ingest's MinIO store task on throughput and latency. Includes configurable image extraction utilities that save charts, tables, and infographics to disk for detailed analysis and debugging.

Results for ablation study in slack.

## Checklist


client/src/nv_ingest_client/util/image_disk_utils.py
```diff
@@ -0,0 +1,260 @@
...
+            else:
+                # Flat directory structure with type in filename
+                image_filename = f"{clean_source_name}_{subtype}_p{page_number}_{doc_idx}.{image_type}"
+                image_path = os.path.join(output_directory, image_filename)
+
+            # Decode base64 content and save as image file
+            try:
+                # Decode base64 image data
+                image_data = base64.b64decode(image_content)
+                image = Image.open(io.BytesIO(image_data))
```
VibhuJawa: We  could use `ensure_base64_format`   `from nv_ingest_api.util.image_processing.transforms` conversation, mostly faster than PIL>

jioffe502: Nice catch @VibhuJawa ! I just changed the functionality to leverage this ensure_base64_format and it is faster. 

Also, set default to jpeg and found ~8x speedup in my latest commit per your question on slack.

client/src/nv_ingest_client/util/image_disk_utils.py
```diff
@@ -24,6 +24,13 @@
 from typing import Any, Dict, List
 from PIL import Image
 
+try:
+    from nv_ingest_api.util.image_processing.transforms import base64_to_numpy
+
+    OPENCV_AVAILABLE = True
+except ImportError:
```
VibhuJawa: Interesting, Given this a client side fix, there are 2 ways to fix this. 

1. One we add open cv as a dependency on client (which should be fine) 

or 

2. We  fall back to PIL always, 

We should not fork this.

client/src/nv_ingest_client/util/image_disk_utils.py
```diff
@@ -138,33 +149,47 @@ def save_images_to_disk(
...
-                # Save with appropriate format (handle JPEG naming)
-                image_ext = "jpg" if image_type == "jpeg" else image_type
-                image.save(image_path, format=image_ext.upper())
+                if OPENCV_AVAILABLE:
+                    # Use OpenCV-based decoding for better performance
+                    img_array = base64_to_numpy(image_content)
+
+                    # Convert numpy array back to PIL Image for saving
+                    # (base64_to_numpy returns RGB format)
+                    image = Image.fromarray(img_array)
```
VibhuJawa: If all we want to do is save the image to jpeg we should not create ` image = Image.fromarray(img_array)` here because that will cause a image_content->jpeg->PIL->jpeg round trip. 

I wonder if we can directly write it :

```python3
def base64_to_disk(b64_string: str, output_path: str):
    # Strip off metadata if present (e.g., "data:image/jpeg;base64,")
    if "," in b64_string:
        b64_string = b64_string.split(",")[1]

    # Decode and write directly as JPEG
    with open(output_path, "wb") as f:
        f.write(base64.b64decode(b64_string))

b64_string = ensure_base64_format(image_content, format)
base64_to_disk(b64_string)
```

Like, PIL decoding/encoding round trip is expensive and we should avoid it.

VibhuJawa (changes requested):
We should either stick to PIL or add opencv as a dependency to client. My suggestion is adding that dependency because it is much faster in my earlier experiments

client/src/nv_ingest_client/util/image_disk_utils.py
jioffe502: I am torn because in nv_ingest_api.util.image_processing.transforms import base64_to_numpy OpenCV is used and in: nv_ingest_client/cli/util/processing.py PIL is used but there is no dependencies for either in the client if I am looking in the right place (client/pyproject.toml)

client/src/nv_ingest_client/util/image_disk_utils.py
jioffe502: Gave this a shot and handled it by reading the format and performing a direct write if it matches, preventing any conversion loop.

Also, tested full opencv implementation vs full PIL implementation on Bo20. There is no difference in image processing perf, but openCV makes the full pipeline faster by ~10% 

Still up for debate based on versioning of course but I have two versions ready

randerzander: I think the store_task is already likely to be slow due to storage latency (cloud, or minio)

Given the intent here is to make it easy for users to easily interact w/ extracted images, why not just make the format it uses to save to disk configurable?

Use the fast base64 route by default, and jpeg if user is fine w/ paying the cost because that's easier/preferred for them

jioffe502:
@VibhuJawa @drobison00 reduced duplicate code by moving necessary utilities into transforms.py on the api side and made a few new additions to transforms that include unit tests. Additionally, I simplified some of the logic on the client side when it came to image format processing. All using opencv now from the api side so no new dependencies needed.

Also, this PR establishes the foundation for future enhancements to the .store() task. The new base64_to_disk() and save_image_to_disk() functions in transforms.py provide reusable, well-tested components that can be leveraged when we add fsspec support to enable cloud storage (S3, GCS, Azure) in the ingestion pipeline. I also have benchmark baselines of how our current store task is impacting performance. 

Next steps is a future PR that will integrate fsspec into the .store() task for multi-cloud storage or disk storage during ingestion

scripts/tests/cases/store_benchmark.py
```diff
@@ -0,0 +1,716 @@
...
+        Args:
+            collection_name (str): Name for the Milvus collection
+            store_enabled (bool): Whether to enable MinIO store task
+            store_params (dict): Configuration parameters for store task
+
+        Returns:
+            Ingestor: Configured ingestor instance ready for ingestion
+        """
+        # Set up temporary file directory in /raid instead of /tmp
+        temp_dir = f"/raid/jioffe/tmp/store_task/{self.run_id}_{collection_name}"
```
edknv: This should be generalized and also configurable.

scripts/tests/cases/store_benchmark.py
```diff
@@ -0,0 +1,716 @@
...
+            except Exception as e:
+                print(f"MinIO validation failed: {e}")
+                metrics["minio_validation_error"] = str(e)
+
+        # Clean up temporary files after analysis
+        try:
+            import shutil
+
+            temp_dir = f"/raid/jioffe/tmp/store_task/{self.run_id}_{collection_name}"
+            image_dir = f"/raid/jioffe/tmp/store_task/{self.run_id}_{collection_name}_images"
```
edknv: Related to the previous comment, if `temp_dir` is changed because user is able to set it different, it should change here too.

scripts/tests/cases/store_benchmark.py
```diff
@@ -0,0 +1,716 @@
...
+                metrics["chart_chunks"] = chart_chunks
+
+                print(f"Total elements: {total_elements}")
+                print(f"Chunk breakdown - Text: {text_chunks}, " f"Tables: {table_chunks}, Charts: {chart_chunks}")
+
+                # Analyze images for store task conditions
+                if store_enabled and not skip_image_analysis:
+                    try:
+                        # Save images to disk for detailed analysis
+                        image_output_dir = f"/raid/jioffe/tmp/store_task/{self.run_id}_{collection_name}_images"
```
edknv: Make this configurable?

scripts/tests/cases/store_benchmark.py
```diff
@@ -0,0 +1,716 @@
...
+        Returns:
+            dict: Comprehensive metrics including timing, throughput, and counts
+        """
+        collection_name = f"{self.dataset_name}_{condition_name}_{uuid.uuid4().hex[:8]}"
+
+        print(f"\n{'='*60}")
+        print(f"Running {condition_name} benchmark for {self.dataset_name}")
+        print(f"Store enabled: {store_enabled}")
+        print(f"Collection: {collection_name}")
+        temp_path = f"/raid/jioffe/tmp/store_task/{self.run_id}_{collection_name[:20]}..."
```
edknv: Does this need to hard coded here?

edknv (approved):

client/src/nv_ingest_client/util/image_disk_utils.py
```diff
@@ -0,0 +1,269 @@
...
+
+            if not should_save:
+                continue
+
+            # Process output format and determine file extension
+            target_format = output_format.lower()
+
+            # Determine file extension for naming - API handles format conversion
+            if target_format == "auto":
+                file_ext = "jpg"  # Default extension when preserving original format
```
VibhuJawa: If `image_content` is `png` and `target_format` is `auto` wont we write it to `.jpg` instead of `.png`, can you add a test for this please.

api/src/nv_ingest_api/util/image_processing/transforms.py
```diff
@@ -669,3 +669,107 @@ def scale_numpy_image(
...
+        True if successful, False otherwise.
+
+    Examples
+    --------
+    >>> # Preserve original format (fastest)
+    >>> success = save_image_to_disk(image_b64, "/path/to/output.jpg", "auto")
+    >>>
+    >>> # Convert to JPEG with specific quality
+    >>> success = save_image_to_disk(image_b64, "/path/to/output.jpg", "JPEG", quality=85)
+    """
```
VibhuJawa, line 764: Minor nit here on normalizing format. 

```suggestion
    """
    target_format =  target_format.lower()
```

api/api_tests/util/image_processing/test_transforms.py
```diff
@@ -388,3 +390,290 @@ def test_ensure_base64_is_png_unsupported_format():
...
+        img = Image.new("RGB", (10, 10), color="blue")
+        buffer = io.BytesIO()
+        img.save(buffer, format="JPEG")
+        return base64.b64encode(buffer.getvalue()).decode("utf-8")
+
+    def test_assertion_framework_validation(self):
+        """Validate that test assertions work correctly."""
+        # Verify pytest assertion handling works as expected
+        with pytest.raises(AssertionError):
+            assert False, "Expected assertion failure for test framework validation"
```
VibhuJawa: Remove this , do not think we use it anywhere

client/src/nv_ingest_client/util/image_disk_utils.py
```diff
@@ -0,0 +1,269 @@
...
+    save_raw_images : bool, optional
+        Whether to save raw/natural images. Default is False.
+    organize_by_type : bool, optional
+        Whether to organize images into subdirectories by type. Default is True.
+    output_format : str, optional
+        Output image format for saved files. Default is "auto".
+        - "auto": Preserve original format (fastest, no conversion)
+        - "jpeg": Convert to JPEG (smaller files, good compression)
+        - "png": Convert to PNG (lossless quality)
+        Use "auto" for maximum speed by avoiding format conversion.
```
VibhuJawa, line 94: We might want to add a quick logic here to ensure format is normalized and validated

- jpeg/JPEG/JPG-> jpeg
- png/PNG ->png
- auto/AUTO -> auto

and complain if invalid ?

scripts/tests/cases/store_benchmark.py
```diff
@@ -0,0 +1,716 @@
...
+import argparse
+import json
+import os
+import sys
+import time
+import uuid
+from datetime import datetime
+
+# Add utils path - adjusted for scripts/tests/cases location
+sys.path.append(os.path.join(os.path.dirname(os.path.dirname(os.path.dirname(__file__))), "interact"))
```
VibhuJawa: Do not think we can add `sys` path here. 

We prefer relative imports within a package or add a local runner module. If this stays as a standalone script, gate it under `if __name__ == "__main__"` and keep it out of installed distributions.

scripts/tests/cases/store_benchmark.py
```diff
@@ -0,0 +1,716 @@
...
+
+# Add log function
+logs = []
+
+
+def log(key, val):
+    print(f"{date_hour()}: {key}: {val}")
+    caller_fn = get_parent_script_filename()
+
+    log_fn = "test_results/" + caller_fn.split(".")[0] + ".json"
```
VibhuJawa: We should ensure the result file is unique to prevent over rides, maybe do something like this. 


```suggestion
     caller_stem = os.path.splitext(os.path.basename(caller_fn))[0]
    os.makedirs("test_results", exist_ok=True)
    log_fn = f"test_results/{caller_stem}_{RUN_ID}.json"
```

VibhuJawa (changes requested):
Most changes to the core codebase look good, added nits. 

We  should clean up the `scripts/tests/cases` a bit to ensure its generalized and cleaned up.

client/src/nv_ingest_client/util/image_disk_utils.py
jioffe502: did this with new: test_auto_format_preserves_original_png_vs_jpeg

jioffe502:
Thanks for the reviews @VibhuJawa @edknv 

I went ahead and took out scripts_benchmarks and put into my private local. I came to the conclusion that It shouldn't be included as a core test in nv-ingest and something that I can benchmark privately. 

I also made some logic changes with the format preservation, simplified it, and added unit tests. Also tested this with my store_benchmark to make sure it works in real scenarios.

client/src/nv_ingest_client/util/image_disk_utils.py
```diff
@@ -0,0 +1,294 @@
...
+    try:
+        original_format = _detect_base64_image_format(image_content)
+        if original_format == "PNG":
+            return "png"
+        elif original_format in ["JPEG", "JPG"]:
+            return "jpeg"
+        else:
+            return "jpeg"  # Default for unknown formats
+    except Exception:
+        return "jpeg"  # Default if detection fails
```
VibhuJawa: Nit: Lets add logging here for fallback 

```suggestion
    DEFAULT_EXT = "jpg"  # must be either "jpg" or "png"

    try:
        fmt = _detect_base64_image_format(image_content).upper()
    except Exception:
        logger.warning("Image format detection failed; falling back to default '%s'.", DEFAULT_EXT)
        return DEFAULT_EXT

    ext_map = {
        "JPEG": "jpg",
        "JPG":  "jpg",
        "PNG":  "png",
    }

    ext = ext_map.get(fmt, None)
    if ext:
        return ext

    logger.warning("Unsupported image format '%s'; falling back to default '%s'.", fmt, DEFAULT_EXT)
    return DEFAULT_EXT
```

VibhuJawa (approved):
Minor nit around logging, looks good other wise

client/src/nv_ingest_client/util/image_disk_utils.py
jioffe502: Done and tested! @VibhuJawa

PR #1007, opened 2025-08-28
Add support for max batch size for nr-ocr nim

edknv:
## Checklist

### DIFF
--- a/api/src/nv_ingest_api/internal/primitives/nim/nim_client.py
+++ b/api/src/nv_ingest_api/internal/primitives/nim/nim_client.py
@@ -94,6 +94,9 @@ class NimClient:
         if model_name == "yolox_ensemble":
             model_name = "yolox"
 
+        if model_name == "scene_text_ensemble":
+            model_name == "scene_text_pre"
+
         if model_name in self._max_batch_sizes:
             return self._max_batch_sizes[model_name]
 
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -115,7 +115,8 @@ services:
       - NGC_API_KEY=${NIM_NGC_API_KEY:-${NGC_API_KEY:-ngcapikey}}
       - CUDA_VISIBLE_DEVICES=0
       - NIM_TRITON_RATE_LIMIT=3
-      - NIM_TRITON_CUDA_MEMORY_POOL_MB=${OCR_CUDA_MEMORY_POOL_MB:-3072}
+      - NIM_TRITON_CUDA_MEMORY_POOL_MB=${OCR_CUDA_MEMORY_POOL_MB:-8192}
+      - NIM_TRITON_MAX_BATCH_SIZE=${OCR_BATCH_SIZE:-32}
     deploy:
       resources:
         reservations:


randerzander:
Benchmark Results: 
  GPUs: NVIDIA H100 NVL

  nv-ingest-cli check: :white_check_mark:
  helm check: :x:
  library mode check: :x:
  audio check: :white_check_mark:
  pptx/docx/image/txt check: :white_check_mark:

  Latency Test (bo20 PDFs): :white_check_mark:
  20/20 docs, 496 pages in 63.63 extraction seconds: 7.80 pages/sec
  63.63 e2e (includes indexing) seconds: 7.80 pages/sec

  Retrieval Latency Test (N/A) :x::
    Results file not found.

  Throughput Test (bo767 PDFs): :white_check_mark:
  767/767 docs in 33 minutes:
  27.64 e2e pages/sec

  Bo767: 52 queries/sec: 
  {'recall @5': 0.8506559031281534}

  w/ reranker 1 queries/sec:
  {'recall @5': 0.9041372351160444}

  Earnings:
  {'recall @5': 0.6401273885350318}

  w/ reranker:
  {'recall @5': 0.7468152866242038}

  Docker Images:
  llama-3.2-nv-rerankqa-1b-v2:1.8.0-rc2-latest-release-33821100
llama-3.2-nv-embedqa-1b-v2:1.10.0-rc0-latest-dev-33669982
nemoretriever-ocr-v1:1.1.0-rc1-latest-release-33936472
nv-ingest:25.8.0-RC8
dcgm-exporter:4.1.1-4.0.4-ubuntu22.04

api/src/nv_ingest_api/internal/extract/image/infographic_extractor.py
```diff
@@ -144,28 +137,35 @@ def _update_infographic_metadata(
...
+            ocr_res[0],
+            ocr_res[1],
+        )
 
     return results
 
 
-def _create_clients(
+@contextmanager
+def _create_ocr_client(
```
VibhuJawa, line 148: I see we create multiple OCR clients here , one in each instance/replica of `infographic_extractor`, `table_extractor`, `chart_extractor` ?

Will the dynamic batching here on a per client basis or across clients ? 

I think having an ability to share a single client (and controlling) the concurrency, batch sizes we submit to the service is very useful but if its indeed just per client basis, i dont follow how we effective dynamic batching will be here as each extractor might create a  short-lived client.

api/src/nv_ingest_api/internal/primitives/nim/model_interface/ocr.py
```diff
@@ -548,6 +340,386 @@ def _postprocess_ocr_response(
...
+            # Single-image fallback
+            img = base64_to_numpy(data["base64_image"])
+            data["images"] = [img]
+
+        else:
+            raise KeyError("Input data must include 'base64_image' or 'base64_images'.")
+
+        return data
+
+    def coalesce_requests_to_batch(
```
VibhuJawa, line 593: Add a  note explaining that model interface supports dynamic batching via `coalesce_requests_to_batch()`.

api/src/nv_ingest_api/internal/primitives/nim/model_interface/ocr.py
```diff
@@ -548,6 +340,386 @@ def _postprocess_ocr_response(
...
+
+        Raises
+        ------
+        KeyError
+            If either "images" or "image_dims" is not found in `data`.
+        ValueError
+            If an invalid protocol is specified.
+        """
+
+        images = data["images"]
```
VibhuJawa, line 459: Nit: Lets stick with `base64_list` here, just to align it with code base and `NemoRetrieverOCRModelInterface`. 

```suggestion
        base64_list = data["images"]
```

api/src/nv_ingest_api/internal/primitives/nim/nim_client.py
```diff
@@ -390,6 +456,137 @@ def _http_infer(self, formatted_input: dict) -> dict:
...
+
+                start_time = time.monotonic()
+
+                while len(requests_batch) < self._batch_size:
+                    if (time.monotonic() - start_time) >= self._batch_timeout:
+                        break
+
+                    if self._batch_memory_budget_bytes:
+                        try:
+                            next_req_peek = self._request_queue.queue[0]
```
VibhuJawa: Check for race condition here

api/src/nv_ingest_api/internal/primitives/nim/nim_client.py
```diff
@@ -390,6 +456,137 @@ def _http_infer(self, formatted_input: dict) -> dict:
...
+
+                            next_h = max(current_max_h, next_req_peek.dims[0])
+                            next_w = max(current_max_w, next_req_peek.dims[1])
+                            potential_batch_size = len(requests_batch) + 1
+
+                            num_channels = 3
+                            bytes_per_element_fp32 = 4
+                            potential_memory_bytes = (
+                                potential_batch_size * next_h * next_w * num_channels * bytes_per_element_fp32
+                            )
```
VibhuJawa: This logic is probably dependent on model/values we are sending and may change for other models. Lets abstract/send this logic to a model specific code if possible

VibhuJawa (commented):
I like the logic and the ability to control how we send/batch requests but given we are doing it on a single client basis and we may have multiple clients in play, i dont know how it will work. 

Can you confirm how it will work, maybe i am not sure how we are queing stuff .

randerzander:
Benchmark Results: 
  GPUs: NVIDIA H100 NVL

  nv-ingest-cli check: :white_check_mark:
  helm check: :x:
  library mode check: :x:
  audio check: :white_check_mark:
  pptx/docx/image/txt check: :white_check_mark:

  Latency Test (bo20 PDFs): :white_check_mark:
  20/20 docs, 496 pages in 69.22 extraction seconds: 7.17 pages/sec
  69.22 e2e (includes indexing) seconds: 7.17 pages/sec

  Retrieval Latency Test (N/A) :x::
    Results file not found.

  Throughput Test (bo767 PDFs): :white_check_mark:
  767/767 docs in 33 minutes:
  27.64 e2e pages/sec

  Bo767: 52 queries/sec: 
  {'recall @5': 0.8435923309788093}

  w/ reranker 1 queries/sec:
  {'recall @5': 0.9031281533804238}

  Earnings:
  {'recall @5': 0.64171974522293}

  w/ reranker:
  {'recall @5': 0.7531847133757962}

  Docker Images:
  llama-3.2-nv-rerankqa-1b-v2:1.8.0
nemoretriever-ocr-v1:1.1.0
llama-3.2-nv-embedqa-1b-v2:1.10.0
nv-ingest:25.6.2
dcgm-exporter:4.1.1-4.0.4-ubuntu22.04

api/src/nv_ingest_api/internal/extract/image/infographic_extractor.py
edknv: You are absolutely right. I added an abstraction for the single shared client as `ClientManager`  in `nim_client.py`.

VibhuJawa: I don't understand the ray details of nv-ingest well enough so will lean on @drobison00   and you here. 

But even with shared client across threads, wont we be in  different processes and this will still be different for infographic_extractor, table_extractor, chart_extractors ?

VibhuJawa: Like, my understanding is we will have to create a dedicated Ray actor that owns the client instance and exposes methods the other actors call and  that actor becomes NimClientManager. 

Happy to jump on a call to understand this and talk it out.

edknv: Yes, it will still be separate and different clients for table/chart/info, but it's an improvement from before where we created separate table/chart/info clients for each control message and where we had no batching at all for tables/charts/infographics. I'm not sure if trying to combine 3 table/chart/info clients into 1 is worth the extra complexity because I think that would require going outside of the modularized table/chart/info extraction stages.

VibhuJawa: Thanks for the clarification.

My understanding: Each actor  (table/chart/info) keeps a long‑lived NimClient, and requests coming through that actor (from multiple control messages and PDFs) are dynamically batched by the client up to a configured size or short timeout. 

That reduces the number of replicas of client as well as the concurrent OCR inferences hitting the NIM service while improving throughput under mixed workloads (larger and smaller PDFs) . Parallelism still comes from running multiple actor replicas; within a single actor, on_data is processed serially, but the client-side batcher coalesces near-simultaneous requests for better batch size optimizations.  **_Is that correct ?_** 

This also means we can tune behavior per replica: scale the number of replicas for parallelism across documents and adjust the client’s batching parameters to balance latency/throughput and protect the NIM backend (which is still fragile under load today) .

Longer term (depending on how NIM’s concurrency/robustness evolves), we could tighten or relax these knobs on our side to avoid overloading the service—today NIMs can be a bit unstable under heavy load, so I was trying to confirm whether the goal here was to cap concurrency vs unlocking batching benefits. It sounds like the latter, which makes sense.

The implementation looks good to me at a high level; I’ll do a deeper pass shortly.

VibhuJawa (approved):
LGTM , thanks for making these changes and explaining stuff to me

PR #1012, opened 2025-09-04
Add RAG Blueprint to applications section and Update install versions (main)

nkmcalli:
Add RAG Blueprint to applications section and Update install versions (main)


nkmcalli:
> @nkmcalli I don't think we have released the 25.9.0 pypi pkgs yet. @jdye64 can confirm if we did.

@sosahi That's correct.  This is for the main branch.

sosahi (approved):

PR #1014, opened 2025-09-04
Add RAG Blueprint to applications section and Update install versions (25.9.0)

nkmcalli:
Add RAG Blueprint to applications section and Update install versions (25.9.0)


nkmcalli:
> same here, we should not update the pkgs to 25.9.0 if we have not yet published them on pypi.

@sosahi , That's correct.  This is for the RC branch.  We need the update in the RC branch so that the docs can go live with the correct version when the RC goes live.

PR #1026, opened 2025-09-10
Fix milvus bucket

jperez999:
## Description
This PR removes previously added code to copy data between multiple buckets and adds use of the MINIO_BUCKET env var in the Milvus service  from docker-compose.yaml

## Checklist


jperez999:
@nkmcalli we need to add a documentation snippet somewhere informing users that if they want to change the bucket name they need to use the MINIO_BUCKET environment variable.

edknv (approved):

PR #1051, opened 2025-09-22
Install LGPL version of ffmpeg

edknv:
## Description
This PR installs the LGPL version of ffmpeg in the container by building it from source.

## Checklist

### DIFF
--- a/api/api_tests/internal/extract/image/test_chart_extractor.py
+++ b/api/api_tests/internal/extract/image/test_chart_extractor.py
@@ -49,9 +49,10 @@ class TestExtractChartDataFromImageInternal(unittest.TestCase):
         endpoint_config.workers_per_progress_engine = 8
         self.extraction_config.endpoint_config = endpoint_config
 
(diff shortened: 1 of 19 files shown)


docker/scripts/install_ffmpeg.sh
```diff
@@ -0,0 +1,82 @@
...
+done
+
+# Install video dependency
+export DEBIAN_FRONTEND=noninteractive
+apt-get update
+apt-get install -y \
+    libcrypt-dev \
+    autoconf \
+    automake \
+    build-essential \
```
randerzander, line 38: I know we have to build ffmpeg.. but is there a way to install these deps, do the build, then remove them after the build is complete?

Concerns:
1. Do all these build deps substantially increase the container size?
2. Do any of them come with licenses that are going to be scary in scan results?

edknv: Good point. Let me look into removing those after the build.

PR #1054, opened 2025-09-23
PDF splitting in rest client - api v2

jioffe502:
## Description

The goal is to create a V2 REST API that preprocesses PDFs by splitting them into single-page subjobs before submission to Redis and Ray, improving processing throughput.

Current progress with pdf splitting without changes to any Pydantic schemas.

The main challenge right now is aligning the V2 job submission with the existing `IngestJobSchema` validation.

## Checklist

### DIFF
--- a/tests/service_tests/api/v2/test_aggregation_helpers.py
+++ b/tests/service_tests/api/v2/test_aggregation_helpers.py
@@ -478,38 +478,8 @@ class TestBuildAggregatedResponse:
         assert aggregated["metadata"]["subjobs_failed"] == 1
         assert aggregated["metadata"]["failed_subjobs"] == failed
         assert aggregated["data"] == ["chunk-1-data", "chunk-3-data"]
-        assert aggregated["metadata"]["trace_segments"] == [
-            {
-                "job_id": sample_subjob_descriptors[0]["job_id"],
-                "chunk_index": sample_subjob_descriptors[0]["chunk_index"],
-                "start_page": sample_subjob_descriptors[0].get("start_page"),
-                "end_page": sample_subjob_descriptors[0].get("end_page"),
-                "trace": {"trace::entry::foo": 123},
-            },
-            {
-                "job_id": sample_subjob_descriptors[2]["job_id"],
-                "chunk_index": sample_subjob_descriptors[2]["chunk_index"],
-                "start_page": sample_subjob_descriptors[2].get("start_page"),
-                "end_page": sample_subjob_descriptors[2].get("end_page"),
-                "trace": {"trace::entry::bar": 456},
-            },
-        ]
-        assert aggregated["metadata"]["annotation_segments"] == [
-            {
-                "job_id": sample_subjob_descriptors[0]["job_id"],
-                "chunk_index": sample_subjob_descriptors[0]["chunk_index"],
-                "start_page": sample_subjob_descriptors[0].get("start_page"),
-                "end_page": sample_subjob_descriptors[0].get("end_page"),
-                "annotations": {"annotation::foo": {"task_result": "SUCCESS"}},
-            },
-            {
-                "job_id": sample_subjob_descriptors[2]["job_id"],
-                "chunk_index": sample_subjob_descriptors[2]["chunk_index"],
-                "start_page": sample_subjob_descriptors[2].get("start_page"),
-                "end_page": sample_subjob_descriptors[2].get("end_page"),
-                "annotations": {"annotation::bar": {"task_result": "SUCCESS"}},
-            },
-        ]
+        assert aggregated["metadata"]["trace_segments"] == []
+        assert aggregated["metadata"]["annotation_segments"] == []
 
     def test_handles_missing_telemetry_gracefully(self, sample_subjob_descriptors, sample_parent_metadata):
         """Ensure segments are omitted when telemetry is absent or malformed."""
@@ -533,41 +503,6 @@ class TestBuildAggregatedResponse:
 
         assert aggregated["metadata"]["trace_segments"] == []
         assert aggregated["metadata"]["annotation_segments"] == []
-
-    def test_infers_page_span_from_content_metadata(self, sample_parent_metadata):
-        """Verify that page ranges fallback to content metadata when descriptor lacks info."""
-
-        descriptors = [
-            {"job_id": "job-1", "chunk_index": 1},
-        ]
-
-        subjob_results = [
-            {
-                "data": ["chunk-1-data"],
-                "metadata": {
-                    "content_metadata": {"page_range": "pages_5-8"},
-                    "retrieved_document": {
-                        "data": {
-                            "trace": {"trace::entry::foo": 123},
-                            "annotations": {"annotation::foo": {"task_result": "SUCCESS"}},
-                        }
-                    },
-                },
-            }
-        ]
-
-        aggregated = _build_aggregated_response(
-            "parent-id",
-            subjob_results,
-            [],
-            descriptors,
-            sample_parent_metadata,
-        )
-
-        segment = aggregated["metadata"]["trace_segments"][0]
-
-        assert segment["start_page"] == 5
-        assert segment["end_page"] == 8
 
 
 class TestUpdateJobStateAfterFetch:


src/nv_ingest/api/v2/ingest.py
```diff
@@ -0,0 +1,444 @@
...
+                        parent_uuid = uuid.uuid4()
+
+                    subjob_order = []
+
+                    for page_num, page_pdf in enumerate(page_pdfs, 1):
+                        # Create subjob spec by deep copying and modifying
+                        subjob_spec = copy.deepcopy(job_spec_dict)
+                        
+                        # Update the payload in the proper location
+                        subjob_spec["job_payload"]["content"] = [base64.b64encode(page_pdf).decode("utf-8")]
```
jperez999: Here you are setting the job content as actual data, not a file path. I think doing this will cause increased memeory strain on redis. I think this is what @drobison00 was trying to prevent.

src/nv_ingest/api/v2/ingest.py
```diff
@@ -0,0 +1,444 @@
...
+        200: {"description": "Jobs were successfully submitted"},
+        500: {"description": "Error encountered while submitting jobs."},
+        503: {"description": "Service unavailable."},
+    },
+    tags=["Ingestion"],
+    summary="submit jobs to the core nv ingestion service for processing with PDF splitting",
+    operation_id="submit_job_v2",
+)
+async def submit_job_v2(request: Request, response: Response, job_spec: MessageWrapper, ingest_service: INGEST_SERVICE_T):
+    with tracer.start_as_current_span("http-submit-job-v2") as span:
```
jperez999: can we turn this tracer logic into a decorator. Hopefully it could be applied as a decorator wherever necessary, given the name is actually the name of the function.

src/nv_ingest/api/v2/ingest.py
```diff
@@ -0,0 +1,444 @@
...
+    STATE_RETRIEVED_NON_DESTRUCTIVE,
+    STATE_RETRIEVED_CACHED,
+    STATE_FAILED,
+    STATE_PROCESSING,
+    STATE_SUBMITTED,
+    INTERMEDIATE_STATES,
+)
+
+
+def split_pdf_to_pages(pdf_content: bytes) -> List[bytes]:
```
jperez999: can we set this up to use logic, that allows the user to decide the split page interval (i.e. could be 1 but maybe I to break the pdf up into subsets of 10 pages).

src/nv_ingest/api/v2/ingest.py
```diff
@@ -0,0 +1,444 @@
...
+            payloads = job_payload.get("content", [])
+            
+            # Check if this is a PDF that needs splitting
+            if document_types and payloads and document_types[0].lower() == "pdf":
+                # Decode the payload to check page count
+                pdf_content = base64.b64decode(payloads[0])
+                page_count = get_pdf_page_count(pdf_content)
+                
+                # Split if more than 3 pages
+                if page_count > 3:
```
jperez999: Why do you need to have this check here? Shouldn't the split logic detect if it needs to split or not?

src/nv_ingest/api/v2/ingest.py
```diff
@@ -0,0 +1,444 @@
...
+        404: {"description": "Job ID not found or associated state has expired."},
+        410: {"description": "Job result existed but is now gone (expired or retrieved destructively/cached)."},
+        500: {"description": "Internal server error during fetch processing."},
+        503: {"description": "Job processing failed, or backend service temporarily unavailable preventing fetch."},
+    },
+    tags=["Ingestion"],
+    summary="Fetch the result of a previously submitted job by its job_id (V2 with aggregation)",
+    operation_id="fetch_job_v2",
+)
+async def fetch_job_v2(job_id: str, ingest_service: INGEST_SERVICE_T):
```
jperez999, line 656: This is a very long function, and that base if/else statement makes me feel like we can break this up into smaller functions that we can then put together in the fetch_job_v2 function to get desired behavior. Is there a reason this logic is not split into smaller subsets, maybe I am missing something?

jperez999 (commented):
Great start, moving in the right direction.

src/nv_ingest/api/v2/ingest.py
jioffe502: I just pushed similar changes just now allowing n-page splitting

src/nv_ingest/api/v2/ingest.py
jioffe502: this initially was very defensive programming where I totally keep the fetch Logic for v1 if there is no parent - child mapping in redis. 

im planning on simplifying it down a lot just for now at this point in the draft keeping it verbose

src/nv_ingest/api/v2/ingest.py
jioffe502: changed logic with newest commit to splitting if the page count > n-page splitting criteria

src/nv_ingest/api/v2/ingest.py
jioffe502: great idea

src/nv_ingest/api/v2/ingest.py
jioffe502: will look into this

src/nv_ingest/api/v2/ingest.py
```diff
@@ -0,0 +1,507 @@
...
+
+    pdf = pdfium.PdfDocument(pdf_content)
+    total_pages = len(pdf)
+
+    try:
+        for chunk_index, start_zero in enumerate(range(0, total_pages, pages_per_chunk)):
+            end_zero = min(start_zero + pages_per_chunk, total_pages)
+            page_indices = list(range(start_zero, end_zero))
+
+            new_pdf = pdfium.PdfDocument.new()
```
drobison00, line 91: I think when we do this, we're going to lose meta information on all the child PDF's from the parent. Not 100%, but we should probably check.

src/nv_ingest/api/v2/ingest.py
```diff
@@ -0,0 +1,507 @@
...
+)
+async def submit_job_v2(
+    request: Request, response: Response, job_spec: MessageWrapper, ingest_service: INGEST_SERVICE_T
+):
+    with tracer.start_as_current_span("http-submit-job-v2") as span:
+        try:
+            # Add custom attributes to the span
+            span.set_attribute("http.method", request.method)
+            span.set_attribute("http.url", str(request.url))
+            span.add_event("Submitting file for processing (V2)")
```
drobison00: Re: Julio's comment on monolithic functions. 

We could probably make this a modular helper so its more declarative.

src/nv_ingest/api/v2/ingest.py
```diff
@@ -0,0 +1,507 @@
...
+                        parent_uuid = uuid.UUID(parent_job_id)
+                    except ValueError:
+                        logger.warning(
+                            "Parent job id %s is not a valid UUID; generating fallback namespace for subjobs",
+                            parent_job_id,
+                        )
+                        parent_uuid = uuid.uuid4()
+
+                    for chunk in chunks:
+                        chunk_number = chunk["chunk_index"] + 1
```
drobison00: Re: Juilo's monolithic function comment

This could be its own function to process each chunk, you could also get away with issuing all chunks asynchronously and then wait on all of them at the end of the loop.

src/nv_ingest/api/v2/ingest.py
```diff
@@ -0,0 +1,507 @@
...
+            for idx, subjob_id in enumerate(subjob_ids, 1):
+                ordered_descriptors.append({"job_id": subjob_id, "chunk_index": idx})
+
+            # Check all subjob states
+            all_complete = True
+            any_failed = False
+            subjob_results = []
+            failed_subjobs: List[Dict[str, object]] = []
+
+            for page_index, descriptor in enumerate(ordered_descriptors, 1):
```
drobison00: This can be made parallel, I would issue all the subjob status requests together then process them as the results become available. `asyncio.gather`

src/nv_ingest/api/v2/ingest.py
jioffe502: Current behavior is that the child PDFs get new timestamps/producer but the source_id is still traceable i.e. document.pdf#pages_1-32

the only meta information lost through creating a new document is:

```    
date_created
last_modified
producer
```

jioffe502: we can maybe pass these as additional tracing options?

jioffe502:
Added V2 ingest API that pre-splits PDFs at a configurable page threshold, fans out chunked subjobs, and aggregates results before returning.
Updated REST client, service wiring, and docker env to select the new v2 endpoints and expose PDF_SPLIT_PAGE_COUNT.
Documented the workflow and backfilled tests: new dc20_v2_e2e case plus unit coverage for PDF splitting and aggregation helpers.

Additionally, added unit coverage for the new V2 PDF splitting and aggregation helpers, including chunk metadata, subjob preparation, and the submit_job_v2 split flow. Exercised the full parent fetch path—state checks, batched result retrieval, aggregation, and state updates.

DATA: averaged over 8 runs on an H100 NVL

PDF_SPLIT_PAGE_COUNT = 2 -> 50.22s e2e, 9.98 pps (highest stddev)
PDF_SPLIT_PAGE_COUNT = 4 -> 36.77s e2e, 13.5 pps
PDF_SPLIT_PAGE_COUNT = 8 -> 36.05s e2e, 13.77 pps 
PDF_SPLIT_PAGE_COUNT = 16 -> 36.38s e2e, 13.67 pps
PDF_SPLIT_PAGE_COUNT = 32 -> 41.38 e2e, 12.02 pps

Testing Guide:

### Two .env Files

**1. `.env` (root) - Infrastructure**
```bash
PDF_SPLIT_PAGE_COUNT=32    # Pages per chunk

NV_INGEST_MAX_UTIL=32
MAX_INGEST_PROCESS_WORKERS=32
INGEST_DISABLE_DYNAMIC_SCALING=true

# And the most recent NIM RCs 
```

**2. `scripts/tests/.env` - Test settings** 

Refer to nv-ingest/scripts/tests/README.md for more information on how to set this up

```bash
DATASET_DIR=/raid/jioffe/bo20
TEST_NAME=dc20_v2 # this will be what populates your artifacts directory
```

---

### Test V2

```bash
# Run the v2 test (AUTOMATICALLY uses v2 API):
python scripts/tests/run.py --case=dc20_v2_e2e --doc-analysis

# Expected output for bo20:
# ✓ V2 results match V1 expected values!
# text=496, tables=164, charts=184
```

---

### Test Different Split Sizes

```bash
# Edit ./.env → change PDF_SPLIT_PAGE_COUNT=8

# Restart docker services
docker compose down
docker compose --profile retrieval --profile table-structure up -d

# Run tests again
python scripts/tests/run.py --case=dc20_v2_e2e --doc-analysis
```

### Key Points

- **Root `.env`**: Controls PDF splitting (infrastructure setting)
- **`scripts/tests/.env`**: Controls test parameters (dataset, hostname, etc.)
- **V2 auto-enabled**: `dc20_v2_e2e.py` sets `NV_INGEST_API_VERSION=v2` automatically
- **Results**: Should match V1 exactly

src/nv_ingest/api/v2/ingest.py
drobison00: They could probably be taken from the parent PDF and set on the child.

That said, if this is all we're losing, it may not be necessary.

src/nv_ingest/api/v2/ingest.py
```diff
@@ -0,0 +1,731 @@
...
+            # Check if this is a PDF that needs splitting
+            if document_types and payloads and document_types[0].lower() == "pdf":
+                # Decode the payload to check page count
+                pdf_content = base64.b64decode(payloads[0])
+                page_count = get_pdf_page_count(pdf_content)
+                pages_per_chunk = get_pdf_split_page_count()
+
+                # Split if the document has more pages than our chunk size
+                if page_count > pages_per_chunk:
+                    logger.info(
```
drobison00:
```suggestion
                    logger.debug(
```

scripts/tests/cases/dc20_v2_e2e.py
```diff
@@ -0,0 +1,234 @@
...
+    return val
+
+
+def _default_collection_name() -> str:
+    # Make collection name configurable via TEST_NAME env var, default to dc20_v2
+    test_name = _get_env("TEST_NAME", "dc20_v2")
+    return f"{test_name}_{_now_timestr()}"
+
+
+def main() -> int:
```
jperez999, line 60: This test case does not have any asserts. Is there a situation where this could fail silently but report as a success?

jioffe502: @jperez999 dc20_v2_e2e is our e2e integration test

jioffe502: the unit tests I wrote are in tests/service_tests/api/v2

jperez999: right it is an integration test but are we checking for the results to ensure we are actually getting back what is exected? Do the other integration tests follow this same strategy? Seems here like you are returning 0 regardless of whether the results are the same or not. Shouldnt this fail if the results are different? IMO if you changed the if statement at the end to an assert you would be better off because it would fail if that assert failed.

jioffe502: It makes total sense to assert instead of print you're right. This test was just mimicking RCI and is meant to be run standalone to benchmark performance. The print at the end was a sanity check that we are indexing the same # of chunks as v1 to make sure nothing is being lost between v1 and v2 but I understand your point about assertions

jperez999 (approved):

PR #1065, opened 2025-09-29
Update e2e Examples in Scripts Directory

nv-rliu:
## Description
- Removes some redundancy to improve usability of the `scripts/cases` 
- More updates to `pre-commit` hooks

## Checklist

### DIFF
--- a/client/src/nv_ingest_client/client/interface.py
+++ b/client/src/nv_ingest_client/client/interface.py
@@ -764,6 +764,7 @@ class Ingestor:
         Ingestor
             Returns self for chaining.
         """
+        
         extract_tables = kwargs.pop("extract_tables", True)
         extract_charts = kwargs.pop("extract_charts", True)
         extract_page_as_image = kwargs.pop("extract_page_as_image", False)
@@ -837,6 +838,7 @@ class Ingestor:
             # Create ExtractTask with mapped document type for API schema compatibility
             extract_task_params = {"document_type": api_document_type, "extract_method": method, **params}
             extract_task = ExtractTask(**extract_task_params)
+            print(f" ==> EXTRACT configs: {str(extract_task)}")
             self._job_specs.add_task(extract_task, document_type=document_type)
 
         return self
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,23 +1,72 @@
+# https://pre-commit.com/
+# Before first use: `pre-commit install`
+# To run: `pre-commit run --all-files`
+# To update: `pre-commit autoupdate`
+fail_fast: false
+default_language_version:
+  python: python3
+
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v5.0.0
+    rev: v6.0.0
     hooks:
-    -   id: trailing-whitespace
+      - id: trailing-whitespace
         exclude: '(^(docs|data)/|\.md$)'
-    -   id: end-of-file-fixer
-    -   id: check-added-large-files
+      - id: end-of-file-fixer
+      - id: check-added-large-files
         args: ['--maxkb=1500']
-    -   id: check-ast
-    -   id: debug-statements
+      - id: check-ast
+      - id: debug-statements
+      - id: mixed-line-ending
+      - id: check-merge-conflict
+      - id: check-yaml
 
   - repo: https://github.com/psf/black
-    rev: 24.10.0
+    rev: 25.9.0
     hooks:
       - id: black
         args: ["--line-length=120"]
 
+  - repo: https://github.com/pycqa/isort
+    rev: 6.0.1
+    hooks:
+      - id: isort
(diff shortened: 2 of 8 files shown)


client/src/nv_ingest_client/client/interface.py
```diff
@@ -94,11 +89,10 @@ def wrapper(self, *args, **kwargs):
 
 
 class LazyLoadedList(collections.abc.Sequence):
```
drobison00, line 96: This looks like a regression, unless we're removing @edknv 's compression work.

scripts/tests/cases/e2e_with_llm_summary.py
```diff
@@ -0,0 +1,209 @@
...
+        print("ERROR: DATASET_DIR environment variable is required")
+        print("Example: DATASET_DIR=/datasets/bo20 python e2e.py")
+        return 2
+
+    if not os.path.isdir(data_dir):
+        print(f"ERROR: Dataset directory does not exist: {data_dir}")
+        print("Please check the DATASET_DIR path and ensure it's accessible")
+        return 2
+
+    spill_dir = os.getenv("SPILL_DIR", "/tmp/spill")
```
drobison00, line 44: For tests, we probably want to use mkdtemp, or explicitly clean up files we've created.

nv-rliu: Ah good call. Probably good idea to use tmpdirs if the intention isn't for those files to persist. I can make that change

client/src/nv_ingest_client/client/interface.py
nv-rliu: Let me know if you see any logic that was changed. I think this was just the linter complaining about the usage of `Optional` in favor of `dtype | None` 

Can revert this

edknv: @nv-rliu I think you need to sync your branch with main. These changes revert some recent changes in https://github.com/NVIDIA/nv-ingest/commit/fd36cd17c468fda293b022f0b91ad0fab4b5fb60

jioffe502 (commented):
Most of these are changes to the RCI system I started moving into nv-ingest and linter. I tested and everything works as intended. Thanks for making some well needed updates to the system.

Really like how you now write to console and to the stdout.txt file when saving that's smart and good feedback during testing.

jioffe502 (approved):

client/src/nv_ingest_client/client/interface.py
nv-rliu: @edknv just merged in changes from `main`

scripts/tests/cases/e2e_with_llm_summary.py
nv-rliu: @drobison00 I added a blurb to remove the created spill_dirs. I don't think these are used after the e2e is complete so it should be safe to delete anything created at runtime.

client/src/nv_ingest_client/client/interface.py
```diff
@@ -107,7 +101,7 @@ def __init__(self, filepath: str, expected_len: Optional[int] = None, compressio
 
     def __iter__(self) -> Iterator[Any]:
         try:
-            with self._open(self.filepath, "rt", encoding="utf-8") as f:
+            with open(self.filepath, encoding="utf-8") as f:
```
edknv: This reverts changes in https://github.com/NVIDIA/nv-ingest/commit/fd36cd17c468fda293b022f0b91ad0fab4b5fb60.

client/src/nv_ingest_client/client/interface.py
```diff
@@ -94,11 +89,10 @@ def wrapper(self, *args, **kwargs):
...
 
 class LazyLoadedList(collections.abc.Sequence):
-    def __init__(self, filepath: str, expected_len: Optional[int] = None, compression: Optional[str] = None):
+    def __init__(self, filepath: str, expected_len: int | None = None):
         self.filepath = filepath
-        self._len: Optional[int] = expected_len  # Store pre-calculated length
-        self._offsets: Optional[List[int]] = None
-        self.compression = compression
+        self._len: int | None = expected_len  # Store pre-calculated length
+        self._offsets: list[int] | None = None
```
edknv: This reverts changes in https://github.com/NVIDIA/nv-ingest/commit/fd36cd17c468fda293b022f0b91ad0fab4b5fb60.

client/src/nv_ingest_client/client/interface.py
edknv: I think there are a couple of changes that got missed. Added comments for what I could find, but can you please compare with https://github.com/NVIDIA/nv-ingest/commit/fd36cd17c468fda293b022f0b91ad0fab4b5fb60 and double check?

nv-rliu:
@edknv reverted `interface.py` to what's checked-in to main

edknv (approved):

drobison00 (approved):

PR #1067, opened 2025-10-01
Update UDF: LLM Summarization

nv-rliu:
## Description
This PR alters the behavior of `llm_summarizer_udf.py`
1. It will take a payload and parse the first and last chunks from the stored DataFrame.
2. It will concatenate these chunks and make an API call to an LLM to generate a summary.
3. The returned summary is stored back on the original payload as metadata > `custom_content` (see discussion thread for details)

## Changes from Previous Version
- Summaries are now per-doc and not per-page. Doing per-doc summaries seem to make more sense for the downstream use-cases.
- Retrieval of content from df is parallelized (to my knowledge)
- Additional logging for the number of tokens being parsed and duration for LLM calls.

## TBD
- Haven't fully implemented the selection of first and last chunks. Currently using default of 1. 

## Checklist

### DIFF
--- a/api/api_tests/internal/primitives/nim/test_nim_client.py
+++ b/api/api_tests/internal/primitives/nim/test_nim_client.py
@@ -7,6 +7,8 @@ from unittest.mock import MagicMock, patch, call
 import numpy as np
 import requests
 
+from tritonclient.grpc import InferenceServerException
+
 from nv_ingest_api.internal.primitives.nim.nim_client import NimClient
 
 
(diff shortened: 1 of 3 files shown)


api/src/udfs/llm_summarizer_udf.py
```diff
@@ -2,22 +2,38 @@
...
-- LLM_SUMMARIZATION_BASE_URL: API base URL (default: https://integrate.api.nvidia.com/v1)
-- LLM_SUMMARIZATION_TIMEOUT: API timeout in seconds (default: 60)
-- LLM_MIN_CONTENT_LENGTH: Minimum content length to summarize (default: 50)
-- LLM_MAX_CONTENT_LENGTH: Maximum content length to send to API (default: 12000)
+- LLM_BASE_URL: base URL (default: https://integrate.api.nvidia.com/v1)
+- TIMEOUT: API timeout in seconds (default: 60)
+- MIN_CONTENT_LENGTH: Minimum content length to summarize (default: 50)
+- MAX_CONTENT_LENGTH: Maximum content length to send to API (default: 12000)
+TODO: Implement this
+- NUM_FIRST_LAST_PAGES: (Optional) Number of first and last pages to summarize. default=1
```
nv-rliu: ## Current behavior regarding configs
These values must be set before starting the nv-ingest pipeline. Otherwise, to change them while the pipeline is running requires a user to manually modify and hardcode them in the UDF script.

api/src/udfs/llm_summarizer_udf.py
```diff
@@ -2,22 +2,38 @@
...
+
+PROMPT = """
+Here are the contents from the first and last page of a document. Focus on the main purpose, key topics,
+and important details. Just return the summary as a paragraph. Do not add special characters for formatting.
+This summary will be used for document search and understanding.
+
+[CONTENT]
+{content}
+[END CONTENT]
+"""
```
nv-rliu, line 38: Tried to be as explicit and clear as possible. Different models will still dump out random formatting blurbs. I can also test out different prompt suggestions and share summary results. LMK

api/src/udfs/llm_summarizer_udf.py
```diff
@@ -44,167 +53,169 @@ def content_summarizer(control_message: "IngestControlMessage") -> "IngestContro
...
+            _extract_content,
+            axis=1,
+            min_content_length=min_content_length,
+            max_content_length=max_content_length,
+            stats=stats,
+        )
+    )
+    stats["tokens"] = _estimate_tokens(content)
+    summary, duration = _generate_llm_summary(content, model_name, base_url, api_key, timeout)
+    _store_summary(df, summary, model_name)
```
nv-rliu: Main logic of this PR. Lines 96-104 replaces the previous iterative looping through the rows and summarizing per chunk. Bulk of the performance improvement comes from here.

api/src/udfs/llm_summarizer_udf.py
```diff
@@ -44,167 +55,169 @@ def content_summarizer(control_message: "IngestControlMessage") -> "IngestContro
...
+
+    ###
+    ## REMOVE BEFORE MERGING
+    # Log the current directory
```
nv-rliu: self-explanatory. DONT let this PR be merged until this is removed.

api/src/udfs/llm_summarizer_udf.py
```diff
@@ -44,167 +55,169 @@ def content_summarizer(control_message: "IngestControlMessage") -> "IngestContro
...
+            return summary, duration
+        return None, duration
 
     except Exception as e:
         logger.error(f"API call failed: {e}")
-        return None
+        # TODO: GitHub Thread
+        # Reviewers, tell me if this is a bad idea.
+        # I think the convention is to return timestamp for time even if it fails
+        return None, time.time() - start_time
```
nv-rliu, line 181: Would appreciate a second opinion here! :)

api/src/udfs/llm_summarizer_udf.py
```diff
@@ -44,167 +55,169 @@ def content_summarizer(control_message: "IngestControlMessage") -> "IngestContro
...
-    try:
-        # Get current metadata or create new dict - handle None case properly
-        existing_metadata = row.get("metadata")
-        if existing_metadata is not None and isinstance(existing_metadata, dict):
-            metadata = dict(existing_metadata)  # Create a copy
-        else:
-            metadata = {}
+def _store_summary(df, summary: str, model_name: str):
+    """Add summary to metadata and store in df"""
+    # hardcoded heuristic to store everything on chunk 0's metadata
```
nv-rliu, line 186: Previous discussion concluded that we can store things on page_0 for now

api/src/udfs/llm_summarizer_udf.py
drobison00: This seems fine to me, it keeps the results consistent.

api/src/udfs/llm_summarizer_udf.py
drobison00: This is probably fine, especially since its a UDF based approach. That said, I can see some reasons why we might want to add a 'summarization' primitive or something similar for a job, where there is only one, and it holds any kind of aggregated data for a document.

api/src/udfs/llm_summarizer_udf.py
```diff
@@ -44,167 +55,169 @@ def content_summarizer(control_message: "IngestControlMessage") -> "IngestContro
...
-    from openai import OpenAI
-
-    logger = logging.getLogger(__name__)
     logger.info("UDF: Starting LLM content summarization")
 
-    # Get configuration from environment
-    api_key = os.getenv("NVIDIA_API_KEY", "")
-    model_name = os.getenv("LLM_SUMMARIZATION_MODEL", "nvidia/llama-3.1-nemotron-70b-instruct")
+    api_key = os.getenv("NVIDIA_API_KEY")
+    model_name = os.getenv("LLM_SUMMARIZATION_MODEL", "nvdev/nvidia/llama-3.1-nemotron-70b-instruct")
```
jioffe502: do we post nvdev endpoints to prod?

nv-rliu: we definitely do not. my fault :')

PR #1076, opened 2025-10-05
Add support for filenames with brackets

ChrisJar:
## Description

This PR escapes brackets when reading in filenames

## Checklist


drobison00 (changes requested):
Can we add a unit test for this.

ChrisJar:
@drobison00 I'm realizing that this is going to break the ability to do bracket based pattern matching eg: "file[^0-9].pdf") The reason for this PR was that user was having an issue performing an extraction on a filename that included brackets eg: "123[samplefile].pdf". I'm wondering if we should just warn users that we use glob patterns and direct users to escape their filepaths eg: write "123[samplefile].pdf" as "123[[]samplefile[]].pdf"

PR #1083, opened 2025-10-08
Add support for embedding custom content fields

ChrisJar:
## Checklist

### DIFF
--- a/api/src/nv_ingest_api/internal/transform/embed_text.py
+++ b/api/src/nv_ingest_api/internal/transform/embed_text.py
@@ -282,6 +282,32 @@ def _add_embeddings(row, embeddings, info_msgs):
     return row
 
 
+def _add_custom_embeddings(row, embeddings, result_target_field):
+    """
+    Updates a DataFrame row with embedding data and associated error info
+    based on a user supplied custom content field.
+
+    Parameters
+    ----------
+    row : pandas.Series
+        A row of the DataFrame.
+    embeddings : dict
+        Dictionary mapping row indices to embeddings.
+
+    Returns
+    -------
+    pandas.Series
+        The updated row with 'embedding', 'info_message_metadata', and
+        '_contains_embeddings' appropriately set.
+    """
+    embedding = embeddings.get(row.name, None)
+
+    # Always set embedding, even if None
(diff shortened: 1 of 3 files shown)


api/src/nv_ingest_api/internal/transform/embed_text.py
```diff
@@ -612,4 +652,43 @@ def _content_type_getter(row):
...
+                zip(
+                    extracted_custom_content.loc[valid_custom_content_mask].index,
+                    custom_content_embeddings.get("embeddings", []),
+                )
+            )
+        else:
+            custom_embeddings_dict = {}
+
+        combined_df = combined_df.apply(
+            _add_custom_embeddings, embeddings=custom_embeddings_dict, result_target_field=result_target_field, axis=1
```
ChrisJar, line 694: How do we want to go about including info messages from the custom embedding pass? Should I append them to any info messages that may have been generated from the first pass (embedding the standard primitives)

api/api_tests/internal/schemas/transform/test_transform_text_embedding_schema.py
```diff
@@ -43,6 +47,8 @@ def test_text_embedding_schema_accepts_custom_values():
     assert schema.input_type == "query"
     assert schema.raise_on_failure is True
     assert schema.truncate == "START"
+    assert schema.custom_content_field == "custom_content"
```
drobison00, line 50: This is going to need to be a bit more robust.

In general, UDFs will read/write from something like 
custom_content_field: `metadata.custom_content.path.to.nested.element`
result_target_field: `metadata.custom_content.path.to.output.element`

drobison00: https://pypi.org/project/jsonpath-ng/

ChrisJar: Ahh I see, I'll look into that

api/src/nv_ingest_api/internal/transform/embed_text.py
drobison00: They can just be logged to warning/info I think.

drobison00 (approved):

PR #1090, opened 2025-10-10
Fix milvus polling bug and skip polling for local milvus-lite

edknv:
## Description
- In `wait_for_index`, `pos_movement` is now properly decremented on no progress. Previously it was set to -1, which didn't actually count down and could prematurely or never trigger the watchdog.
- When using a local index (`.db` and milvus-lite), `indexed_rows` doesn't advance during streaming inserts. We now skip the `wait_for_index` call in local mode.

## Checklist


client/src/nv_ingest_client/util/vdb/milvus.py
```diff
@@ -1046,9 +1048,10 @@ def write_to_nvingest_collection(
             client,
             collection_name,
         )
-        # Make sure all rows are indexed, decided not to wrap in a timeout because we dont
-        # know how long this should take, it is num_elements dependent.
-        wait_for_index(collection_name, num_elements, client)
+        if not local_index:
+            # Make sure all rows are indexed, decided not to wrap in a timeout because we dont
+            # know how long this should take, it is num_elements dependent.
+            wait_for_index(collection_name, num_elements, client, local_index=local_index)
```
jperez999: Why do we need to pass local_index to this function if we know already with if statement that we either should or should not call this function. I dont think we need both, do we? (maybe I am missing something here)

edknv: Oops, you are right, something I was trying out got in there. Thanks for catching this.

jperez999 (approved):

PR #1101, opened 2025-10-15
remove openai requirement for embedding

jperez999:
## Description
This removes dependency of OpenAI for the text embedder. 

## Checklist

### DIFF
--- a/api/api_tests/internal/transform/test_embed_text.py
+++ b/api/api_tests/internal/transform/test_embed_text.py
@@ -400,10 +400,7 @@ def test_make_async_request_happy_path(im_mock):
         dtypes=["BYTES"],
     )
 
-    assert result == {
-        "embedding": [[0.1, 0.2, 0.3]],
-        "info_msg": None
-    }
+    assert result == {"embedding": [[0.1, 0.2, 0.3]], "info_msg": None}
 
 
 @patch("nv_ingest_api.internal.transform.embed_text.infer_microservice")


api/src/nv_ingest_api/internal/transform/embed_text.py
```diff
@@ -78,23 +79,37 @@ def _make_async_request(
...
+        #     extra_body["modality"] = modalities
+
+        # resp = client.embeddings.create(
+        #     input=prompts,
+        #     model=embedding_model,
+        #     encoding_format=encoding_format,
+        #     extra_body=extra_body,
+        # )
+
+        resp = infer_microservice(
```
drobison00, line 85: For consistency, this really needs to be a NimClient, so that we're able to use the same configuration process that we do for all the other NIM elements.

drobison00: Something like this but, I would imagine much less pre/post processing code.

https://github.com/NVIDIA/nv-ingest/blob/main/api/src/nv_ingest_api/internal/primitives/nim/model_interface/yolox.py#L92

jperez999: I am a little confused here, because this is running a NIM client. Its just surrounded by a helper function in order to make it available outside of the NIM classes. https://github.com/NVIDIA/nv-ingest/pull/1101/files#diff-134970881072236d7f90878a5b6a8676c477ff4f6814bd6f9a272b6addd85570R132 NimClient is here.

PR #1106, opened 2025-10-17
Fix LLM UDF Bug

nv-rliu:
## Description
Fixes a bug accidentally introduced to https://github.com/NVIDIA/nv-ingest/pull/1067

The DF that is stored back onto the payload is not the original DF, rather, the truncated DF with the chunks selected just for summarization. This is incorrect and should be the original DF but with the summary for that document stored on the first row. 

## Checklist

### DIFF
--- a/api/src/udfs/llm_summarizer_udf.py
+++ b/api/src/udfs/llm_summarizer_udf.py
@@ -57,7 +57,7 @@ def content_summarizer(control_message: "IngestControlMessage") -> "IngestContro
     logger.info("UDF: Starting LLM content summarization")
 
     api_key = os.getenv("NVIDIA_API_KEY")
-    model_name = "moonshotai/kimi-k2-instruct-0905"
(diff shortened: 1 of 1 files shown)


api/src/udfs/llm_summarizer_udf.py
```diff
@@ -57,7 +57,7 @@ def content_summarizer(control_message: "IngestControlMessage") -> "IngestContro
     logger.info("UDF: Starting LLM content summarization")
 
     api_key = os.getenv("NVIDIA_API_KEY")
-    model_name = os.getenv("LLM_SUMMARIZATION_MODEL", "nvidia/llama-3.1-nemotron-70b-instruct")
+    model_name = "moonshotai/kimi-k2-instruct-0905"
```
rlratzel: Just wondering if this was left from a debugging session or if you intentionally meant to remove support for reading the env var.

nv-rliu: Ah this should be changed. Good point.

PR #1115, opened 2025-10-22
Add QoS Scheduler + Simple V2 QoS Routing

drobison00:
## Checklist


src/nv_ingest/api/v2/ingest.py
```diff
@@ -44,6 +45,42 @@
 
 DEFAULT_PDF_SPLIT_PAGE_COUNT = 32
 
+# Default QoS thresholds (pages). Tunable via environment variables:
+# QOS_MAX_PAGES_MICRO, QOS_MAX_PAGES_SMALL, QOS_MAX_PAGES_MEDIUM
+_QOS_DEFAULTS = {
```
jdye64, line 50: I appreciate this being tunable since I'm sure this target will move over time.

src/nv_ingest/api/v2/ingest.py
```diff
@@ -631,6 +668,51 @@ def _build_aggregated_response(
...
+    - V2_SUBMIT_BURST_JITTER_MS (default: 10)
+    """
+    burst_size = int(os.getenv("V2_SUBMIT_BURST_SIZE", "16"))
+    pause_ms = int(os.getenv("V2_SUBMIT_BURST_PAUSE_MS", "50"))
+    jitter_ms = int(os.getenv("V2_SUBMIT_BURST_JITTER_MS", "15"))
+
+    return max(1, burst_size), max(0, pause_ms), max(0, jitter_ms)
+
+
+async def _submit_subjobs_in_bursts(
```
jdye64, line 761: I like this function too

src/nv_ingest/pipeline/default_pipeline_impl.py
```diff
@@ -428,76 +428,76 @@
...
   - from: "source_stage"
     to: "metadata_injector"
-    queue_size: 32
+    queue_size: 4
 
   # Document Extractors
   - from: "metadata_injector"
     to: "pdf_extractor"
-    queue_size: 32
+    queue_size: 8
```
jdye64, line 436: Curious why this change is needed? Not just for this queue_size but all of them in this file

jdye64 (approved):
Had one question about queue_size changes but other than that changes make a lot of sense and seem simple enough (in a good way) given the changes.

src/nv_ingest/pipeline/default_pipeline_impl.py
drobison00: The core purpose here was just to reduce the number of non-productive jobs in the pipeline at any given time; in a situation where you have a large number of queued jobs in the broker, this will reduce the number that we queue up in the pipeline, while still being able to feed the stages at line rate.

If we imagine a large file case, where the 'large' queue is basically never empty, the effect here is that when we see a small/micro job, and pull it off, there will be less 'idle' work ahead of it in the pipeline. If we see performance drop, it would mean we're starving the downstream stages, but so far that has not been the case.

src/nv_ingest/api/v2/ingest.py
drobison00: Yeah, I'd almost expect it to turn into a dynamic/heuristic at some point.

PR #1119, opened 2025-10-23
Dataloader video ingest pipeline support

jperez999:
## Description
This PR adds dataloader support to handle video files that need to be converted and split into predetermined sizes to lower load on service.

## Checklist

### DIFF
--- a/api/src/nv_ingest_api/util/service_clients/rest/rest_client.py
+++ b/api/src/nv_ingest_api/util/service_clients/rest/rest_client.py
@@ -156,7 +156,7 @@ class RestClient(MessageBrokerClientBase):
         # Validate and normalize API version to prevent misconfiguration
         # Default to v1 for backwards compatibility if not explicitly provided
         VALID_API_VERSIONS = {"v1", "v2"}
-        raw_api_version = kwargs.get("api_version", "v2")
+        raw_api_version = kwargs.get("api_version", "v1")
         api_version = str(raw_api_version).strip().lower()
 
         if api_version not in VALID_API_VERSIONS:


src/nv_ingest/api/v2/ingest.py
```diff
@@ -556,67 +627,110 @@ async def submit_job_v2(
...
+            parent_metadata = {"total_pages": page_count}
+
+        elif document_types and payloads and document_types[0].lower() in ["mp4", "mov", "avi", "mp3", "wav"]:
+            # print("IN AUDIO/VIDEO BLOCK")
+            document_type = document_types[0]
+            upload_path = f"./{Path(original_source_id).name}"
+            # dump the payload to a file, just came from client
+            with open(upload_path, "wb") as f:
+                f.write(base64.b64decode(payloads[0]))
+            dataloader = DataLoader(
```
drobison00, line 964: Do we always want to assume the full file is sent through the REST service?

src/nv_ingest/api/v2/ingest.py
```diff
@@ -556,67 +627,110 @@ async def submit_job_v2(
...
+                path=upload_path, output_dir="./audio_chunks/", audio_only=True, split_interval=100000000
```
drobison00: This basically assumes that the files can/should be stored on the local disk, in the REST service's context. At a minimum, I think we need a configuration parameter describing where to land files in the container -- and some way of indicating how/when they should be removed.

I'm still in favor of trying to leverage Ray and its Plasma store for shared objects, letting it handle disk spilling and data movement.

The  Ingest service could be responsible for starting its own Registry actor when it comes up, so that its able to keep object refs alive. This would also give us semantics for expiring objects, so we don't just keep writing files to disk and never cleaning up.

```python
@ray.remote
class PlasmaRegistry:
    def __init__(self):
        # key -> (ObjectRef, expiry_ts or None)
        self._store: Dict[str, Tuple[ray.ObjectRef, Optional[float]]] = {}

    def put(self, key: str, obj_ref: "ray.ObjectRef", ttl_seconds: Optional[float] = None) -> None:
        expiry = time.time() + ttl_seconds if ttl_seconds else None
        self._store[key] = (obj_ref, expiry)

    def get(self, key: str) -> Optional["ray.ObjectRef"]:
        item = self._store.get(key)
        if not item:
            return None
        obj_ref, expiry = item
        if expiry and time.time() > expiry:
            # expired; remove and return None
            self._store.pop(key, None)
            return None
        return obj_ref

    def delete(self, key: str) -> bool:
        return self._store.pop(key, None) is not None

    def exists(self, key: str) -> bool:
        return key in self._store

# Helper to get/create the named actor once
def get_plasma_registry_actor(name: str = "nv_ingest_plasma_registry"):
    try:
        return ray.get_actor(name)
    except ValueError:
        return PlasmaRegistry.options(name=name, lifetime="detached").remote()

```

src/nv_ingest/api/v2/ingest.py
jperez999: When we are processing the file we need to have all the bytes. In the original dataloader PR I explained that for the video files we are using you need the whole file, because the end of the file is where are the metadata about the file is stored. Otherwise the bytes come in and we dont really know how to create the parts if we dont have the original files metadata. This is why when the file is local and we use fsspec to pass file handle to ffmpeg it works, but when you are working with remote files, ffmpeg hangs on the files that are getting streamed. I will work to create a small reproducer that shows this error. I reported this behavior when I first encountered the issues (this is one of the places that talks about the issue I was seeing https://ffmpeg.org/pipermail/ffmpeg-user/2020-May/048550.html)

drobison00 (approved):

PR #1130, opened 2025-10-28
Add support for page-elements-v3

edknv:
## Description
This PR upgrades yolox page-elements to v3 semantics and introduces an OCR text extraction pipeline. 

- The `YoloxPageElementsModelInterface` has been updated to recognize the new text-based classes from page-elements-v3 (`paragraph`, `title`, `header_footer`).
- A dedicated Ray actor stage has been created to perform OCR on text-image primitives. It aggregates text blocks on a per-page basis, sorts them into the correct reading order, and constructs a final, page-level text primitive.
- To provide users with flexible control, two new `extract_method` options have been added:
  1. `pdfium_hybrid`: A hybrid approach that uses pdfium for text-based pages and automatically switches to the OCR pipeline for pages identified as scanned.
  2. `ocr`: An OCR-first approach that processes every page through the full OCR pipeline.
- Backward-compatible, continuing to support page-elements-v2; compatible with both PaddleOCR and NR-OCR.

bo767 e2e performance comparison (on A100 with NR-OCR)
`extract_method="pdfium"` (default): 16.8 pages/sec
`extract_method="pdfium_hybrid"`: 12.4 pages/sec
`extract_method="ocr"`: 5.2 pages/sec

## Checklist

### DIFF
--- a/api/src/nv_ingest_api/internal/extract/image/ocr_extractor.py
+++ b/api/src/nv_ingest_api/internal/extract/image/ocr_extractor.py
@@ -363,10 +363,8 @@ def extract_text_data_from_image_internal(
         # Identify rows that meet the text criteria.
         page_images_mask = df_extraction_ledger.apply(_meets_page_image_criteria, axis=1)
         page_elements_mask = df_extraction_ledger.apply(_meets_page_elements_text_criteria, axis=1)
-        df_extraction_ledger.to_csv("/workspace/data/df_extraction_ledger.csv")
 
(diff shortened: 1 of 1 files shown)


src/nv_ingest/pipeline/default_pipeline_impl.py
```diff
@@ -461,6 +482,9 @@
     to: "chart_extractor"
     queue_size: 4
   - from: "chart_extractor"
+    to: "ocr_extractor"
+    queue_size: 32
```
drobison00: This should be OK at 8-16 unless there is a specific need for higher.

PR #1155, opened 2025-11-12
Updated summarization UDF and created a new pipeline for this UDF

KyleZheng1284:
## Description
This PR upgrades the existing Summarization UDF implemenation to be more dynamic and useful in enterprise level settings

- Updates the existing summarization UDF with a default model change to `nemotron-mini-4b-instruct` which offers significant speed up
- Patched bug in previous implementation as only summarization chunks were being sent to milvus
-  Added a new ray no-op stage for this UDF to allow higher concurrent API calls - allowing for a faster speed up in overall summarization time
- Created `config/custom_summarization_pipeline.yaml` that loads the summarization pipeline, also serves as a baseline example for how it can be reproduced in other workflows
- Added in `INGEST_CONFIG_PATH=` flag within docker compose to allow the override of the default implementation file
- Updated Docs for these changes in `examples/udfs/README.md`

## Checklist

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -9,7 +9,7 @@ services:
       - "6379:6379"
 
   page-elements:
-    image: ${YOLOX_IMAGE:-nvcr.io/nim/nvidia/nemoretriever-page-elements-v2}:${YOLOX_TAG:-latest}
+    image: ${YOLOX_IMAGE:-nvcr.io/nim/nvidia/nemoretriever-page-elements-v3}:${YOLOX_TAG:-latest}
     shm_size: 16gb
     ports:
       - "8000:8000"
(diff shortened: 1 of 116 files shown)


docker-compose.yaml
```diff
@@ -275,6 +275,7 @@ services:
       - AUDIO_GRPC_ENDPOINT=audio:50051
       - AUDIO_INFER_PROTOCOL=grpc
       - CUDA_VISIBLE_DEVICES=-1
+      - USE_SUMMARIZATION_PIPELINE=false
```
KyleZheng1284: This is my current implementation on how to specify the runtime container to use the new summarization implementation file that I defined instead of the default.

Please let me know if there is an easier approach to override this

src/nv_ingest/framework/orchestration/ray/stages/meta/udf_parallel_helper.py
drobison00, line 1: I think here, we should make the stage naming more generic. There is nothing specific to summarization for this stage. Let's put it in something like meta/udf_parallel_helper

drobison00 (changes requested):

PR #1156, opened 2025-11-12
Add Local Disk Storage Support by fsspec to Image Storage Stage

jioffe502:
## Description

Enables saving extracted images (tables, charts, structured content) to local disk via fsspec in addition to MinIO. Uses existing volume mounts. Refactors pipeline to be fully fsspec native with no performance difference from existing code

Future work includes:
parallelizing this idempotent operation in the ray pipeline

## Summary
Replaced the old dual-toggle storage system with a single fsspec/UPath pipeline. Everything now writes via storage_uri, so switching between MinIO and disk is just changing that one value.

Defaulted the entire stack (compose + Helm) to our bundled MinIO bucket (s3://nv-ingest/artifacts/store/images). Disk persistence is still supported by opting into file:///... per job or via environment overrides.

Automatically inject MinIO credentials and endpoint (MINIO_ACCESS_KEY/SECRET_KEY, MINIO_INTERNAL_ADDRESS → http://minio:9000) so the local object store works out of the box—no extra config required. You can still point at external S3/MinIO by overriding the env vars.

Added optional IMAGE_STORAGE_PUBLIC_BASE_URL so metadata can surface HTTP links (e.g., the MinIO console URL) instead of raw s3:// URIs.

Included s3fs in the runtime/client/api environments so S3 uploads work without manual installs, and documented the new storage knobs in compose, Helm values, test configs, and the e2e runner.

## Changes of Note
Core: image_upload.py now uses UPath/fsspec; ImageStorageStage builds storage options from env + per-task overrides; client StoreTask and schemas accept storage_uri, storage_options, public_base_url.

Config: docker-compose.yaml, Helm values, default pipeline, and tests all reference IMAGE_STORAGE_URI / IMAGE_STORAGE_PUBLIC_BASE_URL as the single source of truth (defaults to our MinIO path).

Docs/tests: Updated e2e runner comments, test configs, and CLI tests to reflect the new interface. Added s3fs dependency to the conda/pyproject environments.

## Testing
Unit suites:
pytest api/api_tests/internal/store/test_image_upload.py api/api_tests/internal/schemas/store/test_store_image_schema.py

PYTHONPATH=... pytest client/client_tests/primitives/tasks/test_store.py client/client_tests/client/test_interface.py client/client_tests/cli/util/test_click.py

E2E: Bo20 dataset in all three modes (no-store baseline, disk, MinIO) to verify throughput and storage behavior; confirmed uploads land in /workspace/data/... for disk and nv-ingest/artifacts/store/images in MinIO.

## Checklist

### DIFF
--- a/api/pyproject.toml
+++ b/api/pyproject.toml
@@ -26,6 +26,7 @@ dependencies = [
     "pydantic>2.0.0",
     "pydantic-settings>2.0.0",
     "fsspec>=2025.5.1",
+    "universal_pathlib>=0.2.6",
     "ffmpeg-python==0.2.0",
     "tritonclient",
     "glom",


api/api_tests/internal/store/test_image_upload.py
```diff
@@ -39,6 +40,8 @@ def dummy_task_config():
         "access_key": "key",
         "secret_key": "secret",
         "endpoint": "mock_endpoint",
+        "enable_minio": True,
```
drobison00: If both are enabled do we store in both places (do we want to?)? If not, this should be a string or enum to indicate the storage target.

api/src/nv_ingest_api/interface/store.py
```diff
@@ -196,6 +207,9 @@ def store_images_to_minio(
 
     # Build the task configuration as a flat dictionary, matching the internal function's expectations.
     task_config = {
+        "enable_minio": enable_minio,
+        "enable_local_disk": enable_local_disk,
+        "local_output_path": local_output_path,
```
drobison00: If were using fsspec, this may not be local (?). Could be an S3 target.

api/src/nv_ingest_api/internal/schemas/store/store_image_schema.py
```diff
@@ -17,14 +17,29 @@
...
 
+_DEFAULT_STORAGE_URI = os.environ.get("IMAGE_STORAGE_URI", "s3://nv-ingest/artifacts/store/images")
+
+
 class ImageStorageModuleSchema(BaseModel):
     structured: bool = True
     images: bool = True
+    storage_uri: str = Field(default_factory=lambda: _DEFAULT_STORAGE_URI)
+    storage_options: Dict[str, Any] = Field(default_factory=dict)
+    public_base_url: Optional[str] = None
```
jperez999, line 36: How does this play in, couldn't I get everything I need from the storage_uri? If not, in what cases doesnt that work?

api/src/nv_ingest_api/internal/store/image_upload.py
```diff
@@ -58,13 +41,9 @@ def _upload_images_to_minio(df: pd.DataFrame, params: Dict[str, Any]) -> pd.Data
...
-            - "endpoint": URL for the MinIO service (optional; defaults to _DEFAULT_ENDPOINT).
-            - "bucket_name": Bucket name for storing objects (optional; defaults to _DEFAULT_BUCKET_NAME).
-            - "access_key": Access key for MinIO.
-            - "secret_key": Secret key for MinIO.
-            - "session_token": Session token for MinIO (optional).
-            - "secure": Boolean indicating if HTTPS should be used.
-            - "region": Region for the MinIO service (optional).
+            - "storage_uri": Base URI (file://, s3://, etc.) where images should be written.
+            - "storage_options": Optional dictionary forwarded to UPath/fsspec constructors.
+            - "public_base_url": Optional HTTP(s) base used to surface stored objects.
```
jperez999, line 83: if I want to send this to an http service the storage uri would still have the http scheme, so I dont see why we need the  public_base_url?

api/src/nv_ingest_api/internal/store/image_upload.py
```diff
@@ -134,35 +106,61 @@ def _upload_images_to_minio(df: pd.DataFrame, params: Dict[str, Any]) -> pd.Data
...
+            if public_base_url:
+                public_url = f"{public_base_url.rstrip('/')}/{relative_key}"
+
+            primary_uri = public_url or destination_uri
+            source_metadata["source_location"] = primary_uri
+
+            local_uri: Optional[str] = None
+            if protocol == "file":
+                local_uri = destination.path
+                source_metadata["local_source_location"] = local_uri
```
jperez999, line 165: Why do we need a separate key holding the local_source_location, wouldnt this be the same as source_location?

api/src/nv_ingest_api/internal/store/image_upload.py
```diff
@@ -134,35 +106,61 @@ def _upload_images_to_minio(df: pd.DataFrame, params: Dict[str, Any]) -> pd.Data
...
-                bucket_name,
-                destination_file,
-                source_file,
-                length=len(content),
-            )
-
-            # Construct the public URL
-            public_url = f"{_DEFAULT_READ_ADDRESS}/{bucket_name}/{destination_file}"
-            source_metadata["source_location"] = public_url
+            # Extract just the filename from source_id for cleaner organization
```
jperez999: This is a clear action we are taking here. Is there a way to pull the logic concerning creating a destination path into its on function. This might be something useful we can leverage in another task down the road.

api/src/nv_ingest_api/internal/store/image_upload.py
```diff
@@ -134,35 +106,61 @@ def _upload_images_to_minio(df: pd.DataFrame, params: Dict[str, Any]) -> pd.Data
...
                 image_metadata = metadata.get("image_metadata", {})
-                image_type = image_metadata.get("image_type", "png")
+                raw_image_type = image_metadata.get("image_type", "png")
+                # Handle both enum and string values
+                if hasattr(raw_image_type, "value"):
+                    # It's an enum, get the string value
+                    image_type = raw_image_type.value.lower()
+                else:
+                    image_type = str(raw_image_type).lower()
+            elif doc_type == ContentTypeEnum.STRUCTURED:
```
jperez999: all of this doc_type logic both here and below, can it be pulled out into a function as well?

src/nv_ingest/framework/orchestration/ray/stages/storage/image_storage.py
```diff
@@ -98,3 +129,38 @@ def on_data(self, control_message: IngestControlMessage) -> IngestControlMessage
...
+
+        if parsed_scheme not in {"s3", "s3a", "s3n"}:
+            return merged_options
+
+        def _set_if_absent(key: str, env_var: str) -> None:
+            if key not in merged_options and env_var in os.environ:
+                merged_options[key] = os.environ[env_var]
+
+        _set_if_absent("key", "MINIO_ACCESS_KEY")
+        _set_if_absent("secret", "MINIO_SECRET_KEY")
```
jperez999, line 149: Why do we need to set this here? shouldnt we allow the user to provide us the information inside the task? I feel like this can be dangerous, especially if we have multiple users with different options on the same service.

api/src/nv_ingest_api/internal/store/image_upload.py
jioffe502: public_base_url is optional and only used when the place we write objects (storage_uri, often private MinIO/S3) differs from the URL we want to hand to users (a CDN or proxy). If you’re writing directly to an HTTP endpoint, just leave public_base_url unset and we’ll keep using the same HTTP URI for reads.

api/src/nv_ingest_api/internal/store/image_upload.py
jioffe502: This is defensive for if a public_base_url is provided. maybe for simplicity we can remove the option of the public_base_url and then this would also change

api/src/nv_ingest_api/internal/store/image_upload.py
jioffe502: Yep. split that logic out into _build_destination_path so constructing the on-disk key (safe source name, subdir, relative key) is reusable. _upload_images_via_fsspec now calls the helper with storage_root, source_id, row index, and image type, and the helper returns both the destination UPath and the relative key, so any future task that needs the same layout can call it directly.

api/src/nv_ingest_api/internal/store/image_upload.py
jioffe502: I pulled the shared doc-type/image-type logic into a helper named _extract_image_type. 

_upload_images_via_fsspec now calls that function for every row, so both IMAGE and STRUCTURED cases reuse the same normalization code (enum vs string, defaults, etc.), and future doc types can plug in there without duplicating the branching.

src/nv_ingest/framework/orchestration/ray/stages/storage/image_storage.py
jioffe502: Right now inject_storage_defaults only fills in creds/endpoint if the caller didn’t specify them. We merge user-provided storage_options (including per-task params) on top, so a tenant can pass their own key/secret/token/client_kwargs and the env defaults never win. The helper just keeps default deployments (where only MINIO_* envs exist) working without extra wiring.

api/src/nv_ingest_api/internal/store/image_upload.py
jperez999: Since it is None, we can leave it and so if it is necessary, as you said in the previous comment, the person can supply the correct value (i.e. proxy)

config/default_pipeline.yaml
```diff
@@ -335,6 +335,9 @@ stages:
     type: "stage"
     phase: 5  # RESPONSE
     actor: "nv_ingest.framework.orchestration.ray.stages.storage.image_storage:ImageStorageStage"
+    config:
```
jperez999, line 338: when you use this it maps to minio locally? Or is this going out to a bucket on aws? Would that cost money?

jioffe502: Yep image_storage just honors IMAGE_STORAGE_URI, and in dev that defaults to the bundled MinIO service (s3://nv-ingest/... inside the compose network). No AWS calls or costs unless someone overrides the env var to point at a real bucket.

PR #1163, opened 2025-11-17
Add sysreq doc

drobison00:
## Checklist

### DIFF
--- a/client/client_tests/client/test_interface.py
+++ b/client/client_tests/client/test_interface.py
@@ -10,6 +10,7 @@ import logging
 import os
 import tempfile
 from concurrent.futures import Future
+from io import BytesIO
 from unittest.mock import ANY
 from unittest.mock import MagicMock
 from unittest.mock import patch
@@ -17,7 +18,7 @@ from unittest.mock import patch
 import nv_ingest_client.client.interface as module_under_test
 import pytest
 
-from client.client_tests.utilities_for_test import (
+from client_tests.utilities_for_test import (
     cleanup_test_workspace,
     create_test_workspace,
     get_git_root,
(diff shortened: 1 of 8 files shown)


docs/docs/extraction/system_requirements.md
```diff
@@ -0,0 +1,31 @@
+# System Requirements
+
+## Hardware Requirements
+
+The NV-Ingest full ingestion pipeline is designed to consume significant CPU and memory resources to achieve maximal parallelism. Resource usage will scale up to the limits of your deployed system.
+
+### Recommended Specifications
```
randerzander:
```suggestion
### Recommended Production Deployment Specifications
```

docs/docs/extraction/system_requirements.md
```diff
@@ -0,0 +1,31 @@
...
+
+- **System Memory**: At least 256 GB RAM
+- **CPU Cores**: At least 32 CPU cores
+- **GPU**: NVIDIA GPU with at least 24 GB VRAM (e.g., A100, V100, or equivalent)
+
+**Note:** Using less powerful systems or lower resource limits is still viable, but performance will suffer.
+
+### Resource Consumption Notes
+
+- The pipeline performs runtime allocation of parallel resources based on system configuration
```
randerzander, line 17: Please cross-link to the dynamic scaling docs

docs/docs/extraction/system_requirements.md
```diff
@@ -0,0 +1,31 @@
...
+- Memory usage can reach up to the full system capacity for large document processing
+- CPU utilization scales with the number of concurrent processing tasks
+- GPU is required for image processing NIMs, embeddings, and other GPU-accelerated tasks
```
randerzander, line 20: Think there should be an additional section at the bottom, but xlinked here.

We should say why the CPU and mem requirements are high-

Something like:
> For a representative set of 1000 PDFs, NV-Ingest renders 54,000 jpeg images, one per PDF page. We extract on average N sub-page jpegs (one each per table, chart, header, footer, section title, and text paragraphs). Downstream of each content type, we extract smaller bounding boxed jpegs for every chart element and every table cell (hundreds to thousands per table).

Can be followup, but needs to tell the user the tl;dr of why we use so many resources.

randerzander: Please also link to whatever public materials we have on DC767 - @sosahi will have this

PR #1176, opened 2025-12-03
Fix libmode verbose logging for MacOS

KyleZheng1284:
This PR adds a default quiet mode for library mode (libmode=True), reducing excessive startup logs and configuring minimal logging while allowing optional verbose output. Tests and documentation were updated accordingly.

When running nv-ingest in library mode, we were seeing excessive startup logs including:

- Replica resolver INFO messages (system memory, stage allocation details)
- Pipeline configuration pretty-print output
- Ray accelerator environment variable warnings
- OpenTelemetry trace export errors when this isn't required for lib mode

Solution 
- Introduced a quiet mode that is enabled by default when running in library mode (libmode=True).
- This automatically configures the environment for minimal logging output while preserving the ability to enable verbose logging for debugging.

This is done in the run_pipeline function and it auto sets `INGEST_RAY_LOG_LEVEL=PRODUCTION`, `RAY_ACCEL_ENV_VAR_OVERRIDE_ON_ZERO=0`, and `OTEL_SDK_DISABLED=true`

For Quick testing 

```
# Setup (one-time)
cd path to nv-ingest
uv venv --python 3.12 nvingest
source nvingest/bin/activate
# Create wheels based on src code with latest changes
uv pip install src/ api/ client/

# Set API key
export NVIDIA_API_KEY="nvapi-..."

# Run the test
python examples/launch_libmode_and_run_ingestor.py
```

I have also updated the test scripts and documentation associated with library mode to match up with the latest version.
- Tested on my local MacOS machine and also on my linux box and it seemed to work.


examples/launch_libmode_service.py
```diff
@@ -24,12 +23,13 @@ def main():
...
     """
     try:
+        # Start pipeline and block until interrupted
+        # Note: stdout/stderr cannot be passed when run_in_subprocess=True (not picklable)
+        # Use quiet=False to see verbose startup logs
         _ = run_pipeline(
             block=True,
             disable_dynamic_scaling=True,
             run_in_subprocess=True,
-            stderr=sys.stderr,
```
KyleZheng1284: When `run_in_process=True ` , `stdout=sys.stdout` and `stderr=sys.stderr` were throwing errors. due to how `multiprocessing.Process` pickles arguments, `TextIOWrapper` can't be pickled

examples/launch_libmode_and_run_ingestor.py
```diff
@@ -64,13 +63,15 @@ def main():
...
+    pipeline = None
     try:
+        # Start pipeline in subprocess
+        # Note: stdout/stderr cannot be passed when run_in_subprocess=True (not picklable)
+        # Use quiet=False to see verbose startup logs
         pipeline = run_pipeline(
             block=False,
             disable_dynamic_scaling=True,
             run_in_subprocess=True,
-            stdout=sys.stdout,
```
KyleZheng1284: When run_in_process=True  , stdout=sys.stdout and stderr=sys.stderr were throwing errors. due to how multiprocessing.Process pickles arguments, TextIOWrapper can't be pickled

src/nv_ingest/framework/orchestration/process/execution.py
```diff
@@ -318,14 +323,15 @@ def launch_pipeline(
...
+        logger.debug("Dynamic scaling disabled via function parameter override")
 
     # Resolve static replicas
     pipeline_config = resolve_static_replicas(pipeline_config)
 
     # Pretty print the final pipeline configuration (after replica resolution)
+    # Use DEBUG level to avoid verbose output during normal startup
     pretty_output = pretty_print_pipeline_config(pipeline_config, config_path=None)
-    logger.info("\n" + pretty_output)
+    logger.debug("\n" + pretty_output)
```
edknv: Will this hide the pipeline config for the standard docker/helm microservice deployment? It seem like useful info to display.

examples/launch_libmode_and_run_ingestor.py
```diff
@@ -64,13 +63,15 @@ def main():
...
-            stderr=sys.stderr,
```
edknv, line 73: Does this mean stdout/stderr will go to /dev/null? What happens if there is an error in the pipeline? Does it fail silently?

edknv:
Tagging @nkmcalli to request documentation review on `ocs/docs/extraction/quickstart-library-mode.md`.

src/nv_ingest/framework/orchestration/process/execution.py
KyleZheng1284: Yes, I also agree.

I reverted all of the logger statements back to the original debug in case there was a specific reason why we were keeping them. With the new changes with the `quiet` mode for library mode that I created it seems to do a good enough silencing verbose logs as is.

edknv (approved):

nkmcalli (approved):

PR #1201, opened 2025-12-11
Package testing framework as Python project

charlesbluca:
## Description

Vendors the testing scripts once living under `/scripts/tests/` into a dedicated Python project under `tools/harness`.

This should make it easier to encode new dependencies as they're introduced to the test framework, and preempts the expansion of the test runner logic to include other managed service scenarios (thinking Helm, full OSS)

## Checklist

### DIFF
--- a/client/src/nv_ingest_client/util/vdb/milvus.py
+++ b/client/src/nv_ingest_client/util/vdb/milvus.py
@@ -1049,7 +1049,8 @@ def write_to_nvingest_collection(
     )
     num_elements = len(cleaned_records)
     if num_elements == 0:
-        raise ValueError("No records with Embeddings to insert detected.")
+        logger.warning("No records with Embeddings to insert detected.")
+        return
     logger.info(f"{num_elements} elements to insert to milvus")
     logger.info(f"threshold for streaming is {threshold}")
     if num_elements < threshold:
--- a/tools/harness/src/nv_ingest_harness/utils/recall.py
+++ b/tools/harness/src/nv_ingest_harness/utils/recall.py
@@ -503,16 +503,18 @@ def get_recall_collection_name(test_name: str) -> str:
     """
     Generate collection name for recall evaluation.
 
-    All recall evaluations use a single multimodal collection with the pattern:
-    {test_name}_multimodal
+    For normal datasets, uses the pattern {test_name}_multimodal.
+    If the input is a single file (contains '.' or ends with '.pdf' or similar),
+    it converts the period, which is not allowed in Milvus collection names, to an underscore.
 
     Args:
-        test_name: Test identifier (e.g., 'bo767', 'finance_bench')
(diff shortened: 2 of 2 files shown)


tools/harness/test_configs.yaml
```diff
@@ -61,10 +61,10 @@ recall:
 
 # Pre-configured datasets
 # Each dataset includes path, extraction settings, and recall evaluator
-# Use: python run.py --case=e2e --dataset=bo767
+# Use: uv run nv-ingest-harness-run --case=e2e --dataset=bo767
 datasets:
   bo767:
-    path: /path/to/bo767
+    path: /home/charlesb/datasets/bo767
```
charlesbluca:
```suggestion
    path: /path/to/bo767
```

tools/harness/pyproject.toml
```diff
@@ -0,0 +1,33 @@
...
+version = "0.1.0"
+description = "Add your description here"
+readme = "README.md"
+requires-python = ">=3.12"
+dependencies = [
+    "docker>=7.1.0",
+    "milvus-lite==2.4.12",
+    "nv-ingest==25.9.0",
+    "nv-ingest-api==25.9.0",
+    "nv-ingest-client==25.9.0",
```
jioffe502: for development purposes, if I am using the latest main with a local dev version like 2025.12.16.dev20251216, I get an error like this:

```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
nv-ingest-harness 0.1.0 requires nv-ingest==25.9.0, but you have nv-ingest 2025.12.16.dev20251216 which is incompatible.
nv-ingest-harness 0.1.0 requires nv-ingest-api==25.9.0, but you have nv-ingest-api 2025.12.16.dev20251216 which is incompatible.
nv-ingest-harness 0.1.0 requires nv-ingest-client==25.9.0, but you have nv-ingest-client 2025.12.16.dev20251216 which is incompatible.
Successfully installed nv-ingest-2025.12.16.dev20251216 nv-ingest-api-2025.12.16.dev20251216 nv-ingest-client-2025.12.16.dev20251216
```

maybe we shouldn't have a hard pin of the release but allow any version >= 26.1.0 or whichever the previous release is

jioffe502: in testing, I actually left it unpinned with:

```
    "nv-ingest",
    "nv-ingest-api",
    "nv-ingest-client"
```

charlesbluca: Good point - unpinned these deps in https://github.com/NVIDIA/nv-ingest/pull/1201/commits/a6c450a636233f6d23649ef465ab1e9fffdccfc9

jioffe502 (approved):
Great conversation about this and excited to see this grow.

PR #1213, opened 2025-12-17
Fixed summarization UDF to work with latest upstream changes

KyleZheng1284:
## Description

Fixed changes within summarization UDF

1. Pre Split (32 pages) was causing multiple summaries per file depending on the length. Example if total pages in PDF is 32-64 it will 2 summaries, if it has 64-96 pages it will have 3 summaries, ... 
   - Added simple logic to skip non-first sections, so only the first and last chunk of the first section (1-32 pages) will be used for summarization. 

2. Fixed custom pipeline to use nemotron_parse* from depreciated nemoretriever-parse*
3. Minor upstream fixes 
   - Now using httpx instead of openai for API calls to build since we have removed that dep with https://github.com/NVIDIA/nv-ingest/pull/1101/changes
   - Changed max token limit to 4000, since the max token limit for the `nvidia/nemotron-mini-4b-instruct` is 4096

## Checklist

### DIFF
--- a/api/src/udfs/llm_summarizer_udf.py
+++ b/api/src/udfs/llm_summarizer_udf.py
@@ -15,7 +15,7 @@ Environment variables (can be treated as kwargs):
 - LLM_SUMMARIZATION_BASE_URL: Base URL for OpenAI-compatible API (default: https://integrate.api.nvidia.com/v1)
 - LLM_SUMMARIZATION_TIMEOUT: API timeout in seconds (default: 60)
 - LLM_MIN_CONTENT_LENGTH: Minimum content length to summarize (default: 50)
-- LLM_MAX_CONTENT_LENGTH: Maximum content length to send to API (default: 12000)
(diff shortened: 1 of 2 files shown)


api/src/udfs/llm_summarizer_udf.py
```diff
@@ -201,24 +210,32 @@ def _generate_llm_summary(
...
-
-        client = OpenAI(base_url=base_url, api_key=api_key, timeout=timeout)
-
-        completion = client.chat.completions.create(
-            model=model_name,
-            messages=[{"role": "user", "content": PROMPT.format(content=content)}],
-            max_tokens=400,
-            temperature=0.7,
-        )
+        url = f"{base_url.rstrip('/')}/chat/completions"
```
KyleZheng1284: Not sure if this was the best approach for this?

jperez999, line 218: Is this the only place this code will be used to call the summarizer? If not, should we consider making this into something that we can call in other places like a class or simply a function?

PR #1220, opened 2025-12-18
docs: Update nemoretriever-->nemotron everywhere but filename

nkmcalli:
Update nemoretriever-->nemotron everywhere but filename


ChrisJar (approved):
lgtm! Thanks!

PR #1240, opened 2025-12-24
Fixing quote env vars with special chars in YAML substitution

jioffe502:
## Description

Fixes a crash where environment variables containing YAML special characters (like colons) caused parsing errors during startup per https://nvbugspro.nvidia.com/bug/5757249

- Added `_quote_if_needed` helper to escape values containing special chars (:, #, [, etc.).
- Updated regex replacer to automatically quote unsafe values while preserving existing quotes.
- Updated unit tests to cover edge cases for YAML injection.

## Checklist


jioffe502:
Sorry @nkmcalli ! There is no explicit doc impact that was a random assignment that occurs when I opened the PR!

PR #1241, opened 2025-12-29
Ray profiling tracing 

jioffe502:
## Description

This is purely the tracing work from https://github.com/NVIDIA/nv-ingest/pull/1168 to simplify the additions to the codebase and make it more tractable. 

It allows in our nv-ingest-test-harness for us to populate and analyze ray traces.

1168 also attempts to combine a pdfium optimization that needs to be further debugged and potentially thrown away. This pdfium optimization will be evaluated in a separate future PR. 

## Checklist

### DIFF
--- a/tools/harness/src/nv_ingest_harness/config.py
+++ b/tools/harness/src/nv_ingest_harness/config.py
@@ -18,6 +18,21 @@ from pathlib import Path
 from typing import Optional, List
 
 
+def get_repo_root() -> Path:
+    """Get the nv-ingest repository root directory."""
+    # config.py is at: tools/harness/src/nv_ingest_harness/config.py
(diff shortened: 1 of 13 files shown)


tools/harness/src/nv_ingest_harness/config.py
```diff
@@ -18,6 +18,21 @@
 from typing import Optional, List
 
 
+def get_repo_root() -> Path:
```
drobison00, line 21: Take a look at something like this for a more robust solution: https://github.com/NVIDIA/nv-ingest/blob/release/25.9.0/tests/utilities_for_test.py#L9

drobison00 (changes requested):

PR #1272, opened 2026-01-07
fix(helm): correct NIM Operator service names for NEMOTRON_PARSE and …

enwaiax:
Bug-Ids: 5788611, 5788715

## Problem
When deploying NV-Ingest with NIM Operator, the default endpoint values in Helm chart values.yaml used incorrect service names:
- NEMOTRON_PARSE_HTTP_ENDPOINT: 'nim-vlm-text-extraction-nemotron-parse'
- AUDIO_GRPC_ENDPOINT: 'nv-ingest-riva-nim'

The actual service names created by NIM Operator templates are:
- 'nemotron-parse' (from nemotron-parse.yaml)
- 'audio' (from audio.yaml)

This mismatch caused DNS resolution failures when these services were enabled.

## Solution
1. Updated default values in values.yaml to use correct service names:
   - NEMOTRON_PARSE_HTTP_ENDPOINT: http://nemotron-parse:8000/v1/chat/completions
   - AUDIO_GRPC_ENDPOINT: audio:50051

2. Added conditional override logic in deployment.yaml (similar to existing VLM_CAPTION_ENDPOINT pattern) to automatically set correct endpoints when nimOperator.nemotron_parse.enabled or nimOperator.audio.enabled is true.

3. Updated README.md documentation to reflect correct default values.

## Verification
- docker-compose.yaml already uses these correct service names
- Python code defaults (default_pipeline_impl.py) use these correct names
- Config files (default_pipeline.yaml) use these correct names
- All deployment methods now consistently use the same service names

## Checklist


helm/templates/deployment.yaml
```diff
@@ -86,6 +86,16 @@ spec:
...
+            {{- if and (hasKey .Values "nimOperator") (hasKey .Values.nimOperator "nemotron_parse") (eq .Values.nimOperator.nemotron_parse.enabled true) }}
+            - name: NEMOTRON_PARSE_HTTP_ENDPOINT
+              value: "http://nemotron-parse:8000/v1/chat/completions"
+            {{- end }}
+
+            {{- if and (hasKey .Values "nimOperator") (hasKey .Values.nimOperator "audio") (eq .Values.nimOperator.audio.enabled true) }}
+            - name: AUDIO_GRPC_ENDPOINT
+              value: "audio:50051"
+            {{- end }}
+
```
charlesbluca, line 98: Is this override required? If I'm following correctly, these should already be defined by the `envVars` in values.yaml, and the need for this logic to define `VLM_CAPTION_ENDPOINT` is only required because [it is commented out in values.yaml](https://github.com/NVIDIA/nv-ingest/blob/a68318869f4c922c53a6c4edaa23aece01e84a7b/helm/values.yaml#L1063)

jdye64: Yeah this isn't needed. Wanna just open another PR removing it? I could see that tripping people up if they made changes in values.yaml and then this overrides those to values they are not expecting.

PR #1275, opened 2026-01-07
feat: add extract_text and extract_images defaults to extract method

sosahi:
Add extraction of extract_text and extract_images from kwargs with default values of True, consistent with extract_tables and extract_charts.

## Checklist

### DIFF
--- a/client/src/nv_ingest_client/util/vdb/milvus.py
+++ b/client/src/nv_ingest_client/util/vdb/milvus.py
@@ -415,7 +415,7 @@ def create_nvingest_collection(
     graph_degree: int = 100,
     m: int = 64,
     ef_construction: int = 512,
-    collection_alias: str = None,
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -6,7 +6,7 @@
 
 ## Improvements
 
-- ...
+- Added `extract_text` and `extract_images` parameter defaults (True) to the Ingestor extract method for consistency with `extract_tables` and `extract_charts`
```
edknv, line 9: I wonder if this is the right place to keep track of this. It looks like CHANGELOG.md has not been updated at all since the initial commit, and this is under `# NVIDIA Ingest 24.08.0`.

sosahi: thanks for catching that, I updated the version to 26.1.0

PR #1282, opened 2026-01-08
docs: Rebrand 'audio extraction' as 'speech extraction'

nkmcalli:
Rebrand 'audio extraction' as 'speech extraction' and make it easier to add video language in the future.

### DIFF
--- a/docs/docs/extraction/audio.md
+++ b/docs/docs/extraction/audio.md
@@ -1,7 +1,8 @@
-# Use NeMo Retriever Extraction with Riva for Audio Processing
+# Extract Speech with NeMo Retriever Extraction
 
 This documentation describes two methods to run [NeMo Retriever extraction](overview.md) 
-with the [RIVA ASR NIM microservice](https://docs.nvidia.com/deeplearning/riva/user-guide/docs/index.html) for processing audio files.
+with the [RIVA ASR NIM microservice](https://docs.nvidia.com/deeplearning/riva/user-guide/docs/index.html) 
+to extract speech from audio files.
 
 - Run the NIM locally by using Docker Compose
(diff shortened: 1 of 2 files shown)


docs/docs/extraction/audio.md
```diff
@@ -1,7 +1,8 @@
-# Use NeMo Retriever Extraction with Riva for Audio Processing
+# Extract Speech with NeMo Retriever Extraction
 
 This documentation describes two methods to run [NeMo Retriever extraction](overview.md) 
-with the [RIVA ASR NIM microservice](https://docs.nvidia.com/deeplearning/riva/user-guide/docs/index.html) for processing audio files.
+with the [RIVA ASR NIM microservice](https://docs.nvidia.com/deeplearning/riva/user-guide/docs/index.html) 
+to extract speech from audio files.
```
sosahi, line 5: should we say "from audio and video files"?

docs/docs/extraction/audio.md
```diff
@@ -10,7 +11,7 @@ with the [RIVA ASR NIM microservice](https://docs.nvidia.com/deeplearning/riva/u
 
     NeMo Retriever extraction is also known as NVIDIA Ingest and nv-ingest.
 
-Currently, you can process the following audio file types:
+Currently, you can extract speech from the following file types:
```
sosahi, line 14: technically we are extracting speech from video files in this release too, right @jperez999 ?

jperez999: Yes but that is ea not ga.

docs/docs/extraction/audio.md
nkmcalli: Not yet since we're not exposing video yet formally.  But I rewrote that so that we could add it easily when we're ready.

PR #1307, opened 2026-01-15
Lancedb Retrieval Rework, harness compatibility, example notebook 

jioffe502:
## Description
This PR updates the retrieval function for our lancedb implementation and integrates it with the harness for testing.

This also adds an initial example notebook similar to the building_vdb_operator. 

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# NVIDIA Ingest 26.1.0
+# NVIDIA Ingest 26.1.2
 
 ## New Features
 
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ Create a fresh Python environment to install nv-ingest and dependencies.
 ```shell
 uv venv --python 3.12 nvingest && \
   source nvingest/bin/activate && \
-  uv pip install nv-ingest==25.9.0 nv-ingest-api==25.9.0 nv-ingest-client==25.9.0 milvus-lite==2.4.12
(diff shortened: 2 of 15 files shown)


tools/harness/src/nv_ingest_harness/utils/recall.py
```diff
@@ -9,13 +9,25 @@
...
 from collections import defaultdict
+from functools import partial
 from typing import Dict, Optional, Callable
 
 from nv_ingest_client.util.milvus import nvingest_retrieval
 
 from nv_ingest_harness.utils.cases import get_repo_root
 
 
+def _get_retrieval_func(vdb_backend: str, table_path: Optional[str] = None):
```
jperez999: should we use this as the action for the get_recall_scores as the place to store all the logic around what retrieval function to select? This works also, just cleaner if all the logic is in one place.

tools/harness/src/nv_ingest_harness/utils/recall.py
```diff
@@ -99,52 +121,62 @@ def get_recall_scores_pdf_only(
...
-        collection_name,
-        hybrid=sparse,
-        embedding_endpoint=f"http://{hostname}:8012/v1",
-        model_name=model_name,
-        top_k=top_k,
-        gpu_search=gpu_search,
-        nv_ranker=nv_ranker,
-        **reranker_kwargs,
-    )
+    if vdb_backend == "lancedb":
```
jperez999, line 212: Given this is in multiple places I feel we should be able to refactor this into a function that we can then call with appropriate parameters to run the correct retrieval and return the results

PR #1360, opened 2026-02-03
Update NIM service configs to reflect latest releases

charlesbluca:
## Description

- With page-elements-v3 using Triton BLS, we should be safe to remove rate limiting / memory pool overrides used to circumvent OOMs we encountered with Triton ensemble
- With OCR using Triton BLS, we no longer have the OOM issues that necessitated `NIM_TRITON_ENABLE_MODEL_CONTROL=1`
- 1.10.x embedding NIMs have deprecated `NIM_TRITON_MAX_SEQ_LENGTH`, meaning this has no impact on VRAM utilization
- Update max util / process workers to reflect more sensible defaults (cc @jioffe502)

## Checklist

### DIFF
--- a/.github/workflows/ci-pull-request.yml
+++ b/.github/workflows/ci-pull-request.yml
@@ -30,32 +30,22 @@ jobs:
 
       - uses: pre-commit/action@v3.0.1
 
-  # Docker build and test for x86_64
-  docker-build-test:
+  # Docker build + test for x86_64 (single job so built image is available locally)
+  docker-build-and-test:
     name: Build & Test Docker (amd64)
-    uses: ./.github/workflows/reusable-docker-build.yml
+    uses: ./.github/workflows/reusable-docker-build-and-test.yml
     with:
       platform: 'linux/amd64'
       target: 'runtime'
-      push: false
       tags: 'nv-ingest:pr-${{ github.event.pull_request.number }}'
       base-image: 'ubuntu'
       base-image-tag: 'jammy-20250415.1'
-      runner: 'linux-large-disk'
-    secrets:
-      HF_ACCESS_TOKEN: ${{ secrets.HF_ACCESS_TOKEN }}
-
-  docker-test:
-    name: Run Tests
-    needs: docker-build-test
-    uses: ./.github/workflows/reusable-docker-test.yml
-    with:
-      image-tag: 'nv-ingest:pr-${{ github.event.pull_request.number }}'
-      platform: 'linux/amd64'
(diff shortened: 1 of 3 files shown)


jioffe502 (approved):

helm/values.yaml
```diff
@@ -149,8 +149,8 @@ envVars:
   OMP_NUM_THREADS: 1
   RAY_num_grpc_threads: 1
 
-  MAX_INGEST_PROCESS_WORKERS: 16
-  NV_INGEST_MAX_UTIL: 48
+  MAX_INGEST_PROCESS_WORKERS: 32
+  NV_INGEST_MAX_UTIL: 32
```
VibhuJawa, line 153: Love it :-D

docker-compose.yaml
```diff
@@ -109,7 +107,6 @@ services:
       - NGC_API_KEY=${NIM_NGC_API_KEY:-${NGC_API_KEY:-ngcapikey}}
       - CUDA_VISIBLE_DEVICES=0
       - NIM_TRITON_MAX_BATCH_SIZE=${OCR_BATCH_SIZE:-32}
-      - NIM_TRITON_ENABLE_MODEL_CONTROL=1
```
VibhuJawa, line 112: Why did we remove this ? Did not this help stability ?

VibhuJawa (commented):
Mostly looks good to me.  Only question around `NIM_TRITON_ENABLE_MODEL_CONTROL`

docker-compose.yaml
charlesbluca: This was helpful for stability back when we had recurring CUDA errors from OCR bugs / Triton ensemble OOMing, but now that we've squashed those bugs and moved to BLS in practice we're no longer needing to reload the OCR NIM out of a "bad state"

Best case scenario is that we never need to switch this back on and enjoy faster NIM startup; worst case scenario is that we eventually see some new CUDA errors that necessitate re-enabling this flag, but IMO would be better to make that case more apparent for now so that we aren't overlooking what would probably be independent CUDA errors from those seen throughout the OCR 1.2 release

WDYT?

VibhuJawa: Okay. Thought about it . I agree with it.  

> worst case scenario is that we eventually see some new CUDA errors that necessitate re-enabling this flag,

Yup. this makes sense, we should fail fast/internally , if we find out we do , we put it back in.

VibhuJawa (approved):
LGTM

PR #1399, opened 2026-02-11
(retriever) Simplify installation

charlesbluca:
## Description

Rebasing on main and updating pyproject to simplify installation now that moviepy is no longer core to nv-ingest-api

## Checklist


retriever/pyproject.toml
charlesbluca, line 1: These are the core changes of this PR outside of the rebasing - now pulling in all the NV-Ingest libs, and installing directly from source so that we get #1395

PR #1483, opened 2026-03-03
Sohail/remove ray jars

sosahi:
## Description

1. Initiated by a flagged HIGH CVE on jackson-core pkg in the nSpect container scan:
>  Remove Ray's Java JAR (ray_dist.jar). It bundles shaded Jackson (e.g. jackson-core) and is only
needed for Ray's Java API / cross-language. This image runs Python-only; removing it drops
the bundled Java deps and reduces image size.

2. also bumped riva-client to the latest version 2.25.0

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -229,6 +229,7 @@ scripts/private_local/
 test_results/
 sweep_results/
 artifacts/
+nemo_retriever/artifacts/
 
 # Ignore cursor rules
 .cursor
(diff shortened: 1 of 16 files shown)


.github/workflows/reusable-integration-test.yml
```diff
@@ -77,7 +77,7 @@ jobs:
             pymilvus \
             'pymilvus[bulk_writer, model]' \
             'milvus-lite==2.4.12' \
-            'nvidia-riva-client==2.20.0' \
+            'nvidia-riva-client==2.25.0' \
```
charlesbluca, line 80: Should we match this in [src/pyproject.toml](https://github.com/NVIDIA/NeMo-Retriever/blob/379ca3fcdd7c9cfca0f836899acc87bc5256dff1/src/pyproject.toml#L68) ?

sosahi: good catch, fixed.

charlesbluca (approved):

PR #1494, opened 2026-03-05
(lib) Add support for vLLM-backed text/VL embedder

charlesbluca:
## Description

Adds vLLM as the default ingest-time backend for local text and VL embedding, replacing the previous HuggingFace-only path. The HF path is preserved as an explicit opt-in for recall/query workloads.

**Core embedder changes:**
- `LlamaNemotronEmbed1BV2Embedder` now uses vLLM's pooling runner by default; `device` parameter deprecated (vLLM follows process-level GPU placement via `CUDA_VISIBLE_DEVICES`)
- New `LlamaNemotronEmbed1BV2HFEmbedder` — HF-only text embedder, used for query/recall evaluation
- New `LlamaNemotronEmbedVL1BV2VLLMEmbedder` — vLLM-backed VL embedder with multimodal (image+text) support
- All four embedder classes gain `unload()` and lazy `_ensure_loaded()` (defers GPU allocation to first use)

**Factory / config:**
- `create_local_embedder` gains a `backend` kwarg (`"vllm"` | `"hf"`); ingest defaults to `"vllm"`, recall/query defaults to `"hf"`
- New `local_ingest_backend` config field in `LocalIngestParams` with a field validator
- New `--embed-local-ingest-backend` CLI flag plumbed through the text-embed pipeline
- `gpu_memory_utilization`, `enforce_eager`, and `dimensions` forwarded through factory, processor, and operator

**Tests:**
- New `test_vllm_embed.py` with comprehensive coverage of both the text and VL vLLM embedders (guarded with `pytest.importorskip("vllm")`)
- Updated `test_create_local_embedder.py` to cover all four embedder classes and the `backend` dispatch logic

## Checklist


nemo_retriever/src/nemo_retriever/model/__init__.py
```diff
@@ -40,14 +42,18 @@ def create_local_embedder(
     *,
     device: str | None = None,
     hf_cache_dir: str | None = None,
-    normalize: bool = True,
```
jperez999, line 63: should we keep these values as well?

nemo_retriever/src/nemo_retriever/text_embed/vllm.py
```diff
@@ -0,0 +1,224 @@
...
+            "Install with: uv pip install -e '.[embed-vllm]' or pip install -e '.[embed-vllm]'"
+        ) from e
+
+    if not enforce_eager:
+        cache_dir = compile_cache_dir if compile_cache_dir is not None else _default_compile_cache_dir()
+        if cache_dir:
+            os.makedirs(cache_dir, mode=0o700, exist_ok=True)
+            os.environ["TORCHINDUCTOR_CACHE_DIR"] = cache_dir
+            os.environ["TRITON_CACHE_DIR"] = cache_dir
+            logger.debug("vLLM: using compile cache dir %s", cache_dir)
```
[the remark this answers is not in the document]
jperez999: Is this a real problem for us, like changing values if we spawn actors on other machines?

nemo_retriever/src/nemo_retriever/text_embed/vllm.py
```diff
@@ -0,0 +1,233 @@
...
+        If False (default), use torch.compile/CUDAGraphs for speed. If True, disable for stability.
+    compile_cache_dir : str, optional
+        Directory for PyTorch inductor/Triton JIT cache.
+
+    Returns
+    -------
+    list of list of float
+        One embedding vector per prompt, in order.
+    """
+    llm = create_vllm_llm(
```
jperez999: Is this always creating a new version? Cant we do a thing where we give the Operator a reference to the model and use that everytime? Would that free up the startup times here?

jperez999 (changes requested):

nemo_retriever/src/nemo_retriever/text_embed/vllm.py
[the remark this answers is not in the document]
charlesbluca: Ended up dropping this mutation in https://github.com/NVIDIA/NeMo-Retriever/pull/1494/commits/fbb433f2099b89e2f5b90307875d42b6c4587dae; had initially added this as a workaround to get around some vLLM startup failures when the default cache directory is `noexec` mounted, but in hindsight more reasonable workaround here was just necessitating a custom TMPDIR

nemo_retriever/src/nemo_retriever/text_embed/vllm.py
charlesbluca: Good point - took a closer look and this `embed_via_vllm` function isn't actually getting used anymore, dropped in https://github.com/NVIDIA/NeMo-Retriever/pull/1494/commits/572b8b7bcb2da2c6d53bf52ce5a4ece73aad14fa

In practice, embedding is happening through an embedder instance, which is where this LLM object is stored for reuse:

https://github.com/charlesbluca/NeMo-Retriever/blob/fc5b10b2a9d10b7122f9080a39ac14fa9b96e32a/nemo_retriever/src/nemo_retriever/model/local/llama_nemotron_embed_1b_v2_embedder.py#L51-L58

nemo_retriever/src/nemo_retriever/model/__init__.py
charlesbluca: Yeah, readded these back with https://github.com/NVIDIA/NeMo-Retriever/pull/1494/commits/8c6ec7fdd16aacc9e8830b31567937b697c6d246

jperez999 (approved):

jperez999:
Might need to coordinate with @edknv and the VL reranker PR

jperez999 (approved):

PR #1510, opened 2026-03-07
fix lancedb out for recall after pipeline

jperez999:
## Description
works on fixing the lancedb search and recall steps in the batch example

## Checklist


nemo_retriever/src/nemo_retriever/examples/batch_pipeline.py
```diff
@@ -204,11 +201,56 @@ def _ensure_lancedb_table(uri: str, table_name: str) -> None:
 
     import pyarrow as pa  # type: ignore
 
-    schema = lancedb_schema(2048)
-    empty = pa.table({f.name: [] for f in schema}, schema=schema)
+    schema = pa.schema(
```
jdye64, line 205: Is something different about this than what is in the new central `lancedb_schema` function?

jperez999: we have a ton of places where we reference lancedb. We need to consolidate. I will change for that call to be used in all functions.

jdye64: yes, agreed. Some of the functions that had been centralized seem to have been re-introduced. Lets make some time to tidy this up soon.

PR #1521, opened 2026-03-09
(retriever) add pdfium_hybrid/ocr extract methods for scanned pages

edknv:
## Description
This PR adds `pdfium_hybrid` and `ocr` text extraction methods to the nemo_retriever library, matching the nv-ingest pipeline's existing support.

- `--method pdfium` (default): no behavioral change, native text extraction only
- `--method pdfium_hybrid`: scanned pages get OCR'd text, non-scanned pages keep native text
- `--method ocr`: all pages get OCR'd text

## Checklist

### DIFF
--- a/nemo_retriever/src/nemo_retriever/examples/batch_pipeline.py
+++ b/nemo_retriever/src/nemo_retriever/examples/batch_pipeline.py
@@ -323,7 +323,7 @@ def main(
         help="LanceDB URI/path for this run.",
     ),
     method: str = typer.Option(
-        "pdfium_hybrid",
+        "pdfium",
         "--method",
         help="PDF text extraction method: 'pdfium' (native only), 'pdfium_hybrid' (native + OCR for scanned), or 'ocr' (OCR all pages).",  # noqa: E501
     ),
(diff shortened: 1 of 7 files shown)


nemo_retriever/src/nemo_retriever/examples/batch_pipeline.py
```diff
@@ -305,6 +305,11 @@ def main(
         "--lancedb-uri",
         help="LanceDB URI/path for this run.",
     ),
+    method: str = typer.Option(
+        "pdfium",
```
jdye64: should pdfium or pdfium_hybrid be the default? Asking because I really don't know?

nemo_retriever/src/nemo_retriever/examples/inprocess_pipeline.py
```diff
@@ -92,6 +92,11 @@ def main(
         "--embed-model-name",
         help="Embedding model name passed to .embed().",
     ),
+    method: str = typer.Option(
+        "pdfium",
```
jdye64: same question about default

nemo_retriever/src/nemo_retriever/examples/batch_pipeline.py
edknv: I changed the default to `pdfium_hybrid` in 4c7e8379ff7ca1c54aba87468bf65277f3906fcd.

It behaves identically to pdfium for text-based pages but for scanned pages it automatically switches to OCR instead of returning empty strings. There's no downside for non-scanned PDFs since the OCR path only activates when the page has zero native text.

This should gives us a 1% recall boost on bo767.

PR #1524, opened 2026-03-10
(retriever) add image file ingestion (PNG, JPEG, BMP, TIFF, SVG)

edknv:
## Checklist

### DIFF
--- a/nemo_retriever/pyproject.toml
+++ b/nemo_retriever/pyproject.toml
@@ -74,6 +74,9 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
+svg = [
+  "cairosvg>=2.7.0",
+]
 dev = [
   "build>=1.2.2",
   "pytest>=8.0.2",


nemo_retriever/src/nemo_retriever/examples/batch_pipeline.py
```diff
@@ -521,9 +538,23 @@ def main(
...
-            exts = ext_map.get(input_type, ["*.pdf"])
-            file_patterns = [str(input_path / e) for e in exts]
+            # If a specific image extension was requested, use only that extension's globs.
+            if _original_input_type in _image_ext_map:
+                exts = _image_ext_map[_original_input_type]
+            else:
+                exts = ext_map.get(input_type, ["*.pdf"])
+            import glob as _glob
+
+            all_candidates = [str(input_path / e) for e in exts]
```
jdye64, line 568: I think using local is fine for now but something we should consider later is Ray data allows users to specify object stores like S3 and this logic would break in that situation

nemo_retriever/src/nemo_retriever/image/load.py
```diff
@@ -0,0 +1,160 @@
...
+from io import BytesIO
+from typing import Any, Dict
+
+import pandas as pd
+from PIL import Image
+
+# Raster formats handled natively by Pillow.
+_RASTER_EXTENSIONS = {".png", ".jpg", ".jpeg", ".bmp", ".tiff", ".tif"}
+
+# SVG requires cairosvg (optional dependency).
```
jdye64, line 28: should we add this in the pyproject.toml as an optional dependency then?

nemo_retriever/src/nemo_retriever/image/load.py
```diff
@@ -0,0 +1,160 @@
...
+        ]
+    )
+
+
+def _svg_to_pil(content_bytes: bytes) -> Image.Image:
+    """Convert SVG bytes to a PIL RGB Image via cairosvg."""
+    try:
+        import cairosvg  # type: ignore[import-untyped]
+    except ImportError:
+        raise ImportError("cairosvg is required for SVG image support. " "Install it with: pip install cairosvg")
```
jdye64, line 68: Maybe this is good enough for now

nemo_retriever/src/nemo_retriever/ingest_modes/fused.py
```diff
@@ -166,6 +166,10 @@ class FusedIngestor(BatchIngestor):
     RUN_MODE = "fused"
     _fused_extract_flags: Dict[str, Any] = {}
 
+    def extract_image_files(self, params: ExtractParams | None = None, **kwargs: Any) -> "FusedIngestor":
+        """Fused mode does not yet support extract_image_files."""
+        raise NotImplementedError("Fused mode does not yet support extract_image_files")
```
jdye64, line 171: yeah probably gonna nuke fused mode for now anyway ...

nemo_retriever/src/nemo_retriever/image/load.py
edknv: I think it's a good idea. Added to pyproject.toml as an optional dep.

PR #1533, opened 2026-03-10
Change default OCR_MODEL_NAME to 'pipeline' instead of 'scene_text_ensemble'

jdye64:
## Checklist


jdye64:
@charlesbluca do I understand correctly that we need to wait until the new ocr model is released to actually make this switch or will it also work with the current version? Basically is this safe to merge or do we need to wait?

PR #1534, opened 2026-03-10
Pass NVIDIA_API_KEY from env to remote hosted NIMs

ChrisJar:
## Checklist

### DIFF
--- a/docker-compose.a10g.yaml
+++ b/docker-compose.a10g.yaml
@@ -6,14 +6,17 @@ services:
   page-elements:
     environment:
       - NIM_TRITON_MAX_BATCH_SIZE=1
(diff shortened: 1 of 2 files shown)


nemo_retriever/src/nemo_retriever/ingest_modes/batch.py
```diff
@@ -82,6 +82,13 @@ def _runtime_env_vars() -> dict[str, str]:
     return {key: value for key, value in env_vars.items() if isinstance(value, str)}
 
 
+def _resolve_remote_api_key(explicit_api_key: Optional[str] = None) -> Optional[str]:
```
jdye64: One more nit. Since this function is defined here and in `batch_pipeline.py` can we move it to a common file in `utils`

PR #1539, opened 2026-03-10
harness: add inherited presets and auto worker sizing

jioffe502:
## Description

- add preset inheritance to reduce duplicated harness tuning
- limit auto mode to worker counts and document the user story
- wire harness tuning into batch requested plan refreshes

## Checklist


nemo_retriever/src/nemo_retriever/ingest_modes/batch.py
```diff
@@ -82,6 +82,79 @@ def _runtime_env_vars() -> dict[str, str]:
     return {key: value for key, value in env_vars.items() if isinstance(value, str)}
 
 
+def _normalize_requested_plan_int(value: Any) -> int | None:
```
jdye64, line 86: What purpose do these added functions in `batch.py` serve? Can we move them elsewhere?

nemo_retriever/src/nemo_retriever/ingest_modes/batch.py
```diff
@@ -253,6 +327,18 @@ def __init__(
         self._extract_html_kwargs: Dict[str, Any] = {}  # noqa: F821
         self._use_nemotron_parse_only: bool = False
 
+    def _refresh_requested_plan(self, batch_tuning: dict[str, Any]) -> None:
```
jdye64, line 331: I think most of this can already be accomplished with the existing functions in [ray_resource_hueristics.py](https://github.com/NVIDIA/NeMo-Retriever/blob/main/nemo_retriever/src/nemo_retriever/utils/ray_resource_hueristics.py) but if they are needed for some reason lets move them there and name them appropriately not not clutter the core logic

jdye64 (changes requested):

PR #1544, opened 2026-03-10
(retriever) add markdown render helpers

jioffe502:
### Description

Adds rewrite-native helpers for rendering a single retriever extract result as page-grouped markdown. This keeps the feature inside `nemo_retriever` instead of depending on the legacy client helpers.

- add `to_markdown()` and `to_markdown_by_page()` under `nemo_retriever.io`
- support saved result wrappers and cover the new behavior with focused tests
- document the new `results[0]` markdown flow in the retriever README

## Checklist


nemo_retriever/src/nemo_retriever/io/markdown.py
```diff
@@ -0,0 +1,284 @@
...
+        rendered[page_number] = header + ("\n\n" + "\n\n".join(blocks) if blocks else "\n")
+
+    return rendered
+
+
+def to_markdown(results: object) -> str:
+    """Render a single document result as one markdown document."""
+    pages = to_markdown_by_page(results)
+    if not pages:
+        return f"# {_DOCUMENT_TITLE}\n\n_No content found._"
```
jdye64, line 66: Shouldn't we just return `None` here and handle that appropriately upstream? Seems this could throw off embeddings

nemo_retriever/src/nemo_retriever/io/markdown.py
```diff
@@ -0,0 +1,284 @@
...
+            return [dict(item) for item in value]
+    if _looks_like_record(results):
+        return [dict(results)]
+    raise ValueError("Markdown rendering expects a document row, row list, or saved results payload.")
+
+
+def _looks_like_record(record: Mapping[str, Any]) -> bool:
+    return any(
+        key in record
+        for key in (
```
jdye64, line 117: Lets add a TODO here that we should centralize this later. I see cases where an Actor might change its output format and this break. I want to do this in the Actors in general as well

jdye64: The TODOs just help us remember that for later

PR #1553, opened 2026-03-10
(retriever) follow up markdown review

jioffe502:
## Description

- return None from to_markdown when no content is present
- add TODO to centralize result-shape detection heuristics

## Checklist

### DIFF
--- a/nemo_retriever/src/nemo_retriever/io/markdown.py
+++ b/nemo_retriever/src/nemo_retriever/io/markdown.py
@@ -59,11 +59,11 @@ def to_markdown_by_page(results: object) -> dict[int, str]:
     return rendered
 
 
-def to_markdown(results: object) -> str:
+def to_markdown(results: object) -> str | None:
     """Render a single document result as one markdown document."""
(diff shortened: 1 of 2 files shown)


nemo_retriever/src/nemo_retriever/io/markdown.py
```diff
@@ -110,6 +110,9 @@ def _records_from_mapping(results: Mapping[str, Any]) -> list[dict[str, Any]]:
         return [dict(results)]
     raise ValueError("Markdown rendering expects a document row, row list, or saved results payload.")
 
+    # TODO(jioffe): Centralize retriever result-shape detection so helpers and
```
jperez999, line 113: Do we need this comment here?

jioffe502: @jdye64 is this the comment/todo you were looking for? 

"Let's add a TODO here that we should centralize this later. I see cases where an Actor might change its output format and this break. I want to do this in the Actors in general as well"

PR #1565, opened 2026-03-11
Add rerank

jperez999:
## Description
Adds reranker to be used in batch example. Also fixed recall to work with the retriever object to get queries.

## Checklist

### DIFF
--- a/.github/workflows/perform-release.yml
+++ b/.github/workflows/perform-release.yml
@@ -19,11 +19,16 @@ on:
         required: false
         type: boolean
         default: false
+      skip-docker:
+        description: '⛔ SKIP Docker — disable Docker build and publish entirely'
+        required: false
+        type: boolean
+        default: false
       skip-helm-chart:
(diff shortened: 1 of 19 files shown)


nemo_retriever/src/nemo_retriever/retriever.py
```diff
@@ -27,6 +52,20 @@ class Retriever:
     local_hf_device: Optional[str] = None
     local_hf_cache_dir: Optional[Path] = None
     local_hf_batch_size: int = 64
+    # Reranking -----------------------------------------------------------
+    reranker: Optional[str] = "nvidia/llama-nemotron-rerank-1b-v2"
```
jioffe502, line 56: The default here is the model name, which means any code that creates a Retriever() without explicitly setting reranker will download and load the 1B model. The docstring on line 58 says "Set to None to skip reranking (default)"

should this default to None instead so reranking is opt-in?

The batch_pipeline.py handles this correctly with the --reranker bool flag

 this is just about the dataclass default for direct Retriever() callers.

PR #1594, opened 2026-03-12
(retriever) Install as part of docker build

charlesbluca:
## Description

Refactors Dockerfile logic so that nemo-retriever is installed for use as part of build

## Checklist

### DIFF
--- a/.github/workflows/reusable-pypi-build.yml
+++ b/.github/workflows/reusable-pypi-build.yml
@@ -114,7 +114,7 @@ jobs:
           PY
           RETRIEVER_RELEASE_TYPE=${{ inputs.release-type }} \
           RETRIEVER_VERSION=${{ steps.set-version.outputs.version }} \
-          RETRIEVER_BUILD_NUMBER=${{ github.run_number }} \
+          RETRIEVER_BUILD_NUMBER=${{ inputs.release-type == 'release' && '0' || github.run_number }} \
           RETRIEVER_GIT_SHA=${{ github.sha }} \
           python -m build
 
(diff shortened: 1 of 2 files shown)


nemo_retriever/Dockerfile
```diff
@@ -99,17 +104,21 @@ WORKDIR /workspace
...
-
-# Use base stage's venv at /opt/retriever_runtime; install nemo_retriever in editable mode (path deps: ../src, ../api, ../client)
-# SHELL ["/bin/bash", "-c"]
-# RUN --mount=type=cache,target=/root/.cache/pip \
-#     --mount=type=cache,target=/root/.cache/uv \
-#     . /opt/retriever_runtime/bin/activate \
-#     && uv pip install -e ./nemo_retriever
+# Activate venv by default so CLI and python see nemo_retriever; mount over /workspace for dev.
+ENV VIRTUAL_ENV=/opt/retriever_runtime
+ENV PATH=/opt/retriever_runtime/bin:/root/.local/bin:$PATH
```
jperez999, line 109: This will break because $PATH does not exist, right?

charlesbluca: Think there's already a pre-populated PATH at this point, so this just ends up getting pre-pended to it:

```
docker run --rm -it nemo-retriever:latest bash -c "echo \$PATH"
/opt/retriever_runtime/bin:/root/.local/bin:/root/.local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```

In practice, think even if PATH was unset this would just expand to something like `PATH=/opt/retriever_runtime/bin:/root/.local/bin:`, so the desired binaries would still be exposed

jperez999: nvm the actual issue was the ld_library_path that you removed.

PR #1598, opened 2026-03-12
(retriever) Improve batched audio extraction performance

charlesbluca:
## Description

Some tweaks here to improve batched audio extraction perf:
- Plumbing `num_gpus` to ASR actor so local Parakeet runs with GPUs
- Refactoring actor ➡️ model batching so an actor's entire batch is passed through to the model

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -39,13 +39,13 @@ RUN chmod +x scripts/install_ffmpeg.sh \
 # For GPL-licensed components, we provide their source code in the container
 # via `apt-get source` below to satisfy GPL requirements.
 ARG GPL_LIBS="\
+    libfreetype6 \
     libltdl7 \
     libhunspell-1.7-0 \
     libhyphen0 \
     libdbus-1-3 \
 "
 ARG FORCE_REMOVE_PKGS="\
-    libfreetype6 \
     ucf \
     liblangtag-common \
     libjbig0 \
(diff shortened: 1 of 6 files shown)


nemo_retriever/src/nemo_retriever/audio/asr_actor.py
```diff
@@ -158,6 +164,91 @@ def __call__(self, batch_df: pd.DataFrame) -> pd.DataFrame:
...
+                    f = tempfile.NamedTemporaryFile(suffix=".audio", delete=False)
+                    f.write(raw)
+                    f.close()
+                    path_to_use = f.name
+                    temp_created = f.name
+                except Exception as e:
+                    logger.warning("Failed to write temp file for ASR: %s", e)
+                    path_to_use = ""
+            else:
+                if path:
```
jperez999, line 195: what is happening in this code here, above you are writing to a file but here you are reading from a possible file then writing to it again? What is this logic meant to catch that the above logic does not handle?

nemo_retriever/src/nemo_retriever/model/local/parakeet_ctc_1_1b_asr.py
```diff
@@ -137,44 +137,70 @@ def _ensure_loaded(self) -> None:
...
+            if audio is not None and audio.size > 0:
+                valid.append(audio)
+                indices.append(i)
+        if not valid:
+            return [""] * len(audios)
+        try:
+            transcripts = self._transcribe_audio_batch(valid)
+        except Exception as e:
+            logger.warning("ASR (transformers) batch failed: %s", e)
+            transcripts = [""] * len(valid)
```
jperez999, line 174: This will fail the entire batch if we get an exception from one audio right?

PR #1654, opened 2026-03-18
Add punctuation based audio segmenting support to nemo_retriever

ChrisJar:
## Checklist


nemo_retriever/src/nemo_retriever/audio/asr_actor.py
```diff
@@ -122,7 +125,9 @@ class ASRActor:
 
     When audio_endpoints are set, uses Parakeet (Riva ASR) via gRPC. When both are
     null/empty, uses local HuggingFace/NeMo Parakeet (nvidia/parakeet-ctc-1.1b).
-    Output rows have path, text, page_number, metadata for downstream embed.
+    Output rows have path, text, page_number, metadata for downstream embed. When
+    ``params.segment_audio`` is enabled for remote Parakeet, punctuation-delimited
```
charlesbluca, line 130: Mind adding mention around here that segmentation is limited to remote Parakeet for now? Might be nice to revisit the context (IIRC, segmentation is done in a follow-up model that we don't have access to) for later if we come around to enabling local model inference

PR #1668, opened 2026-03-20
feat: add allow_no_gpu option to skip GPU requirement during ingest

tomer-levin-nv:
Adds an allow_no_gpu flag (default False) to IngestorCreateParams, BatchIngestor, and resolve_requested_plan. When True, skips the "No GPUs available" error and returns a CPU-only RequestedPlan with all GPU allocations set to 0.0.

Made-with: Cursor

## Checklist

### DIFF
--- a/nemo_retriever/tests/test_batch_ingestor.py
+++ b/nemo_retriever/tests/test_batch_ingestor.py
@@ -72,7 +72,7 @@ def test_batch_ingestor_filters_none_runtime_env_vars(monkeypatch) -> None:
     )
     monkeypatch.setattr(
         "nemo_retriever.ingest_modes.batch.resolve_requested_plan",
-        lambda cluster_resources: {"plan": "dummy"},
+        lambda cluster_resources, allow_no_gpu=False: {"plan": "dummy"},
     )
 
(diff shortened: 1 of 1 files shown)


nemo_retriever/src/nemo_retriever/utils/ray_resource_hueristics.py
```diff
@@ -411,11 +411,38 @@ def resolve_requested_plan(
...
     override_pdf_extract_tasks: Optional[int] = None,
+    allow_no_gpu: bool = False,
 ) -> RequestedPlan:
     available_gpu_count = max(0, int(cluster_resources.available_gpu_count()))
 
     if available_gpu_count == 0:
-        raise ValueError("No GPUs available")
+        if not allow_no_gpu:
+            raise ValueError("No GPUs available")
+        return RequestedPlan(
```
jperez999: by setting these defaults do we hurt the user if they have resources to spawn more actors on a CPU machine? I could see having resources to spawn multiple page_elements actors or embedding model actors to handle more load faster (hitting an endpoint that can handle the load)?

PR #1699, opened 2026-03-23
nemo_retriever: Add structured harness metrics reports

jioffe502:
Switch the retriever harness from stdout scraping to structured run reports shared across supported run modes. Follow-on commits also harden artifact metadata handling so these reports stay reliable in real container/worktree runs.

- add shared batch, inprocess, and fused run-report plumbing and wire the harness to consume it
- persist runtime and detection metrics as structured artifacts instead of parsing console output
- keep `latest_commit` populated in harness artifacts by falling back to `.git/HEAD` refs when `git rev-parse` fails in container/worktree setups
- preserve detection counters through LanceDB round-trips by writing metadata as JSON and accepting legacy metadata literals during detection-summary reads
- validate behavior with real bo20/jp20 harness runs as integration coverage


nemo_retriever/src/nemo_retriever/application/modes/run_fused.py
```diff
@@ -4,10 +4,156 @@
...
+    persist_detection_summary_artifact,
+    print_evaluation_metrics,
+    resolve_lancedb_target,
+    resolve_mode_file_patterns,
+    resolve_input_pages,
+    shutdown_ray_safely,
+)
+
+
+class FusedPipelineConfig(ModePipelineConfigModel):
```
jperez999: This is almost exactly the same as the other PipelineConfig is there nothing we can reuse from both files, maybe create a generalized version of the config and tweak only where necessary. Also we have no support for fused yet, so don't think you need to add it.

nemo_retriever/src/nemo_retriever/application/modes/run_fused.py
```diff
@@ -4,10 +4,156 @@
...
+        return persist_run_report_artifacts(
+            report,
+            runtime_metrics_dir=cfg.execute_params.runtime_metrics_dir,
+            prefix=cfg.execute_params.runtime_metrics_prefix,
+        )
+    finally:
+        shutdown_ray_safely()
+
+
+def render_fused_run_report(report: RunReport) -> None:
```
jperez999: rendering the report should be the same no matter the pipeline. We should not have different logic in each module particular to an application mode.

nemo_retriever/src/nemo_retriever/application/modes/run_inprocess.py
```diff
@@ -4,10 +4,164 @@
...
+    evaluate_lancedb_metrics,
+    persist_detection_summary_artifact,
+    print_evaluation_metrics,
+    resolve_lancedb_target,
+    resolve_mode_file_patterns,
+    resolve_input_pages,
+)
+
+
+class InProcessPipelineConfig(ModePipelineConfigModel):
```
jperez999, line 45: same as above. dry

nemo_retriever/src/nemo_retriever/application/modes/run_inprocess.py
```diff
@@ -4,10 +4,164 @@
...
+        ingestor = ingestor.files(file_patterns).extract(cfg.extract_params or ExtractParams())
+
+    if cfg.enable_text_chunk:
+        ingestor = ingestor.split(chunk_params)
+
+    ingestor = ingestor.embed(cfg.embed_params).vdb_upload(cfg.vdb_upload_params)
+    return ingestor, file_patterns
+
+
+def run_inprocess_pipeline(cfg: InProcessPipelineConfig) -> RunReport:
```
jperez999, line 86: Shouldnt the only difference between this run pipeline and batch is the run_mode string? If it is, then this shouldnt exist as code it should just be a switch to one generalized function.

nemo_retriever/src/nemo_retriever/application/modes/shared.py
```diff
@@ -0,0 +1,336 @@
...
+logger = logging.getLogger(__name__)
+
+
+class ModePipelineConfigModel(BaseModel):
+    """Strict config base for mode pipeline runners."""
+
+    model_config = ConfigDict(extra="forbid")
+
+
+class _TeeStream:
```
jperez999: Why do we need stream controls, I thought this was removing the need for that? We dont care what the display onscreen looks like (we leave it to lower level libraries, like ray), we just need the datapoints from the run, and we no longer get those from std out/err, right?

nemo_retriever/src/nemo_retriever/application/modes/shared.py
```diff
@@ -0,0 +1,336 @@
...
+            handle.flush()
+        finally:
+            handle.close()
+
+
+def lancedb_module() -> Any:
+    return import_module("lancedb")
+
+
+def ensure_lancedb_table(uri: str, table_name: str) -> None:
```
jperez999: Dont we have this in other parts of the code base already? If its not a duplicate, why move it here. This is lance stuff nothing to do with run modes. I should not have to import the lance package in this module.

nemo_retriever/src/nemo_retriever/harness/artifacts.py
```diff
@@ -18,21 +19,104 @@ def now_timestr() -> str:
     return datetime.now(timezone.utc).strftime("%Y%m%d_%H%M%S_UTC")
 
 
+_COMMIT_RE = re.compile(r"^[0-9a-fA-F]{7,40}$")
```
jperez999, line 22: Why are we changing commit string conversion logic in this PR? Did you do other stuff in the PR as well?

nemo_retriever/src/nemo_retriever/harness/config.py
```diff
@@ -47,13 +48,37 @@
     "gpu_ocr",
     "gpu_embed",
 }
+INPROCESS_TUNING_FIELDS = {
+    "max_workers",
+    "gpu_devices",
+}
+FUSED_TUNING_FIELDS = {
```
jperez999: we dont have support for fused , so why add the fused entries? Also why are some of the kwargs prefixed with fused... is it supposed to mean something different?

nemo_retriever/src/nemo_retriever/application/modes/shared.py
jioffe502: Removed stream tee/logging controls from application/modes/shared.py to keep this path focused on structured datapoints and mode orchestration

nemo_retriever/src/nemo_retriever/application/modes/shared.py
jioffe502: Moved LanceDB lifecycle/read helpers out of application/modes/shared.py into vector_store/lancedb_store.py, and updated mode/detection consumers to import from there

nemo_retriever/src/nemo_retriever/harness/artifacts.py
jioffe502: Kept commit fallback hardening because it stabilizes latest_commit in container/worktree runs, and added focused tests for gitdir ref-file and packed-refs fallback to make scope explicit

it came from running inside of a dev container

nemo_retriever/src/nemo_retriever/harness/config.py
jioffe502: Removed fused harness scope from this PR. harness/config.py and harness/run.py now only support batch/inprocess for this metrics work, and fused-specific harness test coverage was removed

 run_fused is back to wrapper-only

nemo_retriever/src/nemo_retriever/application/modes/run_fused.py
jioffe502: Consolidated rendering into shared report helpers in application/modes/reports.py and switched batch/inprocess to use the same render path.

jioffe502:
closed for #1803

PR #1703, opened 2026-03-24
Add graph 2

jperez999:
## Description
adds graph dag creation support

## Checklist

### DIFF
--- a/nemo_retriever/src/nemo_retriever/examples/batch_pipeline.py
+++ b/nemo_retriever/src/nemo_retriever/examples/batch_pipeline.py
@@ -21,18 +21,18 @@ from typing import Optional, TextIO
 import typer
 
 from nemo_retriever import create_ingestor
-from nemo_retriever.ingest_modes.batch import BatchIngestor
-from nemo_retriever.ingest_modes.lancedb_utils import lancedb_schema
-from nemo_retriever.model import resolve_embed_model
 from nemo_retriever.params import CaptionParams
 from nemo_retriever.params import DedupParams
 from nemo_retriever.params import EmbedParams
 from nemo_retriever.params import ExtractParams
-from nemo_retriever.params import IngestExecuteParams
-from nemo_retriever.params import IngestorCreateParams
 from nemo_retriever.params import TextChunkParams
-from nemo_retriever.recall.beir import BeirConfig, evaluate_lancedb_beir
-from nemo_retriever.recall.core import RecallConfig, retrieve_and_score
+from nemo_retriever.params.models import BatchTuningParams
+from nemo_retriever.recall.beir import evaluate_lancedb_beir
+from nemo_retriever.utils.detection_summary import (
+    collect_detection_summary_from_df,
+    print_run_summary,
+    write_detection_summary,
+)
 from nemo_retriever.utils.remote_auth import resolve_remote_api_key
 from nemo_retriever.vector_store.lancedb_store import handle_lancedb
 
@@ -385,7 +385,6 @@ def main(
             "Uses --text-chunk-max-tokens and --text-chunk-overlap-tokens (defaults: 1024, 150)."
         ),
     ),
-    text_chunk: bool = typer.Option(False, "--text-chunk", help="Split extracted text into chunks before embedding."),
     text_chunk_max_tokens: Optional[int] = typer.Option(
         None, "--text-chunk-max-tokens", help="Maximum tokens per text chunk."
     ),
@@ -524,6 +523,10 @@ def main(
             "inference_batch_size": embed_batch_size or None,
         }
         embed_params = EmbedParams(**{key: value for key, value in embed_params_kwargs.items() if value is not None})
+        text_chunk_params = TextChunkParams(
+            max_tokens=text_chunk_max_tokens or 1024,
+            overlap_tokens=text_chunk_overlap_tokens if text_chunk_overlap_tokens is not None else 150,
+        )
         ingestor = create_ingestor(
             run_mode="batch",
             ray_address=ray_address,
@@ -532,21 +535,19 @@ def main(
         )
 
         if input_type == "txt":
-            ingestor = ingestor.files(file_patterns).extract_txt(_text_chunk_params)
+            ingestor = ingestor.files(file_patterns).extract_txt(text_chunk_params)
         elif input_type == "html":
-            ingestor = ingestor.files(file_patterns).extract_html(_text_chunk_params)
+            ingestor = ingestor.files(file_patterns).extract_html(text_chunk_params)
         elif input_type == "image":
-            ingestor = ingestor.files(file_patterns).extract_image_files(_extract_params(_detection_batch_tuning))
+            ingestor = ingestor.files(file_patterns).extract_image_files(extract_params)
         elif input_type == "doc":
-            ingestor = ingestor.files(file_patterns).extract(_extract_params(_pdf_batch_tuning))
+            ingestor = ingestor.files(file_patterns).extract(extract_params)
         else:
-            ingestor = ingestor.files(file_patterns).extract(
-                _extract_params(_pdf_batch_tuning, inference_batch_size=page_elements_batch_size)
-            )
+            ingestor = ingestor.files(file_patterns).extract(extract_params)
 
         enable_text_chunk = text_chunk or text_chunk_max_tokens is not None or text_chunk_overlap_tokens is not None
         if enable_text_chunk:
-            ingestor = ingestor.split(_text_chunk_params)
+            ingestor = ingestor.split(text_chunk_params)
 
         enable_caption = caption or caption_invoke_url is not None
         enable_dedup = dedup if dedup is not None else enable_caption
@@ -565,13 +566,6 @@ def main(
                     device=caption_device,
                     context_text_max_chars=caption_context_text_max_chars,
                     gpu_memory_utilization=caption_gpu_memory_utilization,
-                )
-            )
-        if text_chunk or text_chunk_max_tokens is not None or text_chunk_overlap_tokens is not None:
-            ingestor = ingestor.split(
-                TextChunkParams(
-                    max_tokens=text_chunk_max_tokens or 1024,
-                    overlap_tokens=text_chunk_overlap_tokens if text_chunk_overlap_tokens is not None else 150,
                 )
             )
         ingestor = ingestor.embed(embed_params)
@@ -620,13 +614,6 @@ def main(
         handle_lancedb(ingest_local_results, lancedb_uri, LANCEDB_TABLE, hybrid=hybrid, mode="overwrite")
         lancedb_write_time = time.perf_counter() - lancedb_write_start
 
-        db = _lancedb().connect(lancedb_uri)
-        table = db.open_table(LANCEDB_TABLE)
-        if int(table.count_rows()) == 0:
-            logger.warning("LanceDB table is empty; skipping %s evaluation.", evaluation_mode)
-            ray.shutdown()
-            return
-
         from nemo_retriever.model import resolve_embed_model
         from nemo_retriever.recall.beir import BeirConfig
         from nemo_retriever.recall.core import RecallConfig, retrieve_and_score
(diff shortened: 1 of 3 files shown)


nemo_retriever/src/nemo_retriever/audio/asr_actor.py
```diff
@@ -120,7 +122,7 @@ def _get_client(params: ASRParams):  # noqa: ANN201
     )
 
 
-class ASRActor:
+class ASRActor(AbstractOperator, CPUOperator):
```
jdye64, line 125: What makes the ASRActor a `CPUActor` instead of `GPUActor`? does it use a NIM or something?

nemo_retriever/src/nemo_retriever/audio/chunk_actor.py
```diff
@@ -28,7 +30,7 @@
 CHUNK_COLUMNS = ["path", "source_path", "duration", "chunk_index", "metadata", "page_number", "bytes"]
 
 
-class MediaChunkActor:
+class MediaChunkActor(AbstractOperator, CPUOperator):
```
jdye64: Same question about `CPUOperator` vs `GPUOperator`

nemo_retriever/src/nemo_retriever/vector_store/lancedb_utils.py
```diff
@@ -0,0 +1,215 @@
...
+
+import json
+from pathlib import Path
+from typing import Any, Dict, List, Optional, Tuple
+
+
+def extract_embedding_from_row(
+    row: Any,
+    *,
+    embedding_column: str = "text_embeddings_1b_v2",
```
jdye64, line 17: note for me to remember to look for this being overridden if VL model is used ...

nemo_retriever/src/nemo_retriever/vector_store/lancedb_utils.py
```diff
@@ -0,0 +1,215 @@
...
+
+from __future__ import annotations
+
+import json
+from pathlib import Path
+from typing import Any, Dict, List, Optional, Tuple
+
+
+def extract_embedding_from_row(
+    row: Any,
```
jdye64, line 15: Can we explicitly type this if a pd.Series, Dict, if that is what it is? That helps assist AI agents

nemo_retriever/src/nemo_retriever/vector_store/lancedb_utils.py
```diff
@@ -0,0 +1,215 @@
...
+
+    payload = getattr(row, embedding_column, None)
+    if isinstance(payload, dict):
+        emb = payload.get(embedding_key)
+        if isinstance(emb, list) and emb:
+            return emb  # type: ignore[return-value]
+    return None
+
+
+def extract_source_path_and_page(row: Any) -> Tuple[str, int]:
```
jdye64, line 35: Same ask for typing `row`

nemo_retriever/src/nemo_retriever/vector_store/lancedb_utils.py
```diff
@@ -0,0 +1,215 @@
...
+        path = value.strip()
+
+    value = getattr(row, "page_number", None)
+    try:
+        if value is not None:
+            page = int(value)
+    except Exception:
+        pass
+
+    meta = getattr(row, "metadata", None)
```
jdye64, line 51: General comment that we can do later but most of this logic is checking types and getting magic value strings. Probably worth us creating a Pydantic object eventually to prevent any missteps if that structure changes and just accessing the attributes from that Pydantic object.

nemo_retriever/src/nemo_retriever/vector_store/lancedb_utils.py
```diff
@@ -0,0 +1,215 @@
...
+            if isinstance(hierarchy, dict) and "page" in hierarchy:
+                try:
+                    page = int(hierarchy.get("page"))
+                except Exception:
+                    pass
+
+    return path, page
+
+
+def _build_detection_metadata(row: Any) -> Dict[str, Any]:
```
jdye64, line 68: `row` typing

nemo_retriever/src/nemo_retriever/vector_store/lancedb_utils.py
```diff
@@ -0,0 +1,215 @@
...
+        "vector": embedding,
+        "pdf_page": pdf_page,
+        "filename": filename,
+        "pdf_basename": pdf_basename,
+        "page_number": int(page_number) if page_number is not None else -1,
+        "source_id": str(source_id),
+        "path": str(path),
+        "metadata": json.dumps(metadata_obj, ensure_ascii=False),
+        "source": json.dumps(source_obj, ensure_ascii=False),
+    }
```
jdye64, line 136: Could we define this schema globally somewhere so other portions of the code can import/analyze it and be aware of the resulting schema as to not be caught off guard? Just a global in this file of `lancedb_store.py` maybe??

I see `lancedb_schema()` further down. Could we use the names from that function and doc it as the "golden copy". That would work too. Say that after ^^ remark.

nemo_retriever/src/nemo_retriever/vector_store/lancedb_utils.py
```diff
@@ -0,0 +1,215 @@
...
+        text = getattr(row, text_column, None)
+        row_out["text"] = str(text) if isinstance(text, str) else ""
+    else:
+        row_out["text"] = ""
+
+    return row_out
+
+
+def build_lancedb_rows(
+    df: Any,
```
jdye64, line 148: Is this always guaranteed to be a pd.Dataframe? If so can we type it that way

nemo_retriever/src/nemo_retriever/vector_store/lancedb_store.py
```diff
@@ -12,7 +12,7 @@
 from datetime import timedelta
 
 from nv_ingest_client.util.vdb.lancedb import LanceDB
-from nemo_retriever.ingest_modes.lancedb_utils import lancedb_schema
+from nemo_retriever.vector_store.lancedb_utils import lancedb_schema
```
jdye64, line 15: nice. I have been wanting to move this. Didn't make sense in `ingest_modes`

nemo_retriever/src/nemo_retriever/utils/error_rows.py
jdye64, line 1: Good also hated this logic being in the `batch_pipeline.py` file previously

nemo_retriever/src/nemo_retriever/text_embed/runtime.py
```diff
@@ -0,0 +1,189 @@
...
+"""Embedding runtime helpers shared by graph-based example pipelines."""
+
+from __future__ import annotations
+
+from typing import Any, List, Optional, Sequence
+
+import pandas as pd
+
+from nemo_retriever.model import resolve_embed_model
+from nemo_retriever.params.models import IMAGE_MODALITIES
```
jdye64, line 14: LOVE me some common imports!

nemo_retriever/src/nemo_retriever/text_embed/runtime.py
```diff
@@ -0,0 +1,189 @@
...
+                vectors = model.embed(batch, batch_size=int(inference_batch_size))
+                tolist = getattr(vectors, "tolist", None)
+                if callable(tolist):
+                    return tolist()
+                return vectors  # type: ignore[return-value]
+
+    default_remote_image_batch_size = 4
+    effective_batch_size = inference_batch_size
+    if endpoint is not None and group_modality in IMAGE_MODALITIES:
+        effective_batch_size = min(inference_batch_size, default_remote_image_batch_size)
```
jdye64, line 51: Curious why min() is used here? I'm sure you have a good reason

nemo_retriever/src/nemo_retriever/text_embed/runtime.py
```diff
@@ -0,0 +1,189 @@
...
+
+    endpoint = (embedding_endpoint or embed_invoke_url or "").strip() or None
+    if endpoint is None and model is None:
+        raise ValueError("Either a local model or an embedding_endpoint must be provided.")
+
+    resolved_model_name = resolve_embed_model(model_name)
+
+    has_per_row_modality = "_embed_modality" in batch_df.columns
+    if has_per_row_modality:
+        modalities = batch_df["_embed_modality"].fillna(embed_modality).unique().tolist()
```
jdye64, line 112: Not a huge fan of using explicit/magic strings to access columns since the names could change but since we don't have any other mechanism in place to get them think that is all we can do for now.

nemo_retriever/src/nemo_retriever/text_embed/runtime.py
```diff
@@ -0,0 +1,189 @@
...
+                return int(len(payload.get("embedding") or []))
+            return 0
+
+        out_df[embedding_dim_column] = out_df.apply(dim, axis=1)
+    else:
+        out_df[embedding_dim_column] = [0 for _ in range(len(out_df.index))]
+
+    out_df[has_embedding_column] = [bool(int(dim) > 0) for dim in out_df[embedding_dim_column].tolist()]
+
+    for column in ("_image_b64", "_embed_modality"):
```
jdye64, line 185: Same thing about magic strings for column names. Think we have to live with it for now

nemo_retriever/src/nemo_retriever/ingest_modes/batch.py
```diff
@@ -145,14 +148,49 @@ def __call__(self, batch_df: Any) -> Any:
         return batch_df
 
 
-class _BatchEmbedActor:
+class ExplodeContentActor(AbstractOperator, CPUOperator):
```
jdye64, line 154: I kinda feel like we should move this to another file like we do the other Actors. Seems like it could have some common uses and people might not look in `batch.py` for it

nemo_retriever/src/nemo_retriever/ingest_modes/batch.py
jdye64, line 1: I looked this over but it is so much I'll be honest I think just doing post merge testing here is gonna be the best option. Seems completely fine but too much for my small brain to take in

nemo_retriever/src/nemo_retriever/graph/content_operators.py
```diff
@@ -0,0 +1,43 @@
...
+from __future__ import annotations
+
+from typing import Any
+
+from nemo_retriever.graph.abstract_operator import AbstractOperator
+from nemo_retriever.graph.cpu_operator import CPUOperator
+from nemo_retriever.graph.content_transforms import explode_content_to_rows
+
+
+class ExplodeContentActor(AbstractOperator, CPUOperator):
```
jdye64, line 16: Is this basically the same logic in the `batch.py` file `_` method?

nemo_retriever/src/nemo_retriever/examples/batch_pipeline.py
```diff
@@ -93,521 +88,173 @@ def _configure_logging(log_file: Optional[Path], *, debug: bool = False) -> tupl
...
-        0.0,
-        "--nemotron-parse-actors",
-        min=0.0,
-        help=(
-            "Actor count for Nemotron Parse stage "
-            "(enables parse-only mode when > 0.0 with parse GPU/batch-size)."
-        ),  # noqa: E501
-    ),
-    # fmt: on
+        None, "--detection-summary-file", path_type=Path, dir_okay=False
```
jdye64: Can we add back to `help` to all of these arguments? Those are loaded by the Typer cli and present the user with the message about what the arguments do

nemo_retriever/src/nemo_retriever/graph/content_operators.py
jperez999: it is, I was moving that specific logic used within the Actors to be able to remove the ingest_modes folder all together.

nemo_retriever/src/nemo_retriever/text_embed/runtime.py
jperez999: assuming default is max, so ensuring threshold is respected.

jperez999: if its not we can pass that arg completely to the user and just set a regular default.

nemo_retriever/src/nemo_retriever/audio/chunk_actor.py
jperez999: Here its because everything the mediachunk actor does is in fsspec and that is not on the GPU IIRC. If it is, then it should get changed, but if it works the same with and without GPU we can keep the lowest common denominator (CPUOperator) as the identifier, until we find a reason not too. What do you think @jdye64 ?

nemo_retriever/src/nemo_retriever/audio/asr_actor.py
jperez999: so this model should be getting loaded via cuda, so this is an oversight, will fix. Thanks.

nemo_retriever/src/nemo_retriever/audio/chunk_actor.py
jdye64: leave it as is. Makes sense why its that way. Was mostly just curious to learn myself.

nemo_retriever/src/nemo_retriever/vector_store/lancedb_utils.py
jperez999: Because of the way these functions are called, they are not coming back as a dict or a pd.series but a tuple or attribute row but that is not the only way to call them. Also pd.Series would not be right here because it represents a column. We could make something custom, but figure we could come back to that?

nemo_retriever/src/nemo_retriever/vector_store/lancedb_utils.py
jperez999: same as above

nemo_retriever/src/nemo_retriever/vector_store/lancedb_utils.py
jdye64: Yeah, lets just come back later. I just naively assumed they were pd.Series but if not I'll just resolve these.

Issue #1669, opened 2026-03-20
[FEA]: Make it easier to configure different inference options in nemo_retriever library


randerzander:
### Is this a new feature, an improvement, or a change to existing functionality?

New Feature

### How would you describe the priority of this feature request

Significant improvement

### Please provide a clear description of problem this feature solves

We should make the [no GPU required](https://github.com/NVIDIA/NeMo-Retriever/tree/main/nemo_retriever#run-with-remote-inference-no-local-gpu-required) experience simpler.

### Describe the feature, and optionally a solution or implementation and any alternatives

Currently the user has to specify all the build.nvidia.com URLs. It would be much nicer if they could just setup the ingestor more simply with something like:
```
ingestor = create_ingestor(run_mode="batch", inference="build.nvidia.com")
```

So `inference` could have several values:
`local`: use visible local GPUs (and fail with relevant error details if none found)

`build.nvidia.com` - use build hosted inference

`nims` - self hosted NIM services . This probably needs an additional argument `endpoints` which could be path to a yaml file defining the endpoints, or a python dict of NIM endpoints

`openrouter` - future need once more retriever models are hosted on openrouter

PR #1711, opened 2026-03-24
Making build.nvidia.com hosted inference easier

randerzander:
Partially addresses #1669 

Hopefully successfully signed commits this time

### DIFF
--- a/nemo_retriever/src/nemo_retriever/nim/nim.py
+++ b/nemo_retriever/src/nemo_retriever/nim/nim.py
@@ -217,7 +217,6 @@ def invoke_nemotron_parse_batches(
     api_key: Optional[str] = None,
     timeout_s: float = 120.0,
     max_batch_size: int = 8,
-    max_pool_workers: int = 16,
     max_retries: int = 10,
     max_429_retries: int = 5,
 ) -> List[Any]:
@@ -239,13 +238,13 @@ def invoke_nemotron_parse_batches(
         return []
 
     ranges = _chunk_ranges(n, int(max_batch_size))
-    flattened: List[Optional[Any]] = [None] * n
+    out: List[Any] = []
 
     tool_spec = [{"type": "function", "function": {"name": str(tool_name)}}]
     tool_choice = {"type": "function", "function": {"name": str(tool_name)}}
 
-    def _invoke_one_batch(start: int, end: int, endpoint_url: str) -> Tuple[int, int, List[Any]]:
-        per_image: List[Any] = []
+    for idx, (start, end) in enumerate(ranges):
+        endpoint_url = invoke_urls[idx % len(invoke_urls)]
         for b64 in image_b64_list[start:end]:
             mime = _mime_from_b64(b64)
             payload = {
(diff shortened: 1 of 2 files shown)


nemo_retriever/src/nemo_retriever/ocr/ocr.py
```diff
@@ -995,9 +1018,10 @@ def nemotron_parse_page_elements(
                 crop_meta: List[Tuple[str, List[float]]] = [(label, bbox) for label, bbox, _b64 in crops]
 
                 if crop_b64s:
-                    response_items = invoke_image_inference_batches(
+                    response_items = invoke_nemotron_parse_batches(
```
jperez999, line 1021: are you trying to remove the use of non nemotron models? this seems to overwrite logic completely and no longer use other OCR models.

nemo_retriever/src/nemo_retriever/ocr/ocr.py
```diff
@@ -885,10 +886,31 @@ def _extract_parse_text(response_item: Any) -> str:
     if isinstance(response_item, str):
         return response_item.strip()
     if isinstance(response_item, dict):
+        tool_calls = response_item.get("tool_calls")
```
jperez999, line 889: What is this logic supposed to accomplish. Are we expecting to send tool calls to OCR actor?

nemo_retriever/src/nemo_retriever/nim/nim.py
```diff
@@ -115,6 +115,20 @@ def _mime_from_b64(b64: str) -> str:
     return "image/png"
 
 
+def _normalize_chat_completions_response(response_json: Any) -> Any:
```
jperez999, line 118: If this is openai specifc (API) why is this in the NIM file?

nemo_retriever/src/nemo_retriever/ocr/ocr.py
randerzander: see [the example snippet](https://build.nvidia.com/nvidia/nemotron-parse?snippet_tab=Python) for nemotron_parse inference on build

it requires tool calling and specifying the "markdown_bbox" tool

nemo_retriever/src/nemo_retriever/nim/nim.py
randerzander: Claude's explanation:
```
1. nim.py — invoke_nemotron_parse_batches: New function that sends images to the chat completions endpoint using tool calling (markdown_bbox tool). The old code reused invoke_image_inference_batches which doesn't speak the chat completions / tool-call contract that build.nvidia.com requires.                                                                                                                          

  2. ocr.py — nemotron_parse_page_elements: Switch the remote call sites from invoke_image_inference_batches → invoke_nemotron_parse_batches, and add nemotron_parse_model_name kwarg support so callers can specify the hosted model name.                                                                                                                                                                                   

  3. ocr.py — _extract_parse_text: Handle the tool call response format — drill into tool_calls[0].function.arguments (a JSON string), parse it, and extract the "markdown" key from the result.                   

  4. ocr.py — NemotronParseActor: Add nemotron_parse_model_name param so the model name flows through to the remote call.                                                                                          

  5. nim.py — _normalize_chat_completions_response: Helper to unwrap choices[0].message from the chat completions envelope before passing to _extract_parse_text. 
 ```

nemo_retriever/src/nemo_retriever/ocr/ocr.py
randerzander: Claude:

The two functions are in entirely separate code paths: nemotron_parse_page_elements handles nemotron-parse, and the OCR function at line 567 handles nemotron-ocr. They were never sharing the call site.

PR #1715, opened 2026-03-24
Kheiss/5966538 - document Python 3.12+ as a prerequisite for NeMo Retriever Library

kheiss-uwzoo:
document Python 3.12+ as a prerequisite for NeMo Retriever Library

## Description
NeMo Retriever Library (nemo_retriever) and the test harness declare requires-python = ">=3.12" in their pyproject.toml files, but the extraction docs did not state a Python version. Users on Python 3.10 or 3.11 often hit confusing dependency resolution failures when installing from PyPI.

This change makes the Python >= 3.12 requirement explicit in the places called out in the issue: prerequisites, support matrix, Docker Compose quickstart (where the Python client and CLI are used), and library-mode quickstart.

## Changes
prerequisites.md — Add Python >= 3.12 to software requirements; clarify UV as optional; replace the vague “install Python later” note with guidance to use Python 3.12+ (including uv venv --python 3.12).
support-matrix.md — Add a “Software environment” subsection with a supported-versions table and a short note that 3.10/3.11 are not supported for PyPI client/CLI/library installs.
quickstart-guide.md — Add an Important admonition before Step 2 (Python/CLI examples) stating Python 3.12+ is required, with links to prerequisites and the support matrix.
quickstart-library-mode.md — Add one sentence that Python 3.12+ is required, with a link to prerequisites.


docs/docs/extraction/support-matrix.md
```diff
@@ -7,6 +7,15 @@ Before you begin using [NeMo Retriever Library](overview.md), ensure that you ha
     NVIDIA Ingest (nv-ingest) has been renamed NeMo Retriever Library.
 
 
+## Software environment
+
+| Component | Supported versions |
+|-----------|-------------------|
+| Python (for PyPI-installed client, CLI, and library workflows) | `>= 3.12` |
```
sosahi: we have the SW requirements already in the prerequisites.md file. It's probably cleaner to keep the support matrix focused on HW support.

PR #1716, opened 2026-03-24
Align benchmarking.md with tools/harness/test_configs.yaml

kheiss-uwzoo:
# Align benchmarking.md with tools/harness/test_configs.yaml

Updates the NV-Ingest integration testing / benchmarking guide so the documented YAML matches the harness defaults and structure in tools/harness/test_configs.yaml, and removes internal machine paths from examples.

## Changes
Active config sample: api_version is documented as v2 (default). sparse is false. Docker Compose profiles are shown under compose.profiles (including retrieval and reranker for recall), instead of a flat profiles list.
Pre-configured datasets: Example paths use /datasets/nv-ingest/... instead of developer-specific paths. The earnings dataset path is corrected to earnings_consulting (fixes the “conusulting” typo). bo20 documents extract_infographics: true, consistent with the harness.
Dataset extraction table: The bo20 Infographics column is updated from disabled to enabled so it matches the YAML block.
Configuration reference: Infrastructure options describe compose.profiles and note that YAML nests profiles under compose while the loader exposes them as profiles.
Precedence example: Comments and env overrides assume the active section defaults to api_version: v2, and the example shows overriding down to v1 with API_VERSION=v1 so the narrative matches current defaults.

### DIFF
--- a/docs/docs/extraction/benchmarking.md
+++ b/docs/docs/extraction/benchmarking.md
@@ -68,7 +68,7 @@ active:
   test_name: null  # Auto-generated if null
   
   # API Configuration
-  api_version: v2  # v1 or v2
+  api_version: v2  # v2
   pdf_split_page_count: null  # V2 only: pages per chunk (null = default 32)
   
   # Infrastructure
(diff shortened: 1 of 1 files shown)


docs/docs/extraction/benchmarking.md
```diff
@@ -173,11 +176,11 @@ Settings are applied in order of priority:
 
 Example:
 ```bash
-# YAML active section has api_version: v1
+# YAML active section has api_version: v2
 # Dataset bo767 has extract_images: false
 # Override via environment variable (highest priority)
-EXTRACT_IMAGES=true API_VERSION=v2 uv run nv-ingest-harness-run --case=e2e --dataset=bo767
-# Result: Uses bo767 path, but extract_images=true (env override) and api_version=v2 (env override)
+EXTRACT_IMAGES=true API_VERSION=v1 uv run nv-ingest-harness-run --case=e2e --dataset=bo767
```
jdye64: See v2 in some places and then v1 in other. I honestly don't know which one is correct? @jioffe502 can you help clarify?

jioffe502: v2

PR #1720, opened 2026-03-25
tabular data ingestion

tomer-levin-nv:
## Checklist

### DIFF
--- a/nemo_retriever/src/nemo_retriever/tabular_data/ingestion/dal/utils_dal.py
+++ b/nemo_retriever/src/nemo_retriever/tabular_data/ingestion/dal/utils_dal.py
@@ -103,14 +103,14 @@ def add_edges(edges_data):
                 data.v1_label,
                 data.v1_identity_props,
                 data.v1_on_create_props,
-                data.v1_on_match_props,
+                data.v1_on_match_props
             )
             yield node as v1
             call apoc.merge.node.eager(
                 data.v2_label,
                 data.v2_identity_props,
                 data.v2_on_create_props,
-                data.v2_on_match_props,
+                data.v2_on_match_props
             )
             yield node as v2
             with v1, v2, data, data.edge_identity_props as e_identity_props


nemo_retriever/src/nemo_retriever/tabular_data/ingestion/extract_data.py
```diff
@@ -0,0 +1,80 @@
...
+    Returns:
+        data dict with keys: tables, columns, views, pks, fks.
+    """
+    if connector is None:
+        db_path = (
+            params.connection_string
+            if params is not None and params.connection_string is not None
+            else "./spider2.duckdb"
+        )
+        connector = DuckDB(db_path)
```
yuvalshkolar: @tomer-levin-nv Let's remove this. The connector should only be initiated from outside as it's abstract class.

nemo_retriever/tabular-dev-tools/duckdb_connector.py
```diff
@@ -0,0 +1,203 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-25, NVIDIA CORPORATION & AFFILIATES.
+# All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""DuckDB connector for in-process SQL execution.
```
yuvalshkolar, line 5: Move this file into "tabular-dev-tools"

nemo_retriever/src/nemo_retriever/graph/tabular_ingest_operator.py
```diff
@@ -0,0 +1,153 @@
...
+    8. :class:`_LanceDBWriteActor`       — streaming write to LanceDB
+    """
+
+    def __init__(
+        self,
+        *,
+        tabular_params: TabularExtractParams | None = None,
+        embed_params: EmbedParams | None = None,
+        vdb_params: VdbUploadParams | None = None,
+        allow_no_gpu: bool = False,
```
jperez999: I see you created parameters for the other stages 
"TabularDescriptionParams",    "TabularFetchParams",   "TabularSemanticLayerParams",    "TabularUsageWeightsParams"
Is there a reason they are not plumbed here? (because the stages arent available yet...?)

nemo_retriever/src/nemo_retriever/graph/tabular_ingest_operator.py
```diff
@@ -0,0 +1,153 @@
...
+            store_relational_db_in_neo4j,
+        )
+        from nemo_retriever.tabular_data.ingestion.embeddings import fetch_tabular_embedding_dataframe
+        from nemo_retriever.tabular_data.neo4j import get_neo4j_conn
+        from nemo_retriever.utils.ray_resource_hueristics import (
+            gather_cluster_resources,
+            resolve_requested_plan,
+        )
+
+        neo4j_conn = get_neo4j_conn()
```
jperez999: Do we still have the problems with neo4j? Should we have some kind of check that tells the user they need to install it when its not available in the environment?

nemo_retriever/src/nemo_retriever/graph/tabular_ingest_operator.py
```diff
@@ -0,0 +1,153 @@
...
+from nemo_retriever.graph.cpu_operator import CPUOperator
+from nemo_retriever.params import EmbedParams
+from nemo_retriever.params import TabularExtractParams
+from nemo_retriever.params import VdbUploadParams
+
+
+_TABULAR_TABLE = "nv-ingest-tabular"
+
+
+class TabularIngestOperator(AbstractOperator, CPUOperator):
```
jperez999: Make sure to join this to the new archetype pattern.

nemo_retriever/src/nemo_retriever/graph/tabular_ingest_operator.py
```diff
@@ -0,0 +1,153 @@
...
+                allow_no_gpu=self._allow_no_gpu,
+            )
+
+            embed_kwargs = build_embed_kwargs(self._embed_params, include_batch_tuning=True)
+            endpoint = (embed_kwargs.get("embedding_endpoint") or embed_kwargs.get("embed_invoke_url") or "").strip()
+            embed_actor_num_gpus = 0.0 if endpoint else requested_plan.get_embed_gpus_per_actor()
+
+            rd_dataset = rd_dataset.repartition(target_num_rows_per_block=requested_plan.get_embed_batch_size())
+            rd_dataset = rd_dataset.map_batches(
+                _BatchEmbedActor,
```
jperez999: This should be pluggable... right, like I should be able to use any embedder actor/op here so long as I get embeddings right?

nemo_retriever/src/nemo_retriever/graph/tabular_ingest_operator.py
```diff
@@ -0,0 +1,153 @@
...
+                fn_constructor_kwargs={"params": self._embed_params},
+            )
+
+        # Step 8: write to LanceDB
+        if self._vdb_params is not None:
+            tabular_vdb_params = self._vdb_params.model_copy(
+                update={"lancedb": self._vdb_params.lancedb.model_copy(update={"table_name": _TABULAR_TABLE})}
+            )
+            rd_dataset = rd_dataset.map_batches(
+                _LanceDBWriteActor,
```
jperez999: This should be interchangeable, user may want to use something else. Maybe milvus or another VDB that they have created an operator for.

nemo_retriever/src/nemo_retriever/ingest_modes/batch.py
```diff
@@ -965,14 +956,14 @@ def embed(
...
             self._rd_dataset = input_dataset
-        if self._rd_dataset is None:
-            raise RuntimeError(
-                "No Ray Dataset to embed. Provide input_dataset or run .files(...) / .extract(...) first."
-            )
 
         from nemo_retriever.params.utils import build_embed_kwargs
 
         resolved = _coerce_params(params, EmbedParams, kwargs)
+        if self._rd_dataset is None:
```
jperez999: This file no longer exists.

nemo_retriever/src/nemo_retriever/graph/tabular_ingest_operator.py
```diff
@@ -0,0 +1,153 @@
...
+            requested_plan = resolve_requested_plan(
+                cluster_resources=cluster_resources,
+                allow_no_gpu=self._allow_no_gpu,
+            )
+
+            embed_kwargs = build_embed_kwargs(self._embed_params, include_batch_tuning=True)
+            endpoint = (embed_kwargs.get("embedding_endpoint") or embed_kwargs.get("embed_invoke_url") or "").strip()
+            embed_actor_num_gpus = 0.0 if endpoint else requested_plan.get_embed_gpus_per_actor()
+
+            rd_dataset = rd_dataset.repartition(target_num_rows_per_block=requested_plan.get_embed_batch_size())
```
jperez999: I think this needs to be refactored. Your operator should not be doing anything with ray. Otherwise this operator cannot be used with inprocess. If you can create individual operations and connect them like we do with the graph example then you could use our executors under the hood. i.e. 
```
luminex_pipeline = extract_tabular_db_dataOp() >> store_relational_db_in_neo4jOp() >> populate_tabular_semantic_layer() >> populate_tabular_usage_weights() >> generate_tabular_descriptions() >> fetch_tabular_embedding_dataframe() >> BatchEmbedActor() >> LanceDBWriteActor()
```
then we do

```
from nemo_retriever.graph import RayDataExecutor

executor = RayDataExecutor(luminex_pipeline, batch_size=8)
result_ds = executor.ingest(db_uri=path/to/db)
```
Regardless, I think that whether you keep your current setup or adopt the graph setup shown here, we need to get ray out of here because it may clash with the ray executor setup and execution.

nemo_retriever/src/nemo_retriever/tabular_data/ingestion/graph/model/node.py
```diff
@@ -0,0 +1,124 @@
+import logging
+import uuid
+from json import JSONEncoder
+from nemo_retriever.tabular_data.ingestion.graph.model.reserved_words import Labels
+
+logger = logging.getLogger(__name__)
+
+
+class Node:
```
jperez999: This might be confusing to external developers to see a second class called Node. Can we at least add a doc string here that explains that is only used within tabular_data, or if we could make the name descriptive to that effect, that might help limit confusion.

nemo_retriever/src/nemo_retriever/tabular_data/neo4j/neo4j_connection.py
```diff
@@ -0,0 +1,105 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+Neo4j connection management for the tabular_data stack.
+"""
+
+import os
+import logging
+
+from neo4j import GraphDatabase, Result, RoutingControl
```
jperez999, line 12: This is good now, we can import this ourselves (OSRB)? if not I think we should add a check to ensure it is in the environment and if not, we error when the users tries to access and tell them to install with "x" command.

.gitignore
```diff
@@ -248,4 +251,18 @@ lancedb/
...
+
+# Local environment variables (credentials) — never commit
+.env
+
 nemo_retriever/run_results/
+
+nemo_retriever/PIPELINE_FLOW.md
+
+# Batch-generated SQL (debug_retriever_sql --batch output)
+nemo_retriever/src/nemo_retriever/relational_db/generate_sql/generated_sql/
```
jperez999: Are these expected output dump locations, for runs? Is this the result of a variable value (default/input)?

jperez999 (changes requested):

nemo_retriever/src/nemo_retriever/graph/tabular_ingest_operator.py
tomer-levin-nv: removed these params for now

nemo_retriever/src/nemo_retriever/graph/tabular_ingest_operator.py
tomer-levin-nv: we are not familiar with neo4j issues / problems. 
Illiumex used it for 5 years and we never encountered neo4j issues.

nemo_retriever/src/nemo_retriever/graph/tabular_ingest_operator.py
tomer-levin-nv: fixed. it has now CPU Actor only. GPU is not suited here. 
I'll split now (in this PR) the embed so this part will be able to use GPU

.gitignore
tomer-levin-nv: removed, thanks

nemo_retriever/src/nemo_retriever/tabular_data/neo4j/neo4j_connection.py
tomer-levin-nv: The neo4j Python driver is already a declared hard dependency in pyproject.toml, so it will always be present in the environment. No runtime check is needed.

nemo_retriever/src/nemo_retriever/graph/tabular_ingest_operator.py
tomer-levin-nv: @jperez999 I removed the tabular_ingest_operator and broke it to two + _BatchEmbedActor.
The graph looks like this now:
graph = (
        Graph()
        >> TabularSchemaExtractOp(tabular_params=TABULAR_PARAMS)
        >> TabularFetchEmbeddingsOp()
        >> _BatchEmbedActor(params=EMBED_PARAMS)
    )

so all prev emebedding code was removed and we use the same actor as unstructured.
hmm... is there an operator wrapping the _BatchEmbedActor? this is how we expect the operators (users) use the _batch?

nemo_retriever/src/nemo_retriever/graph/tabular_ingest_operator.py
tomer-levin-nv: done by the new operators above.

nemo_retriever/src/nemo_retriever/graph/tabular_ingest_operator.py
tomer-levin-nv: done by the new operators above.

nemo_retriever/src/nemo_retriever/tabular_data/ingestion/graph/model/node.py
tomer-levin-nv: renamed to Neo4jNode.
also removed the graph folder.

PR #1741, opened 2026-03-27
Fix duplicate Table/Chart sections in to_markdown_by_page

randerzander:
when running the core pipeline on multimodal_test.pdf and using to_markdown_by_page, table and charts get duplicated in the markdown text.

Claude's description:
When multiple chunks per page carry table/chart column data, `_collect_page_record` appended a section per chunk, producing 3× Table/Chart headers with identical content. `_dedupe_blocks` couldn't catch this because auto-incremented headers made identical blocks appear distinct.

**Fix:** deduplicate sections by content-only key (stripping the numeric header) before combining with text blocks, and filter out text blocks whose content is already covered by a labeled section.


nemo_retriever/README.md
```diff
@@ -92,31 +92,27 @@ You can inspect how recall accuracy optimized text chunks for various content ty
...
 
 # markdown formatted table from the first page
 >>> chunks[1]["text"]
-'| Table | 1 |\n| This | table | describes | some | animals, | and | some | activities | they | might | be | doing | in | specific |\n| locations. |\n| Animal | Activity | Place |\n| Giraffe | Driving | a | car | At | the | beach |\n| Lion | Putting | on | sunscreen | At | the | park |\n| Cat | Jumping | onto | a | laptop | In | a | home | office |\n| Dog | Chasing | a | squirrel | In | the | front | yard |\n| Chart | 1 |'
+'| This | table | describes | some | animals, | and | some | activities | they | might | be | doing | in | specific |\n| locations. |\n| Animal | Activity | Place |\n| Giraffe | Driving | a | car | At | the | beach |\n| Lion | Putting | on | sunscreen | At | the | park |\n| Cat | Jumping | onto | a | laptop | In | a | home | office |\n| Dog | Chasing | a | squirrel | In | the | front | yard |\n| Chart | 1 |'
```
randerzander: Notice here that the text surrounding the table (captured in crop because it increases relevance for recall) is also included in the markdown format.

ToDo: exclude surrounding text from markdown formatting

PR #1746, opened 2026-03-28
Fix nemotron_parse: disable KV cache and suppress pdfium text

randerzander:
Previously nemotron-parse inference wasn't actually happening (transformers version incompatibility), but fallback logic hid the failure by using pdfium extracted text instead.

This PR fixes severail things:
1. Enable nemotron_parse inference by disabling kv cache
2. Update handling to use nemotron_parse text instead of pdfium text
3. Don't include the x,y and content label headers in extracted text
4. convert latek table format to markdown

Claude description:
- Disable KV cache in `NemotronParseV12.invoke` (`use_cache=False`) to fix `AttributeError: 'tuple' object has no attribute 'update'` with newer transformers versions that use `DynamicCache` objects
- Suppress pdfium text extraction when `text_extraction_method="nemotron_parse"` so pdfium text is not duplicated alongside nemotron_parse output

### DIFF
--- a/nemo_retriever/src/nemo_retriever/model/local/nemotron_parse_v1_2.py
+++ b/nemo_retriever/src/nemo_retriever/model/local/nemotron_parse_v1_2.py
@@ -12,7 +12,11 @@ import numpy as np
 import torch
 from PIL import Image
 
-_POSITION_TAG_RE = re.compile(r"<(?:x|y)_[\d.]+>|<class_\S+>")
+# Matches one detection: <x_...><y_...>CONTENT<x_...><y_...><class_LABEL>
+_DETECTION_RE = re.compile(
+    r"<x_[\d.]+><y_[\d.]+>(.*?)<x_[\d.]+><y_[\d.]+><class_([\w-]+)>",
+    re.DOTALL,
+)
 _TABULAR_RE = re.compile(r"\\begin\{tabular\}\{[^}]*\}(.*?)\\end\{tabular\}", re.DOTALL)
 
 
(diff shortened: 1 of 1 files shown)


nemo_retriever/src/nemo_retriever/model/local/nemotron_parse_v1_2.py
```diff
@@ -135,10 +174,11 @@ def invoke(
         ).to(self._device)
 
         with torch.inference_mode():
-            outputs = self._model.generate(**inputs, generation_config=self._generation_config)
+            outputs = self._model.generate(**inputs, generation_config=self._generation_config, use_cache=False)
 
         decoded = self._processor.batch_decode(outputs, skip_special_tokens=True)
-        return decoded[0] if decoded else ""
+        raw = decoded[0] if decoded else ""
+        return _parse_detections(raw)
```
edknv, line 181: Have we considered the postprocessing function provided by nemotron-parse? https://huggingface.co/nvidia/NVIDIA-Nemotron-Parse-v1.2#postprocessing
```python
# Specify output formats for postprocessing
table_format = 'latex' # latex | HTML | markdown
text_format = 'markdown' # markdown | plain
blank_text_in_figures = False # remove text inside 'Picture' class
texts = [postprocess_text(text, cls = cls, table_format=table_format, text_format=text_format, blank_text_in_figures=blank_text_in_figures) for text, cls in zip(texts, classes)]
```

randerzander: I didn't know it exists. Feel free to either close this PR and replace w/ your own, or commit directly on top of my branch if you can

edknv: Postprocessing from https://huggingface.co/nvidia/NVIDIA-Nemotron-Parse-v1.2#postprocessing was added in https://github.com/NVIDIA/NeMo-Retriever/pull/1764

PR #1754, opened 2026-03-30
QA eval pipeline for retrieval

KyleZheng1284:
## Description
- Adds a pluggable QA evaluation harness for measuring Retrieval quality end-to-end using multi-tier scoring.

### Capabilities:
- Multi-tier scoring -- Tier 1 retrieval recall (answer-in-context), Tier 2 programmatic (exact match + token F1), and Tier 3 LLM-as-judge (1-5 rubric) run together in a single pass at zero extra retrieval cost.
- Full-page markdown retrieval -- Reconstructs complete document pages from NeMo Retriever extraction records via to_markdown_by_page()
- Pluggable retrieval -- Any retrieval system (vector search, agentic, hybrid, BM25) plugs in by producing a standard JSON (queries → chunks); no harness code changes required.
- Pluggable datasets -- Any CSV with query/answer columns loads via csv:path/to/file.csv; default ground truth is data/bo767_annotations.csv (1007 Q&A pairs, all modalities).
- Pluggable LLMs -- Generator and judge models swap via env var or YAML config using litellm prefix routing (nvidia_nim/, openai/, huggingface/).
- Multi-model sweep -- Set GEN_MODELS to evaluate multiple generators in a single run with side-by-side score comparisons.
- Failure classification -- Per-query categorization into correct, partial, retrieval_miss, generation_miss, no_context, thinking_truncated to pinpoint exactly where the pipeline fails.

#### Note - the csv containing the q-a pairs is a subset of the existing https://github.com/NVIDIA/NeMo-Retriever/blob/main/data/digital_corpora_10k_annotations.csv. Currently have an separate PR up with a subset annotations for only bo767 specific files here - https://github.com/NVIDIA/NeMo-Retriever/pull/1730
)## Checklist

### DIFF
--- a/nemo_retriever/src/nemo_retriever/evaluation/retrieval_loader.py
+++ b/nemo_retriever/src/nemo_retriever/evaluation/retrieval_loader.py
@@ -71,9 +71,10 @@ class RetrievalLoaderOperator(EvalOperator):
         source = self._ground_truth_csv
         try:
             loader_fn = get_qa_dataset_loader(source)
-            qa_pairs = loader_fn(self._data_dir)
         except ValueError:
             qa_pairs = load_generic_csv(source)
+        else:
+            qa_pairs = loader_fn(self._data_dir)
         retriever = FileRetriever(file_path=self._retrieval_json)
 
         coverage = retriever.check_coverage(qa_pairs)
--- a/tools/harness/README.md
+++ b/tools/harness/README.md
@@ -295,6 +295,8 @@ DATASET_DIR=/custom/path uv run nv-ingest-harness-run --case=e2e
 | `table_structure` | nemotron-table-structure-v1 model benchmarking (PyPi) | None | ✅ Available |
 | `ocr` | nemotron-ocr model benchmarking (PyPi) | None | ✅ Available |
 
+**QA Evaluation (multi-tier scoring):** For LLM-as-judge QA evaluation (retrieval quality, token F1, semantic scoring), see the **[QA Evaluation Pipeline](../../nemo_retriever/src/nemo_retriever/evaluation/README.md)** in `nemo_retriever`. Run with `retriever eval run --config nemo_retriever/examples/eval_sweep.yaml` from the repo root.
+
 **Note**: Legacy test cases (`dc20_e2e`, `dc20_v2_e2e`) have been moved to `scripts/private_local`.
 
 ### Configuration Synergy


tools/harness/ingest_bo767.py
```diff
@@ -0,0 +1,89 @@
...
+        print(f"Smoke test: ingesting 1 file: {documents[0]}")
+    else:
+        documents = all_pdfs
+        print(f"Ingesting {len(documents)} PDFs from {args.dataset_dir}")
+
+    from nemo_retriever import create_ingestor
+
+    ingestor = create_ingestor(run_mode="batch")
+    ingestor = (
+        ingestor.files(documents)
```
jperez999: How is this different from the normal batch_pipeline run? Why do we need to set this particular dataset and script here. The harness is not supposed to require more scripts to run. It should use batch_pipeline.py to run the graphs. Or we could run graphs directly.

tools/harness/extract_bo767_parquet.py
```diff
@@ -0,0 +1,99 @@
...
+        print(f"Smoke test: extracting 1 file: {documents[0]}")
+    else:
+        documents = all_pdfs
+        print(f"Extracting {len(documents)} PDFs from {args.dataset_dir}")
+
+    print(f"Output dir: {args.output_dir}")
+
+    from nemo_retriever import create_ingestor
+
+    ingestor = create_ingestor(run_mode="batch")
```
jperez999: Again, same as other script, why do we need this, this can be dynamically created?

tools/harness/export_retrieval_nemo.py
```diff
@@ -0,0 +1,283 @@
...
+            f"conventions for source_id.\n"
+            f"  Re-run extract_bo767_parquet.py with the same --dataset-dir "
+            f"used for ingest_bo767.py.",
+            file=sys.stderr,
+        )
+        sys.exit(1)
+    print(f"  Page index key check: {matched}/{len(sampled)} sampled source_ids found")
+
+
+def main() -> int:
```
jperez999: Why not make this a tool we can call via import, instead of a main function.

tools/harness/export_retrieval_nemo.py
```diff
@@ -0,0 +1,283 @@
...
+import sys
+import time
+
+_HERE = os.path.dirname(os.path.abspath(__file__))
+
+DEFAULT_EMBEDDER = "nvidia/llama-nemotron-embed-1b-v2"
+BATCH_SIZE = 50
+
+
+def load_queries(csv_path: str) -> list[dict]:
```
jperez999: I feel like these tools would be good to have even outside the harness, no? Like if I run an evaluation, without the harness, I will need to use all these functions to actually get results right?

tools/harness/build_page_markdown_index.py
```diff
@@ -0,0 +1,117 @@
...
+import time
+from collections import defaultdict
+from pathlib import Path
+
+import numpy as np
+
+_HERE = os.path.dirname(os.path.abspath(__file__))
+
+
+def main() -> int:
```
jperez999: seems like this is also a tool, where I send it a parquet file (could be dataframe) and then you create the page level markdown. This definitely is useful outside of harness.

nemo_retriever/src/nemo_retriever/evaluation/types.py
```diff
@@ -0,0 +1,117 @@
+"""
+Protocol definitions and dataclasses for the QA evaluation pipeline.
```
jperez999, line 6: I think the tools/harness is not where we want our code, I think we want it in the nv-ingest/nemo_retriever/ sub folder.

tools/harness/src/nv_ingest_harness/utils/qa/orchestrator.py
```diff
@@ -0,0 +1,304 @@
...
+from nv_ingest_harness.utils.qa.scoring import answer_in_context, classify_failure, token_f1
+from nv_ingest_harness.utils.qa.types import (
+    AnswerJudge,
+    LLMClient,
+    QAQueryResult,
+    RetrieverStrategy,
+)
+
+
+class QAEvalPipeline:
```
jperez999: We need to refactor this to work as an AbstractOperator type. Inherit from that and put everything this does in the appropriate life cycle hook.

nemo_retriever/src/nemo_retriever/evaluation/judges.py
```diff
@@ -0,0 +1,165 @@
...
+
+_JUDGE_USER_TEMPLATE = """\
+Question: {query}
+
+Reference answer: {reference}
+
+Candidate answer: {candidate}"""
+
+
+class LLMJudge:
```
jperez999, line 61: This should be an AbstractOperator. The process portion is the judge method. I should be able to pass a list of retrievals ( record in dataframe with query, golden_text and candidates) and the judge can provide a response for each record to judge.

nemo_retriever/src/nemo_retriever/evaluation/generators.py
```diff
@@ -0,0 +1,175 @@
...
+    """Build the OpenAI-style messages list for a RAG prompt."""
+    context = "\n\n---\n\n".join(chunks) if chunks else "(no context retrieved)"
+    user_content = _RAG_USER_TEMPLATE.format(context=context, query=query)
+    return [
+        {"role": "system", "content": _RAG_SYSTEM_PROMPT},
+        {"role": "user", "content": user_content},
+    ]
+
+
+class LiteLLMClient:
```
jperez999, line 50: This should be an AbstractOperator class. It should be able to run alone, Like if I pass it a dataframe with query, and context it should provide me an answer for that (basically your generate call, plumbed through process lifecycle hook.

tools/harness/ingest_bo767.py
KyleZheng1284: thank you for pointing this out yes `batch_pipeline.py ` does seem to cover the exact same functionality that this script does and some more

KyleZheng1284: ingestion now goes through `python -m nemo_retriever.examples.graph_pipeline `

tools/harness/extract_bo767_parquet.py
KyleZheng1284: extraction is handled by graph_pipeline --save-intermediate now

tools/harness/export_retrieval_nemo.py
KyleZheng1284: The core evaluation logic has been moved into nemo_retriever.evaluation (importable package, pip-installable via nemo_retriever[eval]).

tools/harness/build_page_markdown_index.py
KyleZheng1284: The core evaluation logic has been moved into nemo_retriever.evaluation (importable package, pip-installable via nemo_retriever[eval]).

tools/harness/src/nv_ingest_harness/utils/qa/orchestrator.py
KyleZheng1284: QAEvalPipeline inherits from EvalOperator, which inherits from graph.AbstractOperator

nemo_retriever/src/nemo_retriever/evaluation/generators.py
KyleZheng1284: QAEvalPipeline now inherits from EvalOperator, which inherits from graph.AbstractOperator

tools/harness/export_retrieval_nemo.py
KyleZheng1284: core evaluation logic has been moved into nemo_retriever.evaluation (importable package, pip-installable via nemo_retriever[eval])

nemo_retriever/src/nemo_retriever/evaluation/config.py
```diff
@@ -0,0 +1,365 @@
...
+
+_ENV_VAR_RE = re.compile(r"\$\{([^}]+)\}")
+
+_VALID_SECTION_SETS = (
+    frozenset(("generators", "judge")),
+    frozenset(("models", "evaluations")),
+)
+
+
+def _expand_env_vars(value: Any) -> Any:
```
jperez999, line 92: For what scenario do we need recursive expansion?

nemo_retriever/src/nemo_retriever/evaluation/config.py
```diff
@@ -0,0 +1,365 @@
...
+            judge_key = entry["judge"]
+            if judge_key not in models:
+                raise ValueError(f"Evaluation references unknown model {judge_key!r}. " f"Available: {list(models)}")
+            if gen_key not in seen_generators:
+                seen_generators[gen_key] = {"name": gen_key, **models[gen_key]}
+
+        config.setdefault("generators", list(seen_generators.values()))
+        first_judge_key = evals[0]["judge"]
+        distinct_judges = {e["judge"] for e in evals}
+        if len(distinct_judges) > 1:
```
jperez999, line 142: Greptile has a good comment there about a potential KeyError when accessing an empty []. But also, when using the legacy method, I can only define one combo, so only one run at a time?

nemo_retriever/src/nemo_retriever/evaluation/config.py
```diff
@@ -0,0 +1,365 @@
...
+        timeout=judge_cfg.get("timeout", default_timeout),
+        max_workers=execution.get("max_workers", 8),
+    )
+
+    scoring_op = ScoringOperator()
+
+    return loader >> gen_op >> judge_op >> scoring_op
+
+
+def build_eval_pipeline(config: dict) -> "QAEvalPipeline":
```
jperez999, line 315: Is this for a collection of sweeps? So essentially this will pipe through any number of LLMs you added in your config and then run all those against one judge?

nemo_retriever/src/nemo_retriever/evaluation/orchestrator.py
```diff
@@ -0,0 +1,454 @@
...
+    GenerationResult,
+    JudgeResult,
+    LLMClient,
+    RetrieverStrategy,
+)
+
+logger = logging.getLogger(__name__)
+
+
+class QAEvalPipeline(EvalOperator):
```
jperez999, line 50: why is this not serializable for ray. All the other Operators have models as well. And we instantiate outside ray and use the references. If your using operators, you should have same capabilities. You are not seeing that?

nemo_retriever/src/nemo_retriever/evaluation/orchestrator.py
```diff
@@ -0,0 +1,454 @@
...
+            reference = pair.get("reference_answer") or pair["answer"]
+            retrieval = self.retriever.retrieve(query, self.top_k)
+            return idx, {
+                "query": query,
+                "reference_answer": reference,
+                "context": retrieval.chunks,
+                "_retrieval_metadata": retrieval.metadata,
+            }
+
+        with ThreadPoolExecutor(max_workers=self.max_workers) as pool:
```
jperez999, line 278: should we be using the threadpool here? or can we just create more ray workers for this?

nemo_retriever/src/nemo_retriever/evaluation/retrieval_loader.py
```diff
@@ -0,0 +1,97 @@
...
+    Constructor kwargs are all serializable strings, so
+    ``get_constructor_kwargs()`` works for Ray reconstruction.
+    """
+
+    required_columns: ClassVar[tuple[str, ...]] = ()
+    output_columns: ClassVar[tuple[str, ...]] = ("query", "reference_answer", "context")
+
+    def __init__(
+        self,
+        retrieval_json: str,
```
jperez999, line 39: So this is the actual chunks retrieved per question right? Just in a nested dict format...?

nemo_retriever/src/nemo_retriever/evaluation/scoring_operator.py
```diff
@@ -0,0 +1,34 @@
...
+
+from __future__ import annotations
+
+from typing import Any, ClassVar
+
+from nemo_retriever.evaluation.eval_operator import EvalOperator
+from nemo_retriever.evaluation.scoring import score_dataframe
+
+
+class ScoringOperator(EvalOperator):
```
jperez999, line 15: why is the logic for this in a separate module? Talking about score_dataframe

nemo_retriever/src/nemo_retriever/examples/parquet_to_lancedb.py
```diff
@@ -0,0 +1,64 @@
...
+
+logger = logging.getLogger(__name__)
+app = typer.Typer()
+
+LANCEDB_URI = "lancedb"
+LANCEDB_TABLE = "nv-ingest"
+
+
+@app.command()
+def main(
```
jperez999: This is only for making it easier to run the  eval on already created ingestion correct? Like all this does is load up results to a lancedb instance. I dont think we want this to be in the examples. An example should run through the entire flow ingestion + retrieval. We should keep this logic for testing ourselves but does not have to be an entrypoint user's should be expected to use.

nemo_retriever/src/nemo_retriever/evaluation/retrievers.py
```diff
@@ -0,0 +1,126 @@
...
+        self._raw_keys: dict[str, str] = {}
+        for raw_key, value in raw_index.items():
+            norm = _normalize_query(raw_key)
+            self._norm_index[norm] = value
+            self._raw_keys[norm] = raw_key
+
+        self._miss_count = 0
+        self._miss_lock = threading.Lock()
+
+    def check_coverage(self, qa_pairs: list[dict]) -> float:
```
jperez999, line 89: why do you need to validate that the full set of questions is in the queries? That shouldn't matter right. I want the recall score for the questions I ask only.

nemo_retriever/src/nemo_retriever/evaluation/types.py
```diff
@@ -0,0 +1,81 @@
...
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from typing import Any, Optional, Protocol, runtime_checkable
+
+
+@runtime_checkable
+class RetrieverStrategy(Protocol):
```
jperez999, line 19: Is this supposed to take over for the current retrieval object. I think what we wanted was to use and extend the current retriever object. Isn't the retrieve API here missing something? Like what are we doing with those queries? (I dont see a parameter for vectordb, or init values that adds a db. Also where is the embedder? Is this supposed to be an abstract?

nemo_retriever/src/nemo_retriever/evaluation/types.py
```diff
@@ -0,0 +1,81 @@
...
+    """Pluggable retrieval strategy interface."""
+
+    def retrieve(self, query: str, top_k: int) -> "RetrievalResult": ...
+
+
+@runtime_checkable
+class LLMClient(Protocol):
```
jperez999, line 26: Why make these protocols, are you expecting to create multiple operators that leverage this? Couldnt these be absorbed by the target Operators?

nemo_retriever/src/nemo_retriever/evaluation/types.py
```diff
@@ -0,0 +1,81 @@
...
+    (API error, parse failure, empty candidate).  Valid scores are 1-5.
+    """
+
+    score: Optional[int] = None
+    reasoning: str = ""
+    error: Optional[str] = None
+
+
+@dataclass
+class QAQueryResult:
```
jperez999: The structure of this dataclass is catching my eye. It seems like it is collecting all the results instead of saving each result separately [list of results]. Why did you choose this setup instead? And I can only have one judgement per generation right?

nemo_retriever/src/nemo_retriever/examples/graph_pipeline.py
```diff
@@ -26,6 +26,11 @@
     python -m nemo_retriever.examples.graph_pipeline /data/pdfs \\
         --run-mode inprocess \\
         --ocr-invoke-url http://localhost:9000/v1
+
+    # Save extraction Parquet for full-page markdown (e.g. page index + export)
+    python -m nemo_retriever.examples.graph_pipeline /data/pdfs \\
+        --lancedb-uri lancedb \\
+        --save-intermediate /path/to/extracted_parquet_dir
```
jperez999, line 33: is this meant to create the parquet for your loader above?

nemo_retriever/src/nemo_retriever/examples/graph_pipeline.py
```diff
@@ -498,6 +513,13 @@ def main(
             ray_download_time = 0.0
             num_rows = result_df["source_id"].nunique() if "source_id" in result_df.columns else len(result_df)
 
+        if save_intermediate is not None:
```
jperez999, line 569: This is going to be confusing given we have a store task. Maybe you should try and use that instead of taking actions out of the graph.

nemo_retriever/src/nemo_retriever/io/markdown.py
```diff
@@ -282,3 +285,132 @@ def _label_for_subtype(subtype: Any, *, fallback: str) -> str:
...
+_MARKDOWN_PARQUET_COLUMNS = frozenset(
+    {
+        "path",
+        "source_id",
+        "page_number",
+        "text",
+        "document_type",
+        "_content_type",
+        "metadata",
+        "table",
```
jperez999: why is are there singular and plural versions of the same value?

nemo_retriever/src/nemo_retriever/export.py
```diff
@@ -0,0 +1,198 @@
...
+                "source_id": source_id,
+                "page_number": page_number,
+                "distance": seen[(source_id, page_number)],
+            }
+        )
+
+    return chunks, metadata, miss_count
+
+
+def export_retrieval_json(
```
jperez999: This function is doing a lot, its creating the retriever object, then running the full experiment then producing the json from the retrieval results. Maybe we should break this apart. From the name, I immediately thought I just gave this a retrievalresult and it creates the json.

nemo_retriever/pyproject.toml
```diff
@@ -83,6 +83,11 @@ dependencies = [
 svg = [
   "cairosvg>=2.7.0",
 ]
+eval = [
```
jperez999, line 113: Does this mean, if we dont do pip install nemo-retriever[eval] we wont be able to use the LLM generator or judge operators? I don't think that is what we want.

tools/harness/retrieval_bench/convert_traces_to_retrieval.py
```diff
@@ -0,0 +1,220 @@
...
+
+
+def _dataset_trace_dir(dataset_name: str) -> str:
+    parts = [p for p in str(dataset_name or "unknown_dataset").split("/") if p]
+    if len(parts) >= 2 and parts[0].lower() == "bright":
+        return _slugify(f"bright__{parts[1]}")
+    return _slugify(parts[-1] if parts else "unknown_dataset")
+
+
+def convert_traces(
```
jperez999: what is a trace?

tools/harness/retrieval_bench/convert_traces_to_retrieval.py
```diff
@@ -0,0 +1,220 @@
...
+    return {
+        "queries_written": len(output_queries),
+        "traces_found": len(trace_files),
+        "skipped": skipped,
+        "missing_docs": missing_docs,
+        "output_path": str(out_path),
+    }
+
+
+def main():
```
jperez999: I don't think we need to expose this as a cli tool for the user. It is a good helper function.

tools/harness/retrieval_bench/run_agentic_retrieval_bo767.py
```diff
@@ -0,0 +1,391 @@
...
+    _load_parquet_corpus,
+    _load_queries,
+)
+
+# ---------------------------------------------------------------------------
+# Main pipeline
+# ---------------------------------------------------------------------------
+
+
+def run_agentic_retrieval(
```
jperez999: So this is something that we need to do separate from graph_pipeline.py entry point? Cant we just add in the operators we want and use that same entrypoint. It would then allow us to make changes to the query file and datasets and should still get same behavior.

tools/harness/retrieval_bench/run_dense_retrieval_bo767.py
```diff
@@ -0,0 +1,457 @@
...
+Usage::
+
+    cd tools/harness
+    python retrieval_bench/run_dense_retrieval_bo767.py \\
+        --backend llama-nv-embed-reasoning-3b \\
+        --top-k 5 \\
+        --output data/test_retrieval/bo767_retrieval_dense.json
+"""
+
+from __future__ import annotations
```
jperez999: Why create a whole new file to do what graph_pipeline already mostly does?

tools/harness/src/nv_ingest_harness/cases/e2e_qa_eval.py
```diff
@@ -0,0 +1,136 @@
...
+to evaluate LLM answer quality against the newly ingested collection.
+
+Use this to measure end-to-end: extract_tables=true vs false, or different
+ingestion settings, and see how they affect downstream answer accuracy.
+"""
+
+import json
+import os
+
+from nv_ingest_harness.cases.e2e import main as e2e_main
```
jperez999: Again it seems like you are creating a whole new graph specifically for this. When what I think we want is to be able to tack on these operations to any graph.

tools/harness/src/nv_ingest_harness/cli/run.py
```diff
@@ -20,6 +20,8 @@
     "e2e_with_llm_summary",
     "recall",
     "e2e_recall",
+    "qa_eval",
```
jperez999: This harness and the one in nemo_retriever are different. We only need to worry about nemo_retriever.

tools/harness/src/nv_ingest_harness/utils/qa/__init__.py
```diff
@@ -0,0 +1,132 @@
...
+
+from __future__ import annotations
+
+import os
+from typing import Optional
+
+from nemo_retriever.evaluation.types import RetrievalResult
+
+
+class TopKRetriever:
```
jperez999: Why are you adding this in the harness. This should exist in nemo_retriever. All code changes in legacy nv-ingest can be removed unless necessary to make nemo_retriever work.

nemo_retriever/src/nemo_retriever/evaluation/config.py
KyleZheng1284: yes it will run through any combinations of llms + judges you set under` tools/harness/eval_sweep.yaml' you can also set X number of runs for any combinations you specify

nemo_retriever/src/nemo_retriever/evaluation/config.py
KyleZheng1284: the recursion walks nested dicts/lists in the YAML tree (not double-dereferencing env vars) so VAR at any nesting depth like models.nemotron.api_key: "NVIDIA_API_KEY" gets expanded

nemo_retriever/src/nemo_retriever/evaluation/config.py
KyleZheng1284: fixed

nemo_retriever/src/nemo_retriever/evaluation/orchestrator.py
KyleZheng1284: Sorry it actually does work the same as other operators where you instantiate once & then pass by reference via InprocessExecutor.

QAEvalPipeline inherits from EvalOperator which inherits from AbstractOperator. I

I'll update the docstring since it was overstating it

nemo_retriever/src/nemo_retriever/evaluation/orchestrator.py
KyleZheng1284: The retrieval here is in-memory dict lookups from FileRetriever which is a preloaded JSON that is needed from the user. I think threadpool would work better in this case than Ray since there is zero network I/O and also due to the nature of this task

nemo_retriever/src/nemo_retriever/evaluation/retrieval_loader.py
KyleZheng1284: Yes the `retrieval_json` is just the path to the pre-computed JSON

KyleZheng1284: the format is ({queries: {query_text: {chunks: [...], metadata: [...]}}})

nemo_retriever/src/nemo_retriever/examples/parquet_to_lancedb.py
KyleZheng1284: yep I have moved this to a reusable function under `nemo_retriever/utils/parquet_to_lancedb.py `(a non-example internal utility)

nemo_retriever/src/nemo_retriever/evaluation/scoring_operator.py
KyleZheng1284: it avoids coupling non-LLM scoring logic to the graph operator abstraction.

nemo_retriever/src/nemo_retriever/evaluation/types.py
KyleZheng1284: Yes its supposed to be a abstract! 

It only defines the retrieve(query, top_k) -> RetrievalResult contract so that QAEvalPipeline can accept any retriever implementation (FileRetriever, a future VectorDB retriever, or whatever you need etc.) without coupling to embedder and or db details

nemo_retriever/src/nemo_retriever/evaluation/types.py
KyleZheng1284: The orchestrator exists because multi-model sweep evaluation is a fan-out pattern (one retrieval becomes -> N generators) and result aggregation into a single comparison report. Which I don't believe the graph framework's linear execute-and-return supports the ability to merge results into a side by side.

nemo_retriever/src/nemo_retriever/examples/graph_pipeline.py
KyleZheng1284: Yep the `save-intermediate` saves the raw extraction Parquet from the ingestion pipeline

nemo_retriever/src/nemo_retriever/evaluation/types.py
KyleZheng1284: Yep, I actually removed this but the query dict structure now lives inline in the orchestrator's _aggregate_from_df and its one judge call per generation.

nemo_retriever/src/nemo_retriever/io/markdown.py
KyleZheng1284: to support backward compatibility I had some older Parquets versions which used singular column names. I can look into replacing it to the plural since that is what is being produced going forward

nemo_retriever/pyproject.toml
KyleZheng1284: This concern is mainly due to the fact that we want to use llm generator or judge operators outside of the pure eval scope for now right?

I can look into refactoring that later alongside this issue (https://github.com/NVIDIA/NeMo-Retriever/issues/1769)  but the only need for it now seems to be in the evals

tools/harness/retrieval_bench/convert_traces_to_retrieval.py
KyleZheng1284: Its a per-query JSON output from the retrieval-bench framework where each file contains a query ID and a {doc_id: score} ranking dict

nemo_retriever/src/nemo_retriever/export.py
KyleZheng1284: I also agree! split into query_lancedb() (retriever creation + querying) and write_retrieval_json() (results -> JSON file). The original function is kept as a thin convenience wrapper

tools/harness/retrieval_bench/convert_traces_to_retrieval.py
KyleZheng1284: fixed

tools/harness/src/nv_ingest_harness/utils/qa/__init__.py
KyleZheng1284: moving it would pull harness dependencies into nemo_retriever right, which isn't what we want. It makes more sense in my mind if the harness consumes the nemo_retriever protocl instead of vice versa.

tools/harness/retrieval_bench/run_dense_retrieval_bo767.py
KyleZheng1284: This script exists because retrieval-bench only works with HuggingFace datasets out of the box. We would need this file to load our extraction Parquets, expand chunk hits to full-page markdown, and output the FileRetriever JSON that our QA eval pipeline expects.

nemo_retriever/src/nemo_retriever/evaluation/ground_truth.py
```diff
@@ -0,0 +1,170 @@
...
+        for row in reader:
+            query = row.get("query", "").strip()
+            answer = row.get("answer", "").strip()
+            if query and answer:
+                records.append({**row, "query": query, "answer": answer})
+
+    return records
+
+
+def load_vidore_v3_qa(dataset_name: str, cache_dir: Optional[str] = None) -> list[dict]:
```
jperez999, line 58: If I have the vidore_v3 question and answer pairs why do I need to have the datasets library? If I have the CSV do I really still need to have datasets library?

nemo_retriever/src/nemo_retriever/evaluation/retrieval_loader.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def process(self, data: Any, **kwargs: Any) -> Any:
+        if isinstance(data, pd.DataFrame) and not data.empty:
+            return data
+
+        from nemo_retriever.evaluation.ground_truth import get_qa_dataset_loader, load_generic_csv
+        from nemo_retriever.evaluation.retrievers import FileRetriever
+
+        source = self._ground_truth_csv
+        try:
```
jperez999, line 72: wouldn't it be better if you just required the question and answer pairs. Instead of trying to handle pulling the data out of a particular file?

jperez999: This is heavily coupled with the specifics of the datasets we support. I know you have the catch all after to read csv, but I wouldnt even do that. What happens if the user has this information in a different format. Now we require them to read the information in and change it to csv. If we made it to where we took in the question answer pairs and ground truth, then we put loading that information on the user.

nemo_retriever/src/nemo_retriever/evaluation/retrievers.py
```diff
@@ -0,0 +1,135 @@
...
+
+    This is the integration point for **any** retrieval method. Vector search,
+    agentic retrieval, hybrid pipelines, BM25, rerankers, or a completely
+    custom system -- as long as it produces a JSON file in the format below,
+    the QA eval harness will generate answers and judge them identically.
+
+    Minimal required JSON format::
+
+        {
+          "queries": {
```
jperez999, line 48: I think this example is missing a set of list brackets, queries should be a list of dict records right?

nemo_retriever/src/nemo_retriever/evaluation/scoring.py
```diff
@@ -0,0 +1,318 @@
...
+            return "retrieval_miss"
+        return "generation_miss"
+
+    return "partial"
+
+
+def score_dataframe(df: pd.DataFrame) -> pd.DataFrame:
+    """Apply all programmatic scoring metrics to a DataFrame.
+
+    Input DataFrame must have ``reference_answer``, ``answer``, ``context``
```
jperez999, line 278: would be nice, if we made it possible to calculate this for any dataframe. We could keep the column names as defaults but if we add a place where the user can set their own mappings, it removes an extra step requiring the column name normalization.

nemo_retriever/src/nemo_retriever/evaluation/scoring_operator.py
jperez999: But it is only used in this module. If its only called in your scoring operator how does putting it nearer to where it is called couple it to non-LLM scoring logic? It seems the evaluation subfolder is all tied to llm and judge evaluation.

nemo_retriever/src/nemo_retriever/evaluation/types.py
jperez999: Ok but what is the different between this and the Retriever https://github.com/NVIDIA/NeMo-Retriever/blob/1ae0ed753edde8c4df9a196902689d50852c684f/nemo_retriever/src/nemo_retriever/retriever.py#L32

nemo_retriever/src/nemo_retriever/evaluation/types.py
jperez999:
```
retrieval_results = [<results inside>]
analysis_results = []
for llm_ref in llms:
      res = analyze_results(retrieval_results, llm_ref)
      analysis_results.append(res)
```
Why wouldn't something like this work?

nemo_retriever/src/nemo_retriever/io/markdown.py
```diff
@@ -282,3 +285,132 @@ def _label_for_subtype(subtype: Any, *, fallback: str) -> str:
...
+    columns = sorted(_MARKDOWN_PARQUET_COLUMNS & available)
+    table = pf.read(columns=columns)
+    try:
+        return table.to_pandas(split_blocks=False)
+    except Exception:
+        return _arrow_table_to_pandas_via_pylist(table)
+
+
+def build_page_index(
+    parquet_dir: str | Path | None = None,
```
jperez999, line 331: I think this would break if I did something like build_page_index(dataframe). It would need to be build_page_index(None, dataframe) to work correctly, right?

nemo_retriever/src/nemo_retriever/utils/parquet_to_lancedb.py
```diff
@@ -0,0 +1,69 @@
...
+from nemo_retriever.io.dataframe import read_extraction_parquet
+from nemo_retriever.vector_store.lancedb_store import handle_lancedb
+
+logger = logging.getLogger(__name__)
+
+LANCEDB_URI = "lancedb"
+LANCEDB_TABLE = "nv-ingest"
+
+
+def reload_parquet_to_lancedb(
```
jperez999, line 30: Since this is lancedb specific it should go in the vector_store subfolder.

jperez999 (approved):
There are things we should polish in another round but this can get merged. we are going to want to move away from the eval CLI command completely. This needs to be incorporated to be usable, if activate in graph_pipeline.py

PR #1764, opened 2026-04-01
(retriever) use vLLM for nemotron-parse inference

edknv:
## Checklist


nemo_retriever/src/nemo_retriever/graph/ingestor_runtime.py
```diff
@@ -265,14 +266,22 @@ def build_graph(
...
+            extract_kwargs["extract_page_as_image"] = True
+            graph = graph >> PDFExtractionActor(**extract_kwargs)
+
             parse_kwargs: dict[str, Any] = {
                 "extract_text": extract_params.extract_text,
                 "extract_tables": extract_params.extract_tables,
                 "extract_charts": extract_params.extract_charts,
                 "extract_infographics": extract_params.extract_infographics,
             }
+            if extract_params.invoke_url:
```
jperez999: do these changes clash with the changes in your other PR about api_key resolution?

nemo_retriever/src/nemo_retriever/nim/nim.py
```diff
@@ -228,6 +236,94 @@ def _invoke_one_batch(start: int, end: int, endpoint_url: str) -> Tuple[int, int
...
+            return str(tool_calls[0]["function"]["arguments"]).strip()
+        content = choice.get("content")
+        if content:
+            return str(content).strip()
+    except (KeyError, IndexError, TypeError):
+        pass
+    return ""
+
+
+def invoke_chat_completions(
```
jperez999: are these LLM specific? can this tool be generalized to any LLM if so, maybe we should move this to another place were it can be pulled as a tool. I know Kyle is working on adding generation via LLM so he could probably leverage what you have here.

jperez999 (approved):
Let me know if those comments make sense. Otherwise looks good.

nemo_retriever/src/nemo_retriever/graph/ingestor_runtime.py
edknv: They should be okay because API keys are resolved before they get to this point.

PR #1823, opened 2026-04-08
harness: stabilize embed OOM tuning and defaults

jioffe502:
## TLDR
This PR hardens embedding stability by separating two different control knobs:
- `embed_batch_size` for Ray Data transport/scheduling, and
- `embed_inference_batch_size` for local model forward-pass VRAM pressure.

It keeps the change focused on core runtime behavior: decoupled tuning controls plus adaptive OOM retry.

## Problem
During bo767 validation, we observed long-sequence tail batches driving CUDA OOM-like failures and throughput volatility. The key discovery is that one batch knob is doing two jobs:
- scheduler/transport sizing (`embed_batch_size`), and
- model microbatch sizing (`inference_batch_size`).

That coupling makes stability tuning hard, especially when sequence lengths are heterogeneous.

### 1) Decoupled controls in harness/CLI
- Added `embed_inference_batch_size` to harness config/defaults/validation.
- Added CLI wiring so harness passes `--embed-inference-batch-size` through to runtime.
- Bound embed params to prefer `embed_inference_batch_size` over `embed_batch_size` for model forward-pass batching.

### 2) Adaptive OOM stabilization in local embed runtime
- On CUDA OOM-like errors, automatically reduce local embed microbatch and retry.
- After successful streaks, recover batch size gradually toward target.
- Emit warnings with char/token diagnostics for failing chunks.

### 3) Runtime parity + operational controls
- Added Ray object-store sizing support via `RAY_DEFAULT_OBJECT_STORE_MEMORY_PROPORTION` / `RAY_OBJECT_STORE_MEMORY_BYTES`.
- Forwarded runtime env/logging controls into Ray init path.

## Quantitative Evidence (bo767)
All runs below use `embed_inference_batch_size=32`.

| Run | gpu_embed | ingest PPS | recall@5 | Artifact |
|---|---:|---:|---:|---|
| bo767_20260406_193158_UTC | 1.0 | 152.99 | 0.8345 | `nemo_retriever/artifacts/bo767_20260406_193158_UTC/results.json` |
| bo767_20260406_202408_UTC | 1.0 | 153.56 | 0.8365 | `nemo_retriever/artifacts/bo767_20260406_202408_UTC/results.json` |
| bo767_20260406_204929_UTC | 1.0 | 151.22 | 0.8325 | `nemo_retriever/artifacts/bo767_20260406_204929_UTC/results.json` |
| bo767_20260408_154723_UTC | 1.0 | 146.45 | 0.8315 | `nemo_retriever/artifacts/bo767_20260408_154723_UTC/results.json` |
| bo767_20260408_160637_UTC | 1.0 | 146.05 | 0.8365 | `nemo_retriever/artifacts/bo767_20260408_160637_UTC/results.json` |
| bo767_20260408_172412_UTC | 1.0 | 146.53 | 0.8335 | `nemo_retriever/artifacts/bo767_20260408_172412_UTC/results.json` |
| bo767_20260408_161941_UTC | 0.25 | 160.26 | 0.8365 | `nemo_retriever/artifacts/bo767_20260408_161941_UTC/results.json` |
| bo767_20260408_173706_UTC | 0.25 | 155.54 | 0.8355 | `nemo_retriever/artifacts/bo767_20260408_173706_UTC/results.json` |

## Key Takeaways
- Decoupled microbatch control (`embed_inference_batch_size=32`) is stable and repeatable across runs.
- Recall stays in a tight band (~0.833–0.836 recall@5).
- Throughput is strongly sensitive to `gpu_embed` policy in these validations:
  - `gpu_embed=1.0` cohort: ~146 PPS
  - `gpu_embed=0.25` cohort: >150 PPS
- Adaptive OOM handling keeps runs progressing through long-tail batches instead of hard failing.

## Recommendation
- Merge this as the system-level OOM stabilization/control layer.
- Keep default `embed_inference_batch_size=32` in harness presets.
- Treat `gpu_embed` as an explicit policy knob for follow-up performance decisioning (isolation vs throughput).

## Why This Is Safe
- Additive changes around tuning and runtime behavior; no API removals.
- Retry logic activates only on OOM-like failures.
- Conservative default microbatch (`32`) aligns with observed stability.
- Harness/graph/resource tests pass.

## When To Override
- Increase `embed_inference_batch_size` only with sustained VRAM headroom and zero OOM retry pressure.
- Decrease it for long-sequence-heavy corpora or increasing OOM retry frequency.
- Tune `gpu_embed` according to target objective:
  - stability isolation: closer to `1.0`
  - throughput: evaluate lower fractions (e.g., `0.25`) with recall guardrails.

## Rollout
1. Merge with `embed_inference_batch_size=32` defaults.
2. Validate bo767/jp20 in nightly.
3. Track OOM retry frequency and recall@5 drift.
4. Finalize `gpu_embed` performance policy in a dedicated follow-up decision.

## Rollback
- Revert this commit to restore previous runtime behavior.
- Immediate mitigation without code rollback: override `embed_inference_batch_size` and/or `gpu_embed` via harness CLI.

## Open Questions For Lead Decision
- Which `gpu_embed` policy should define productized performance baselines?
- Should OOM retry events be informational telemetry only, or a gating signal for stability-required runs?

## Test Plan
- [x] `uv run pytest tests/test_harness_config.py tests/test_harness_run.py tests/test_resource_heuristics.py`
- [x] `uv run pytest tests/test_ingest_interface.py tests/test_graph_pipeline_registry.py`
- [x] `uv run pytest tests/test_create_local_embedder.py tests/test_multimodal_embed.py tests/test_operator_flags_and_cpu_actors.py`
- [x] bo767 repeated harness validations with `embed_inference_batch_size=32` (artifacts listed above)


jioffe502:
## DGX bo767 sweep evidence

Ran the dedicated 6-run `dgx_8gpu` matrix on `bo767` with:
- `embed_inference_batch_size` in `{32,64,96}`
- `gpu_embed` in `{0.25,1.0}`
- fixed `embed_batch_size=256`, `use_heuristics=false`

Observed results:

| `embed_inference_batch_size` | `gpu_embed=0.25` | `gpu_embed=1.0` | Delta vs `1.0` |
| --- | --- | --- | --- |
| `32` | `158.68 PPS`, `recall@5=0.8355` | `147.72 PPS`, `recall@5=0.8335` | `+10.96 PPS` (`+7.42%`) |
| `64` | `162.26 PPS`, `recall@5=0.8355` | `141.94 PPS`, `recall@5=0.8355` | `+20.32 PPS` (`+14.32%`) |
| `96` | `158.00 PPS`, `recall@5=0.8355` | `135.93 PPS`, `recall@5=0.8375` | `+22.07 PPS` (`+16.24%`) |

- `gpu_embed=0.25` beat `gpu_embed=1.0` at every batch size with effectively flat `recall@5`.
- Best observed throughput was `mb64,gpu_embed=0.25` at `162.26` PPS.
- `mb64,gpu=0.25` is only `2.26%` faster than `mb32,gpu=0.25`, so `mb32` remains a plausible conservative fallback.
- `mb96` should not be the default. It is slower than `mb64` under both GPU policies and still showed OOM-retry behavior in the finished session output.
- Historical `gpu_embed=1.0` controls show the same shape: `mb32=152.99`, `mb64=145.45`, `mb96=145.64` PPS.

Recommendation:
- Default `gpu_embed=0.25`.
- If OOM retries are telemetry-only, default `embed_inference_batch_size=64`.
- If we want a more conservative default, use `embed_inference_batch_size=32` and document `64` as the throughput override.
- Do not default `96`.

jioffe502:
## bo767 repeat sweep follow-up at `gpu_embed=0.25`

Ran a 40-run DGX `bo767` sweep with `gpu_embed=0.25` fixed and `embed_inference_batch_size` in `{1,32,64,256}`, repeated `10x` per setting.

| `embed_inference_batch_size` | Passes | Mean PPS | PPS stdev | PPS range | Mean `recall@5` | `recall@5` stdev | OOM evidence |
| --- | --- | ---: | ---: | ---: | ---: | ---: | --- |
| `1` | `10/10` | `152.82` | `1.38` | `151.46-156.59` | `0.8355` | `0.0006` | none |
| `32` | `10/10` | `159.34` | `1.40` | `156.94-162.11` | `0.8354` | `0.0014` | none |
| `64` | `10/10` | `158.64` | `1.67` | `156.68-163.34` | `0.8309` | `0.0108` | none |
| `256` | `9/10` | `158.51`* | `1.59`* | `156.12-161.49`* | `0.8124`* | `0.0246`* | `1` failed run, `8` OOM retries in the failed run |

`*` computed over the `9` successful `mb=256` runs only.

Takeaways:
- `32` is the best default candidate in this repeat sweep: highest mean PPS, no OOM retries, and stable recall.
- `64` is within observed run-to-run noise on ingest (`-0.70 PPS`, `-0.44%` vs `32`) and shows worse recall stability, so this does not support moving the default from `32` to `64`.
- `1` is clearly slower (`-6.52 PPS`, `-4.09%` vs `32`) with no stability upside.
- `256` is not a viable default candidate: one run failed, the failed run logged `8` `CUDA OOM during embedding; retrying` events, and the successful runs still showed materially worse recall stability.
- This is directionally consistent with the current HF/local embedding path not realizing reliable end-to-end gains from larger microbatches in this workload.
- The VDB tail is already de-bottlenecked (`VDBUploadOperator batch_size=64`), so pushing embed microbatch higher is not buying a downstream write-path win here; in practice it mainly adds memory pressure and instability.

Recommendation / next step:
- If we want one default from the current evidence, this repeat sweep supports `embed_inference_batch_size=32` at `gpu_embed=0.25`.
- From here, I think there are two reasonable paths:
  1. Land this change with `32` as the default and treat broader SKU validation as follow-up work.
  2. Keep testing in this PR, but narrow it to cross-SKU validation focused on `32` vs `64` rather than continuing to probe larger batch sizes.

jioffe502:
closing in favor of #1897

PR #1830, opened 2026-04-09
feat(install): slim install for remote/NIM-only inference on Mac/Windows

charlesbluca:
## Description

This branch streamlines the default **NeMo Retriever** install and dependency story, and tightens optional vs core packages in `nemo_retriever/pyproject.toml`.

**Dependency and install model**

- **Core** dependencies focus on orchestration (Ray), ingest API/client, HTTP stack, LanceDB, PDF/image/NIM-related libs, and shared utilities. **`nv-ingest` is no longer a direct dependency** of `nemo-retriever`; editable installs still wire `nv-ingest-api` and `nv-ingest-client` from the monorepo paths.
- New optional groups: **`[local]`** (GPU-oriented torch, transformers, Nemotron model packages, vLLM on Linux, etc.), **`[multimedia]`** (audio/ASR + SVG via `cairosvg`), **`[stores]`** (DuckDB and Neo4j), **`[benchmarks]`** (datasets / open-clip), and **`[all]`** as a convenience meta-extra.
- **Torch / torchvision** resolution: CUDA 13 wheel index on **Linux and Windows**; macOS falls back to PyPI CPU wheels. The renamed UV index (`pytorch-cu130`) matches that layout.
- **`uv.lock`** is regenerated to match the slimmer dependency graph.

**Docs**

- **`nemo_retriever/README.md`**: clearer split between **remote NIM-only** vs **local GPU** setups; documents **`[local]`** and the CUDA 13 torch override; SVG install now points at **`[multimedia]`** (with a direct `cairosvg` alternative).

**Runtime and API behavior**

- **Lazy / guarded imports** and small compatibility fixes (e.g. **Ray 2.49**) across ingest API utilities and retriever modules so optional stacks fail more gracefully when extras are not installed.
- **Text embeddings**: CPU embedding path supports **remote HTTP** when an endpoint URL is set, otherwise **local Hugging Face** on CPU; shared URL normalization avoids **doubling `/embeddings`** on HTTP embedding URLs (delegates to `nv_ingest_api` helper). Tests aligned with the local-HF default when no remote URL is configured.

**CI**

- Unit tests install **`nemo_retriever[all,dev]`** instead of separate editable `src`/`api`/`client` plus `nemo_retriever`.
- New **scheduled / dispatch** workflow **`integration-test-library-mode.yml`**: library-mode graph pipeline on **Windows** and **macOS** (ARM and Intel), using `NVCF_API_KEY` for remote NIM endpoints.

**Misc**

- Removes obsolete agent documentation; assorted small fixes (e.g. Milvus client usage, executor/ocr/pdf paths) driven by the slimmer dependency surface.

## Checklist


nemo_retriever/src/nemo_retriever/graph/executor.py
```diff
@@ -129,14 +129,22 @@ def ingest(self, data: Any, **kwargs: Any) -> Any:
...
 
+        import logging as _logging
+
+        _exec_log = _logging.getLogger(__name__)
+
         if self._show_progress and tqdm is not None:
             pbar = tqdm(operators, desc="Pipeline stages", unit="stage")
             for name, op in pbar:
                 pbar.set_postfix_str(name)
+                _exec_log.debug("[inprocess] %s: input rows=%d, cols=%s", name, len(df), list(df.columns))
```
jperez999: Are these logs necessary? I feel like they might just end up creating noise.

nemo_retriever/src/nemo_retriever/text_embed/cpu_operator.py
```diff
@@ -16,27 +16,60 @@
...
-        if not endpoint:
-            self._kwargs["embedding_endpoint"] = self.DEFAULT_EMBED_INVOKE_URL
-        self._model = None
+        if endpoint:
+            self._model = None
+            if "embedding_endpoint" not in self._kwargs and self._kwargs.get("embed_invoke_url"):
+                self._kwargs["embedding_endpoint"] = self._kwargs.get("embed_invoke_url")
+            return
+
+        from nemo_retriever.model import create_local_embedder
```
jperez999: we should not create a local embedder in the CPU operators, we should not run local models unless they are on device.

nemo_retriever/src/nemo_retriever/text_embed/cpu_operator.py
```diff
@@ -16,27 +16,60 @@
...
         return data
 
     def process(self, data: Any, **kwargs: Any) -> Any:
-        return embed_text_main_text_embed(data, model=self._model, **self._kwargs)
+        import logging as _logging
+        import pandas as pd
+
+        _log = _logging.getLogger(__name__)
+        if isinstance(data, pd.DataFrame):
+            text_col = self._kwargs.get("text_column", "text")
```
jperez999: Isn't this column supposed to be `content`?

jperez999 (changes requested):

nemo_retriever/src/nemo_retriever/text_embed/cpu_operator.py
charlesbluca: Did a quick passthrough with Cursor and seems like our general default is `text`, does this like of thinking make sense?

- **`EmbedParams.text_column`** defaults to **`"text"`**  
  [`models.py` (L226–237)](https://github.com/charlesbluca/NeMo-Retriever/blob/0988bd7852fdf74d2086e4124ca26fcb7f93721b/nemo_retriever/src/nemo_retriever/params/models.py#L226-L237)
- In **`graph_pipeline`**, we build **`EmbedParams`** from model / URL / modality / batch fields and **never pass `text_column`**, so the example graph always sticks with that default  
  [`graph_pipeline.py` (L425–441)](https://github.com/charlesbluca/NeMo-Retriever/blob/0988bd7852fdf74d2086e4124ca26fcb7f93721b/nemo_retriever/src/nemo_retriever/examples/graph_pipeline.py#L425-L441)
- For PDFs, before embed we optionally run **`explode_content_to_rows`** / **`collapse_content_to_page_rows`**. Those are wired from **`ingestor_runtime.py`** with a `partial(...)` that passes modality / content columns but **does not thread `embed_params.text_column` through**, so both helpers keep their own default **`text_column="text"`**  
  [`ingestor_runtime.py` (L370–395)](https://github.com/charlesbluca/NeMo-Retriever/blob/0988bd7852fdf74d2086e4124ca26fcb7f93721b/nemo_retriever/src/nemo_retriever/graph/ingestor_runtime.py#L370-L395)
- You can see the same default and how page vs element text is written back onto **`text_column`** in **`explode_content_to_rows`**  
  [`content_transforms.py` (L53–115)](https://github.com/charlesbluca/NeMo-Retriever/blob/0988bd7852fdf74d2086e4124ca26fcb7f93721b/nemo_retriever/src/nemo_retriever/graph/content_transforms.py#L53-L115)
- When we embed, kwargs come from **`build_embed_kwargs`**, which dumps the full **`EmbedParams`** (including **`text_column`**) into the operator  
  [`shared.py` (L12–19)](https://github.com/charlesbluca/NeMo-Retriever/blob/0988bd7852fdf74d2086e4124ca26fcb7f93721b/nemo_retriever/src/nemo_retriever/text_embed/shared.py#L12-L19)
- **`_text_from_row`** still has the usual fallbacks if the primary cell is empty  
  [`main_text_embed.py` (L128–141)](https://github.com/charlesbluca/NeMo-Retriever/blob/0988bd7852fdf74d2086e4124ca26fcb7f93721b/nemo_retriever/src/nemo_retriever/text_embed/main_text_embed.py#L128-L141)
- Separately, txt-style chunking **duplicates** the chunk into both **`text`** and **`content`** so downstream code can read either  
  [`split.py` (L234–249)](https://github.com/charlesbluca/NeMo-Retriever/blob/0988bd7852fdf74d2086e4124ca26fcb7f93721b/nemo_retriever/src/nemo_retriever/txt/split.py#L234-L249)

nemo_retriever/src/nemo_retriever/graph/executor.py
charlesbluca: Dropped in https://github.com/NVIDIA/NeMo-Retriever/pull/1830/commits/0988bd7852fdf74d2086e4124ca26fcb7f93721b

nemo_retriever/src/nemo_retriever/text_embed/cpu_operator.py
charlesbluca: Good point, updated in https://github.com/NVIDIA/NeMo-Retriever/pull/1830/commits/0988bd7852fdf74d2086e4124ca26fcb7f93721b

jperez999 (approved):

PR #1833, opened 2026-04-10
(retriever) resolve relative storage_uri to absolute path

edknv:
## Description
When `StoreParams` receives a relative `storage_uri` like `./stored_images`, the path is passed as-is to Ray workers during batch execution. Ray workers run in their own working directory, so the relative path resolves to a location inside Ray's temp directory rather than the user's project directory. The stored images end up written to an unexpected location and appear missing from the expected path.

This PR adds a model validator to `StoreParams` that resolves local relative paths to absolute at construction time, before the params are serialized and sent to Ray. URIs with a scheme (e.g. s3://) are left unchanged.

## Checklist

### DIFF
--- a/nemo_retriever/src/nemo_retriever/params/models.py
+++ b/nemo_retriever/src/nemo_retriever/params/models.py
@@ -4,12 +4,13 @@
 
 from __future__ import annotations
 
-from pathlib import Path
 from typing import Any, Literal, Optional, Sequence, Tuple
 from urllib.parse import urlparse
 
 import warnings
 
+
+from upath import UPath
 
 from nemo_retriever.tabular_data.sql_database import SQLDatabase
 from pydantic import BaseModel, ConfigDict, Field, field_validator, model_validator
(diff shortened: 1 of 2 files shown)


jperez999 (approved):

nemo_retriever/src/nemo_retriever/params/models.py
```diff
@@ -286,6 +288,13 @@ class StoreParams(_ParamsModel):
     image_format: str = "png"
     strip_base64: bool = True
 
+    @model_validator(mode="after")
+    def _resolve_local_storage_uri(self) -> "StoreParams":
+        """Resolve relative local paths to absolute so they survive Ray serialization."""
+        if not urlparse(self.storage_uri).scheme:
+            self.storage_uri = str(Path(self.storage_uri).resolve())
```
jperez999: I think this should use universalPathlib so that if the file is somewhere else we can still get it. We need to account for when people are using fsspec to save things remotely (i.e. s3).

jperez999 (changes requested):

nemo_retriever/src/nemo_retriever/params/models.py
edknv: changed to upath in https://github.com/NVIDIA/NeMo-Retriever/pull/1833/changes/c19cecb6b6669ed9ea57fc9d3914a451f2a23299

jioffe502 (approved):

PR #1839, opened 2026-04-13
Switch Bo767 and Bo10k to BEIR evaluation

ChrisJar:
## Checklist


nemo_retriever/src/nemo_retriever/harness/run.py
```diff
@@ -357,11 +276,14 @@ def _build_command(cfg: HarnessConfig, artifact_dir: Path, run_id: str) -> tuple
     ]
 
     if cfg.evaluation_mode == "beir":
+        beir_dataset_name = cfg.beir_dataset_name or cfg.dataset_label
+        if cfg.beir_loader in {"bo767_csv", "bo10k_csv"} and cfg.query_csv:
+            beir_dataset_name = str(Path(cfg.query_csv).resolve())
```
jperez999, line 281: you've changed this to be a path to a query file?

nemo_retriever/src/nemo_retriever/recall/beir.py
```diff
@@ -107,10 +275,21 @@ def build_qrels_by_query_id(
...
+    loader: str,
+    *,
+    dataset_name: str,
+    split: str = "test",
+    query_language: str | None = None,
+    doc_id_field: str = "pdf_basename",
 ) -> BeirDataset:
     """Load a BEIR-style dataset for evaluation."""
     loader_name = str(loader).strip().lower()
+    if loader_name in {"bo767_csv", "bo10k_csv"}:
```
jperez999, line 287: why is this appended with _csv? are their supposed to be different versions of these dataset?

nemo_retriever/src/nemo_retriever/vector_store/lancedb_store.py
```diff
@@ -138,6 +138,7 @@ def _build_lancedb_rows_from_df(rows: List[Dict[str, Any]]) -> List[Dict[str, An
         meta = row.get("metadata")
         if not isinstance(meta, dict):
             continue
+        meta = dict(meta)
```
jperez999: are we sure this will always be successful?

jperez999 (changes requested):
Got some questions in here. Looks good overall. Do we need all the normalization of paths in here? And is that stuff that we will only use in recall, does it have no use any where else, motivation here is moving that into a utils module?

nemo_retriever/src/nemo_retriever/harness/run.py
ChrisJar: Yeah I'm allowing it to be configurable through the query_csv parameter like it was previously, but I could also hardcode it to always use the csvs in the data/ directory

nemo_retriever/src/nemo_retriever/vector_store/lancedb_store.py
ChrisJar: I'll change this to .copy()

nemo_retriever/src/nemo_retriever/recall/beir.py
ChrisJar: The loader here is specifically for finding the bo767 or bo10k csv and then converting to beir format, that's why I'm referencing the csv specifically, but I can remove the _csv if that makes sense?

jperez999 (approved):

PR #1865, opened 2026-04-15
Add support for end to end in-memory QA evaluation

KyleZheng1284:
## Description

- E2E in-memory QA mode on graph_pipeline (d9af7e4c, 31d4e88f, 632cd365) - one command that ingests, builds the page-markdown index in-memory, queries LanceDB, and runs the full QA sweep. Closes the gap that previously forced Ingest -> Export JSON -> Eval as three separate invocations.

- Live-RAG SDK on Retriever (b04e4ab1) - the much larger commit. Turns Retriever from a raw-hits wrapper into a first-class RAG SDK with .retrieve(), .retrieve_batch(), .answer() (single-query Tier 1+2+3 scoring), and a fluent .pipeline().generate(...).score().judge(...).run(queries) batch builder. Plus a ground-up nemo_retriever/llm/ package (shared types.py, Pydantic LLMInferenceParams + new LLMRemoteClientParams, clients/litellm.py, clients/judge.py, lazy-loaded __init__.py) that the eval framework was refactored to import from.

Built off of this existing PR https://github.com/NVIDIA/NeMo-Retriever/pull/1754

## Checklist


nemo_retriever/src/nemo_retriever/evaluation/config.py
```diff
@@ -0,0 +1,387 @@
...
+                sorted(distinct_judges),
+                first_judge_key,
+            )
+        config.setdefault("judge", models[first_judge_key])
+
+    elif "generators" in config and "judge" in config:
+        generators = config["generators"]
+        judge_cfg = config["judge"]
+        judge_name = judge_cfg.get("name", "judge")
+        config.setdefault("models", {})
```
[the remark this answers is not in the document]
jperez999: This seems like a valid issue, or not?

nemo_retriever/src/nemo_retriever/evaluation/retrievers.py
```diff
@@ -86,6 +89,97 @@ def __init__(self, file_path: str):
...
+        instance.file_path = "<in-memory>"
+        instance._norm_index = {}
+        instance._raw_keys = {}
+        instance._miss_count = 0
+        instance._miss_lock = threading.Lock()
+        for raw_key, value in queries.items():
+            norm = _normalize_query(raw_key)
+            instance._norm_index[norm] = value
+            instance._raw_keys[norm] = raw_key
+        return instance
```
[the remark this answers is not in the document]
jperez999: This seems like something you should add. That divergence will cause issues later down the line.

nemo_retriever/src/nemo_retriever/retriever.py
```diff
@@ -342,19 +372,518 @@ def queries(
...
-            results = self._rerank_results(query_texts, results)
+            results = self._rerank_results(query_texts, results, top_k=effective_top_k)
+
+        return results
+
+    # ------------------------------------------------------------------
+    # Live RAG API (structured retrieval + generation)
+    # ------------------------------------------------------------------
+
+    def retrieve(
```
jperez999, line 387: what is the difference between what you get from retriever and what you get from query/queries?

nemo_retriever/src/nemo_retriever/retriever.py
```diff
@@ -342,19 +372,518 @@ def queries(
...
+        )
+
+        results: list[RetrievalResult] = []
+        for hits in hits_per_query:
+            chunks = [str(hit.get("text", "")) for hit in hits]
+            metadata = [{k: v for k, v in hit.items() if k != "text"} for hit in hits]
+            results.append(RetrievalResult(chunks=chunks, metadata=metadata))
         return results
 
+    def answer(
```
jperez999, line 489: This should all fall within the query/queries APIs. When I have a retriever graph query or queries runs that full retriever graph. When I call query, it should execute the entire graph for that query. Answer(llm), score(f1), retrieve(lancedb) are all individual operators within the graph. All you should have to do here is change the graph that is executed. The user facing API should not change. Or is there something I overlooked that makes these API calls necessary?

nemo_retriever/src/nemo_retriever/retriever.py
```diff
@@ -342,19 +372,518 @@ def queries(
...
+                df["reference_answer"] = refs
+        elif isinstance(queries, list):
+            df = pd.DataFrame({"query": list(queries)})
+            if reference is not None:
+                refs = reference if isinstance(reference, list) else [reference] * len(queries)
+                if len(refs) != len(queries):
+                    raise ValueError("reference length must match queries length")
+                df["reference_answer"] = refs
+        elif isinstance(queries, pd.DataFrame):
+            df = queries.copy()
```
jperez999, line 865: why do you need to copy the df here?

nemo_retriever/src/nemo_retriever/retriever.py
```diff
@@ -342,19 +372,518 @@ def queries(
...
+            >>> from nemo_retriever.llm import LiteLLMClient, LLMJudge
+            >>> retriever = Retriever(lancedb_uri="./kb")
+            >>> llm = LiteLLMClient.from_kwargs(model="nvidia_nim/meta/llama-3.3-70b-instruct")
+            >>> judge = LLMJudge.from_kwargs(model="nvidia_nim/mistralai/mixtral-8x22b-instruct-v0.1")
+            >>> df = (  # doctest: +SKIP
+            ...     retriever.pipeline()
+            ...     .generate(llm)
+            ...     .score()
+            ...     .judge(judge)
+            ...     .run(queries=["What is RAG?"], reference=["Retrieval-augmented generation..."])
```
jperez999, line 687: I think having a simple graph param in the Retriever object that consumed the graph built by RetrieverPipelineBuilder would be sufficient. These commands here that actually execute actions I dont think belong here.

jperez999 (changes requested):
I dont

nemo_retriever/src/nemo_retriever/retriever.py
KyleZheng1284: retrieve() will reshape .query()'s raw LanceDB hits into RetrievalResult which is the return type required by the RetrieverStrategy Protocol so Retriever and FileRetriever will become interchangeable at each Protocol call site

nemo_retriever/src/nemo_retriever/retriever.py
KyleZheng1284: apis exist so that single query can be a fast path (no graph or ray setup cost) used by the cli and mcp. I don't think the pipeline would be the right tool for this use case

nemo_retriever/src/nemo_retriever/retriever.py
KyleZheng1284: LiveRetrievalOperator and QAGenerationOperator will mutate the DataFrame in place (adding context, answer, etc.), so the copy prevents those side-effects from leaking back to the caller's df

nemo_retriever/src/nemo_retriever/retriever.py
KyleZheng1284: pipeline() returns a builder that wraps `LiveRetrievalOperator/QAGenerationOperator/ScoringOperator/JudgingOperator` under the hood, so that folks using this get a short fluent option instead of the four imports and >> syntax. The raw-graph form should still work identically so I'd be happy to split RetrieverPipelineBuilder into evaluation/builder.py if you'd rather it live outside retriever.py.

jperez999 (approved):

PR #1877, opened 2026-04-20
Async ray executor

jdye64:
Please review and merge https://github.com/NVIDIA/NeMo-Retriever/pull/1860 first

## Checklist


nemo_retriever/src/nemo_retriever/graph/abstract_operator.py
```diff
@@ -35,9 +72,42 @@ def run(self, data: Any, **kwargs: Any) -> Any:
...
+        operator executes synchronously via :meth:`run` so that existing
+        ``result = op(data)`` call-sites keep working.  Inside an async
+        context the method returns the :meth:`arun` coroutine which the
+        caller (e.g. Ray Data) must ``await``.
+        """
+        try:
+            asyncio.get_running_loop()
+        except RuntimeError:
+            return self.run(data, **kwargs)
+        return self.arun(data, **kwargs)
```
charlesbluca: Does it make sense to be returning the un-awaited coroutine here even though this function isn't async def'd?

nemo_retriever/src/nemo_retriever/graph/abstract_operator.py
```diff
@@ -35,9 +72,42 @@ def run(self, data: Any, **kwargs: Any) -> Any:
...
 
+    async def aprocess(self, data: Any, **kwargs: Any) -> Any:
+        """Async version of :meth:`process`.
+
+        The default calls ``process()`` synchronously on the event-loop
+        thread.  This is intentional: most operators use C extensions
+        (pypdfium2, OpenCV, torch, …) that are **not** thread-safe, so
+        ``asyncio.to_thread`` would allow Ray Data to run multiple
+        batches in parallel threads inside the same actor, causing
+        memory corruption.
```
charlesbluca, line 66: Is this true? Claude flagged that we're using `asyncio.to_thread` in our local GPU page-elements implementation, which implies this memory corruption claim is dubious

nemo_retriever/src/nemo_retriever/text_embed/operators.py
```diff
@@ -33,27 +33,35 @@ def prefers_cpu_variant(cls, operator_kwargs: dict[str, Any] | None = None) -> b
...
+        from nemo_retriever.text_embed.gpu_operator import BatchEmbedGPUActor
+
+        return BatchEmbedGPUActor
     if name == "_BatchEmbedCPUActor":
-        from nemo_retriever.text_embed.cpu_operator import _BatchEmbedCPUActor
+        from nemo_retriever.text_embed.cpu_operator import BatchEmbedCPUActor
 
-        return _BatchEmbedCPUActor
+        return BatchEmbedCPUActor
     if name == "_BatchEmbedGPUActor":
```
charlesbluca, line 65: Do we also need to add `_BatchEmbedActor` to this compatibility shim?

nemo_retriever/tests/test_asr_actor.py
```diff
@@ -22,6 +23,13 @@
 from nemo_retriever.params import ASRParams
 
 
+def _run(coro_or_result):
+    """Run a coroutine synchronously in tests; pass through plain values."""
+    if not asyncio.iscoroutine(coro_or_result):
+        return coro_or_result
+    return asyncio.new_event_loop().run_until_complete(coro_or_result)
```
charlesbluca: Looks like this event loop is leaking? We should explicitly close after the `run_until_complete`

charlesbluca:
Should we also apply these changes to `NemotronParseCPUActor`?

nemo_retriever/src/nemo_retriever/graph/abstract_operator.py
```diff
@@ -12,6 +13,42 @@
...
+
+    uvloop >= 0.22's ``get_event_loop()`` raises ``RuntimeError`` when no loop
+    is *running*, which breaks Ray Data's async-actor initialisation in freshly
+    spawned worker processes.  We try to create a loop under the *current*
+    policy first so that uvloop (or any other custom policy) is preserved.
+    The default-policy fallback only triggers if the installed policy itself
+    cannot create a usable loop.
+    """
+    try:
+        loop = asyncio.get_event_loop_policy().get_event_loop()
```
charlesbluca: This emits a `DeprecationWarning` on 3.12; could we switch over to `asyncio.get_running_loop()` here with something like

```python
  def _ensure_event_loop() -> None:
      try:
          asyncio.get_running_loop()
          return  # already inside a running loop
      except RuntimeError:
          pass
      asyncio.set_event_loop(asyncio.new_event_loop())
```

nemo_retriever/src/nemo_retriever/recall/core.py
```diff
@@ -522,7 +522,7 @@ def retrieve_and_score(
         lancedb_uri=cfg.lancedb_uri,
         lancedb_table=cfg.lancedb_table,
         embedder=cfg.embedding_model or VL_EMBED_MODEL,
-        embedding_http_endpoint=cfg.embedding_http_endpoint,
+        embedding_http_endpoint=endpoint,
```
jperez999: this should come from the config like everything else right?

nemo_retriever/src/nemo_retriever/retriever.py
```diff
@@ -94,6 +94,16 @@ class Retriever:
     # Internal cache for local HF embedders, keyed by model name.
     _embedder_cache: dict = field(default_factory=dict, init=False, repr=False, compare=False)
 
+    def __str__(self) -> str:
```
jperez999, line 97: where and how is this used?

nemo_retriever/tests/test_asr_actor.py
jperez999: Can we pull this function out into a utility that all the files that need it can use? I see it in the tests multiple times.

jperez999 (approved):

nemo_retriever/src/nemo_retriever/retriever.py
jdye64: Just an overloaded Python __str__ function to get the string representation of the object itself. I use it in other things external to this codebase itself

nemo_retriever/src/nemo_retriever/graph/abstract_operator.py
jdye64: Do we know for sure `asyncio.to_thread` is not causing those types of memory issues in the GPU page-elements implementation?

jdye64:
This PR was originally opened to support "online mode". The strategy has since changed and this is just an un-necessary complication to the codebase. I'm going to close it for now and we can re-open if things change again.

PR #1897, opened 2026-04-21
Set default embed batch size to 32

jioffe502:
## Summary
This replaces the larger embed batching/decoupling direction with a minimal coupled-default change.

Changes:
- set `BatchTuningParams.embed_batch_size` default from `256` to `32`
- set harness `embed_batch_size` default from `256` to `32`

This keeps the product surface coupled while moving the default away from the larger embed batch regime.

## Why
Recent bo767 simple-run benchmarking did not support keeping the more complex decoupling work as a throughput win.

Key learnings:
- `64/64` was the clearly bad regime and produced embed OOMs
- `64/32` did not outperform `32/32` in the simple runner
- on the simplest apples-to-apples `32/32` baseline, PR 1823 was near parity with `upstream/main`, not a clear speedup

Given that, the smallest defensible change is to keep coupling and lower the default.

## Benchmark Notes
bo767 simple `graph_pipeline <dataset> --embed-batch-size 32`, 3 runs each:

- `upstream_main_32_32` median ingest PPS: `126.775`
- `pr1823_32_32` median ingest PPS: `125.584`

This supported a simpler default-tuning PR instead of retaining the extra decoupling control surface.

## Validation
Focused harness tests on this branch:
- `test_harness_config.py`
- `test_harness_run.py`

Result:
- the branch-specific default change itself behaved as expected
- two `test_harness_run.py` failures are already present on clean `upstream/main` and are unrelated to this change


jperez999 (approved):

jdye64 (approved):
yep, 1000% needed

PR #1911, opened 2026-04-22
vdb: Add legacy ADT backend path

jioffe502:
Adds a generic graph ingestion path for legacy NV-Ingest VDB ADT operators while keeping native LanceDB as the default. Milvus is used as an external backend example rather than a core Retriever dependency.

- Adds a legacy VDB adapter that converts Retriever rows into the legacy list-of-records contract and routes explicit `--vdb-op` operators through `VDB.run(records)`.
- Extends `graph_pipeline` with generic VDB operator loading/kwargs and backend-aware run summaries while preserving default LanceDB recall behavior.
- Adds unit coverage for the adapter, graph pipeline routing, and a Milvus contract integration test.

### DIFF
--- a/nemo_retriever/src/nemo_retriever/examples/graph_pipeline.py
+++ b/nemo_retriever/src/nemo_retriever/examples/graph_pipeline.py
@@ -35,12 +35,13 @@ Examples::
 
 from __future__ import annotations
 
+import importlib
 import json
 import logging
 import os
 import time
 from pathlib import Path
-from typing import Optional, TextIO
+from typing import Any, Optional, TextIO
 
 import typer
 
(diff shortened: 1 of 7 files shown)


nemo_retriever/src/nemo_retriever/examples/graph_pipeline.py
```diff
@@ -617,6 +716,84 @@ def main(
...
+                vdb_operator,
+                vdb_target or "<unspecified>",
+                vdb_summary.get("records"),
+            )
+            logger.warning(
+                "Skipping LanceDB recall/BEIR evaluation because --vdb-op selected a legacy VDB ADT operator. "
+                "Backend-neutral evaluation needs a retrieval result contract for custom VDB operators."
+            )
+
+            _write_runtime_summary(
```
jperez999, line 749: did you create a separate write_runtime_summary for vdb op usage? Why?

nemo_retriever/src/nemo_retriever/vector_store/legacy_vdb_adapter.py
```diff
@@ -0,0 +1,323 @@
...
+        return embedding
+
+    payload = _maybe_parse_json(_row_get(row, embedding_column))
+    if isinstance(payload, Mapping):
+        return _as_list(payload.get(embedding_key))
+    return _as_list(payload)
+
+
+def _extract_content(row: Mapping[str, Any], metadata: Mapping[str, Any], text_column: str) -> str:
+    content = _row_get(row, text_column)
```
jperez999, line 130: This _row_get function is weird. In what scenario do you expect this to be used? Why cant the actual dict get call work here? This is going to a blank object() as default then checking if it is missing and then going to the default. The logic/motivation is not clear.

nemo_retriever/src/nemo_retriever/vector_store/legacy_vdb_adapter.py
```diff
@@ -0,0 +1,323 @@
...
+    source_metadata = _build_source_metadata(row, metadata)
+    document_type = _infer_document_type(row, metadata, content_metadata)
+
+    record_metadata = dict(metadata)
+    record_metadata["content"] = content
+    record_metadata["embedding"] = embedding
+    record_metadata["content_metadata"] = content_metadata
+    record_metadata["source_metadata"] = source_metadata
+
+    if document_type == "structured":
```
jperez999, line 230: this is very drawn out. Cant we just check for specific metadata and then just use that as the metadata. Why do we need to anotate the record_metadata with the image/audio/table_metadata and not just generic metadata? What problem does this solve?

nemo_retriever/src/nemo_retriever/vector_store/legacy_vdb_adapter.py
```diff
@@ -0,0 +1,323 @@
...
+    if hasattr(df, "to_dict") and hasattr(df, "columns"):
+        rows = df.to_dict(orient="records")
+    elif isinstance(df, list):
+        rows = df
+    else:
+        raise TypeError(f"Expected a pandas DataFrame or list of dict rows, got {type(df).__name__}.")
+
+    records: List[dict[str, Any]] = []
+    for row in rows:
+        if not isinstance(row, Mapping):
```
jperez999, line 271: In what scenario would this be true?

nemo_retriever/src/nemo_retriever/vector_store/legacy_vdb_adapter.py
```diff
@@ -0,0 +1,323 @@
...
+
+def execution_result_to_legacy_vdb_records(
+    result: Any,
+    *,
+    embedding_column: str = "text_embeddings_1b_v2",
+    embedding_key: str = "embedding",
+    text_column: str = "text",
+) -> list[list[dict[str, Any]]]:
+    """Convert a GraphIngestor execution result into one legacy VDB record batch."""
+    if hasattr(result, "to_dict") and hasattr(result, "columns"):
```
jperez999, line 292: shouldnt this be isinstance pd.Dataframe it is much clearer here if that is the case, right? Otherwise, what else has both these attrs and is an expected return?

nemo_retriever/src/nemo_retriever/vector_store/legacy_vdb_adapter.py
```diff
@@ -0,0 +1,323 @@
...
+    content = _row_get(row, text_column)
+    if _is_missing(content):
+        content = metadata.get("content")
+    if _is_missing(content):
+        content = _row_get(row, "content", "")
+    return "" if _is_missing(content) else str(content)
+
+
+def _build_content_metadata(row: Mapping[str, Any], metadata: Mapping[str, Any]) -> dict[str, Any]:
+    content_metadata = _merge_dicts(_row_get(row, "content_metadata"), metadata.get("content_metadata"))
```
jperez999, line 139: in what scenario does this update actually trigger?

jperez999 (changes requested):

PR #1920, opened 2026-04-23
Extension of Retriever CLI for generation subcommands + MCP

KyleZheng1284:
## Description
- Confines Retriever to query-time operations (.query / .queries); removes answer generation, scoring, judging, and pipeline-building from the Retriever object.
- Adds nemo_retriever.generation: five composable pd.DataFrame-returning functions (retrieve, answer, score, judge, eval) with additive column semantics.
- Exposes three of them as CLI subcommands: retriever retrieve, retriever answer, retriever eval batch.
- Adds retriever mcp serve — an optional stdio MCP server around generation.answer / generation.eval behind the [mcp] extra.

##Key design points
- Retriever is now a Protocol-conformant query engine, decoupled from LLM + scoring concerns.
- Heavy deps (pandas, litellm, lancedb, mcp) are lazy-imported inside command bodies, so retriever --help and per-subcommand --help don't pay the cost.
retriever answer --json preserves the historical AnswerResult JSON schema (incl. numpy-scalar unboxing) so existing agents / CI consumers don't break.
- Schema'd exit codes on retriever answer: 0 ok, 2 below judge threshold, 3 empty retrieval, 4 usage error, 5 generation failed.
- retriever eval batch registers as a peer of the pre-existing retriever eval run / export / build-page-index — no namespace collision.

## Checklist


nemo_retriever/src/nemo_retriever/retriever.py
```diff
@@ -380,511 +379,12 @@ def queries(
```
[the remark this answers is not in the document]
KyleZheng1284: already ported this logic over to `nemo_retriever.generation.* '

nemo_retriever/src/nemo_retriever/llm/types.py
```diff
@@ -7,21 +7,44 @@
...
 class RetrieverStrategy(Protocol):
-    """Pluggable retrieval strategy interface."""
+    """Pluggable retrieval strategy interface.
+
+    Implementations return a :class:`pandas.DataFrame` with exactly the
+    columns ``[query, chunks, metadata]`` and one row per invocation.
+    ``chunks`` is a ``list[str]`` in rank order; ``metadata`` is the
+    aligned ``list[dict]`` holding everything else a hit record carries
+    (source, page_number, distance scores, ...).
+
```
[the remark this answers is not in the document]
KyleZheng1284: moving towards the pd.DataFrame as the universal

nemo_retriever/src/nemo_retriever/adapters/cli/main.py
```diff
@@ -38,6 +41,32 @@
...
+app.command(
+    name="retrieve",
+    help="Batch retrieval: emit one JSONL row per query with (query, chunks, metadata).",
+)(retrieve_command)
+
+# ``retriever eval batch`` -- see nemo_retriever.generation.cli.  Registered
+# on the existing ``eval_app`` group as a peer of ``run`` / ``export`` /
+# ``build-page-index`` so both YAML-driven and batch-JSONL evaluation
+# surfaces live under the same subcommand namespace.
+eval_app.command(
```
jperez999, line 56: is the eval_app separate from the main `retriever` app? If so, please unite them, just use `eval` as the sub command.

nemo_retriever/src/nemo_retriever/evaluation/live_retrieval.py
```diff
@@ -76,17 +75,20 @@ def process(self, data: Any, **kwargs: Any) -> Any:
         # One batched call instead of per-row iteration.  The Retriever
         # embeds all queries in a single NIM round trip and issues a
         # single LanceDB sweep, so an N-row DataFrame pays O(1) network
-        # cost end-to-end rather than O(N).  Order is preserved by
-        # ``retrieve_batch`` so ``results[i]`` aligns with row ``i``.
-        results = self._retriever.retrieve_batch(query_texts, top_k=self._top_k)
+        # cost end-to-end rather than O(N).  ``Retriever.queries``
+        # preserves input order so ``hits_per_query[i]`` aligns with
+        # row ``i``.
+        hits_per_query = self._retriever.queries(query_texts, top_k=self._top_k)
```
jperez999, line 102: appreciate you moving back to queries.

nemo_retriever/src/nemo_retriever/evaluation/orchestrator.py
```diff
@@ -266,12 +266,23 @@ def _prepare_dataframe(self, qa_pairs: list[dict]) -> pd.DataFrame:
         def _retrieve(idx: int, pair: dict) -> tuple[int, dict]:
             query = pair["query"]
             reference = pair.get("reference_answer") or pair.get("answer", "")
-            retrieval = self.retriever.retrieve(query, self.top_k)
+            hit_df = self.retriever.retrieve(query, self.top_k)
```
jperez999, line 316: Can you explain why this is still retrieve instead of queries?

nemo_retriever/src/nemo_retriever/evaluation/README.md
```diff
@@ -39,7 +39,7 @@ End-to-end bo767 + LanceDB + full-page markdown touches these **artifacts** and
 
 | Stage | Artifacts produced | Code / APIs involved |
 |-------|-------------------|----------------------|
-| **1. Ingest + embed** | `lancedb/<uri>/<table>/` (embedded sub-page chunks); optionally `data/bo767_extracted/*.parquet` | `python -m nemo_retriever.examples.graph_pipeline` (extract, embed, VDB upload to LanceDB via the operator graph). Add `--save-intermediate` to also save extraction Parquet for full-page markdown (recommended for best results). **Table name must match** `retriever eval export` (`--lancedb-table`, default `nv-ingest`). |
+| **1. Ingest + embed** | `lancedb/<uri>/<table>/` (embedded sub-page chunks); optionally `data/bo767_extracted/*.parquet` | `retriever pipeline run` (extract, embed, VDB upload to LanceDB via the operator graph). Add `--save-intermediate` to also save extraction Parquet for full-page markdown (recommended for best results). **Table name must match** `retriever eval export` (`--lancedb-table`, default `nv-ingest`). |
```
jperez999, line 42: appreciate this fix

nemo_retriever/src/nemo_retriever/evaluation/README.md
```diff
@@ -241,6 +241,8 @@ uv pip install -e "./nemo_retriever[llm]"
 
 The `[llm]` extra installs `litellm` for LLM generation and judging (and powers both the batch-eval framework and the live-RAG SDK). If you are not using this tree, you can instead `uv pip install "nemo-retriever[llm]"` from PyPI (package name uses a hyphen).
 
+> **Migration note:** the previously-provided `nemo-retriever[eval]` install extra was removed in favor of `nemo-retriever[llm]`. Pin the new extra in any requirements files that still reference `[eval]`.
```
jperez999: remove this, there hasn't been a release between when you introduced this and the change. So no need for this note.

nemo_retriever/src/nemo_retriever/evaluation/README.md
```diff
@@ -1065,14 +1067,24 @@ The same pattern works for custom failure classifiers, alternative judge prompts
...
 
 ```python
-from nemo_retriever.llm.types import RetrieverStrategy, RetrievalResult
+import pandas as pd
+
+from nemo_retriever.llm.types import RetrieverStrategy
 
 class MyRetriever:
-    def retrieve(self, query: str, top_k: int) -> RetrievalResult:
+    def retrieve(self, query: str, top_k: int) -> pd.DataFrame:
```
jperez999, line 1073: So I think we do not need this. This should be taken care of via the operators and setting the execution graph for the Retriever with those operators. thoughts?

nemo_retriever/src/nemo_retriever/evaluation/retrieval_loader.py
```diff
@@ -90,12 +90,17 @@ def process(self, data: Any, **kwargs: Any) -> Any:
             reference = pair.get(self._answer_column, "")
             if not query:
                 continue
-            result = retriever.retrieve(query, self._top_k)
+            hit_df = retriever.retrieve(query, self._top_k)
```
jperez999: lets refactor this to use queries that uses the graph in the retriever.

nemo_retriever/src/nemo_retriever/evaluation/retrievers.py
```diff
@@ -22,8 +22,10 @@
 import re
 import threading
 import unicodedata
+from typing import TYPE_CHECKING
```
jperez999: why is this an if statement? Pandas is guaranteed to be in the environment it is a requirement...

nemo_retriever/src/nemo_retriever/evaluation/retrievers.py
```diff
@@ -223,8 +225,20 @@ def check_coverage(self, qa_pairs: list[dict]) -> float:
...
+        :func:`nemo_retriever.generation.retrieve` -- so that the QA eval
+        harness can consume both live and cached retrieval through the
+        same DataFrame contract.  Missing queries are represented as a
+        row with empty ``chunks`` / ``metadata`` lists rather than a
+        raised exception, mirroring the previous "graceful miss"
+        behaviour that counted misses on ``_miss_count``.
+        """
+        import pandas as pd
+
         norm = _normalize_query(query)
```
jperez999: The input here is a string, why are you normalizing, which is really changing the input the user provided?

nemo_retriever/src/nemo_retriever/generation/cli.py
```diff
@@ -0,0 +1,349 @@
...
+        for _, row in df.iterrows():
+            stream.write(json.dumps(row.to_dict(), default=str, ensure_ascii=False) + "\n")
+        if hasattr(stream, "flush"):
+            stream.flush()
+    finally:
+        if not to_stdout:
+            stream.close()
+
+
+def _build_retriever(
```
jperez999, line 116: why not inline this to where it is used. The only thing you do here is decide the embedder which could be done inline, right?

nemo_retriever/src/nemo_retriever/generation/cli.py
```diff
@@ -0,0 +1,349 @@
...
+        None,
+        "--embedding-endpoint",
+        help="HTTP endpoint for a remote embedding NIM.",
+    ),
+    embedding_api_key: Optional[str] = typer.Option(
+        None,
+        "--embedding-api-key",
+        envvar="NVIDIA_API_KEY",
+        help="Bearer token for the embedding endpoint.",
+    ),
```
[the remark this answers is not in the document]
jperez999: If this is true, there is not reason to have two of these keys, do we?

nemo_retriever/src/nemo_retriever/generation/retrieve.py
```diff
@@ -0,0 +1,90 @@
...
+        query_list,
+        top_k=top_k,
+        embedder=embedder,
+        lancedb_uri=lancedb_uri,
+        lancedb_table=lancedb_table,
+    )
+
+    rows: list[dict[str, Any]] = []
+    for q, hits in zip(query_list, hits_per_query):
+        chunks = [str(hit.get("text", "")) for hit in hits]
```
jperez999: In what scenario do you get here and text does not exist?

nemo_retriever/src/nemo_retriever/generation/score.py
```diff
@@ -0,0 +1,51 @@
...
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    import pandas as pd
+
+
+def score(df: "pd.DataFrame") -> "pd.DataFrame":
```
jperez999: this is clean!

nemo_retriever/src/nemo_retriever/answer_cli.py
```diff
@@ -0,0 +1,407 @@
...
+        df = eval_fn(
+            retriever,
+            [question],
+            llm=llm,
+            reference=reference,
+            judge=judge,
+            top_k=top_k,
+        )
+    else:
+        df = answer_fn(
```
jperez999, line 380: This should be a graph construction than an execution. It does not need to be in a separate function. If it does other stuff you should flatten it out here.

nemo_retriever/src/nemo_retriever/answer_cli.py
```diff
@@ -0,0 +1,407 @@
...
+    # exit 5 so callers are pointed at the knowledge-base / ingestion
+    # problem rather than the symptom.
+    if not chunks:
+        raise typer.Exit(code=EXIT_EMPTY_RETRIEVAL)
+    if gen_error is not None:
+        raise typer.Exit(code=EXIT_GENERATION_FAILED)
+    if min_judge_score is not None and (judge_score is None or judge_score < min_judge_score):
+        raise typer.Exit(code=EXIT_BELOW_JUDGE_THRESHOLD)
+
+    raise typer.Exit(code=EXIT_OK)
```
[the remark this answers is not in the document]
jperez999: Check validity here.

nemo_retriever/src/nemo_retriever/mcp_server.py
```diff
@@ -0,0 +1,320 @@
...
+            "Error: the `mcp` Python SDK is required for `retriever mcp serve`.\n"
+            "Install it with:\n"
+            "    pip install nemo-retriever[mcp]",
+            err=True,
+        )
+        raise typer.Exit(code=1) from exc
+    return mcp.server, mcp.server.stdio, mcp.types
+
+
+def _build_retriever_and_llm(
```
jperez999: again this could be done inline. what does the function give you that inline doesn't? I feel in-line is better because it is easier to follow the code and no need to jump around.

nemo_retriever/src/nemo_retriever/mcp_server.py
```diff
@@ -0,0 +1,320 @@
...
+                    retriever,
+                    [question],
+                    llm=llm,
+                    top_k=call_top_k,
+                )
+            return row_to_answer_dict(df.iloc[0])
+
+        # Offload the synchronous generation call to a thread so we don't
+        # block the event loop that's driving stdio I/O.
+        loop = asyncio.get_running_loop()
```
jperez999, line 266: Do we need to worry about updating? If I use this for a big job, do I get status updates?

jperez999 (changes requested):

nemo_retriever/src/nemo_retriever/adapters/cli/main.py
KyleZheng1284: no its not separate eval_app is mounted as the eval subgroup under the main retriever app, the comment may be slightly misleading but its just like audio_app or pdf_app

nemo_retriever/src/nemo_retriever/evaluation/README.md
KyleZheng1284: Changing it to be handled by operators is definitely possible but the `RetrieverStrategy` Protocol is the public contract so that for non-LanceDB retrievers (i.e. agentic retrieval) as long as its any objects with a single row DF it can be plugged in. IF we were to switch Im worried that might be more of a lift on the side of the user.

nemo_retriever/src/nemo_retriever/answer_cli.py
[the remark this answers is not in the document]
KyleZheng1284: judge timeouts were exiting 2 ("below threshold"). Added EXIT_JUDGE_FAILED = 6 + tests so agents can retry the judge without regenerating.

nemo_retriever/src/nemo_retriever/evaluation/retrievers.py
KyleZheng1284: yep seems it was hard pinned and already imported on every cli path, removing this and just doing top level import

nemo_retriever/src/nemo_retriever/evaluation/retrievers.py
KyleZheng1284: this is so that different retrieval JSONs from different tools match. nothing the user passes in will get mutated

nemo_retriever/src/nemo_retriever/mcp_server.py
KyleZheng1284: would like to defer to a later PR

nemo_retriever/src/nemo_retriever/mcp_server.py
KyleZheng1284: agreed went ahead and inlined the body into `serve_command` and deleted the wrapper.

nemo_retriever/src/nemo_retriever/generation/retrieve.py
KyleZheng1284: actually it should be none, switched to strict hit["text"] here

nemo_retriever/src/nemo_retriever/answer_cli.py
KyleZheng1284: Kept the if reference is not None: eval_fn(...) else: answer_fn(...) dispatch and just trimmed the heavy comments around it.

`generation.answer` and `generation.eval` are separate public APIs where eval requires a reference to score while answer doesn't, the MCP also uses this same shape collapsing it might just cause a bigger drift

nemo_retriever/src/nemo_retriever/generation/cli.py
KyleZheng1284: `_build_retriever` is already shared by both `retrieve_command` and `eval_batch_command`, so inlining it would duplicate the 12-arg Retriever(...) setup in two places

nemo_retriever/src/nemo_retriever/evaluation/retrieval_loader.py
```diff
@@ -84,20 +84,25 @@ def process(self, data: Any, **kwargs: Any) -> Any:
...
+            kept_queries.append(query)
+            kept_references.append(pair.get(self._answer_column, ""))
+
+        # FileRetriever.retrieve_many returns one row per input query in
+        # input order with the canonical [query, chunks, metadata]
+        # columns; misses degrade to empty chunks lists, so we can zip
+        # positionally against kept_queries/kept_references without
+        # guarding against empty frames.
+        hit_df = retriever.retrieve_many(kept_queries, self._top_k)
+        rows = [
```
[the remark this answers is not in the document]
KyleZheng1284: are you sure there is no path by which a "custom retriever satisfying the updated protocol" lands here.
Compare to QAEvalPipeline.__init__ (orchestrator) which does take retriever: RetrieverStrategy as a constructor arg — that's why the orchestrator needs the hasattr gate, and why the loader doesn't.

PR #1944, opened 2026-04-30
Feat/retriever service mode

jdye64:
## Checklist


charlesbluca (commented):
Validation note: I ran the service-mode validation script against `http://10.86.5.28:7670` using this PR's client code and `data/multimodal_test.pdf`. Health/capabilities passed, ingest completed 3/3 PDF pages with 8 output rows and 0 failures in 3.59s, query returned 5 hits in 0.15s, and job inspection showed 1 complete / 0 failed. Rerank was skipped because `/v1/capabilities` reports rerank disabled on that instance. Also, the shared validation script currently has a Python syntax issue (`global SERVICE_URL` after use inside `main()`); moving the global declaration to the top of `main()` fixed the script locally.

nemo_retriever/src/nemo_retriever/service/client.py
```diff
@@ -0,0 +1,1470 @@
...
+
+    @staticmethod
+    def _prepare_file(file_path: Path) -> list[tuple[bytes, int, int]]:
+        """Return ``(page_bytes, page_number, total_pages)`` tuples.
+
+        PDFs are split into individual pages; non-PDFs become a single page.
+        """
+        raw = file_path.read_bytes()
+        if file_path.suffix.lower() == ".pdf":
+            pages = _split_pdf(raw)
```
charlesbluca: Question from validation: `/v1/capabilities` reports `server_pdf_split: true`, but this client path still splits PDFs locally and uploads one single-page PDF per request. Is that intentional for service mode? If server-side splitting is the intended contract, this keeps `pypdfium2` and several PDF import dependencies on the client and makes local split/import failures happen before the service ever registers a job.

nemo_retriever/src/nemo_retriever/service/client.py
```diff
@@ -0,0 +1,1470 @@
...
+                does not silently truncate the result stream.
+
+                If ``pending_jobs`` hasn't shrunk for 120s, polls the REST
+                API to detect silently-completed jobs and removes them.
+                """
+                _STALE_TIMEOUT = 120.0
+
+                await splits_done.wait()
+                if not job_ids:
+                    return
```
charlesbluca: I hit a client-side hang here when `_prepare_file()` raised before any `job_id` was appended. `_split_files()` only sets `splits_done` in that case; it does not enqueue the upload sentinel or a terminal/error item for `result_queue`. Then `_stream_consumer()` returns on `not job_ids` before its later `finally` can put `None`, so `aingest_documents_stream()` waits forever on `result_queue.get()` and the service shows zero jobs. Can we catch split-time exceptions and emit a `stream_error`/terminal sentinel, or make this no-job path always unblock the result queue?

jperez999 (approved):

jperez999 (approved):

PR #1996, opened 2026-05-07
Fix direct image extraction dispatch

jioffe502:
## Summary
- Default `GraphIngestor.extract()` to `extraction_mode="auto"` so supported non-PDF inputs are routed through `MultiTypeExtractOperator` by extension.
- Preserve `extraction_mode="pdf"` as the explicit dedicated PDF/document graph path.
- Add coverage for BMP/TIFF/TIF page image materialization and mixed PDF+BMP auto dispatch.
- Add `graph_ingestor.md` documenting GraphIngestor usage, run modes, extraction modes, supported auto-dispatch groups, split configuration, and post-extraction stages.

## Testing
- `python -m pytest nemo_retriever/tests/test_ingest_interface.py -q`
- Full JP20 pipeline smoke-tested with the auto dispatch path.


nemo_retriever/src/nemo_retriever/graph_ingestor.md
```diff
@@ -0,0 +1,213 @@
+# GraphIngestor
```
jperez999, line 1: make sure this is still up to date with all the changes we made recently.

jperez999 (approved):

PR #1999, opened 2026-05-08
Surface remote NIM stage errors

charlesbluca:
## Description

This PR makes graph ingestion fail visibly when a caller explicitly configures a remote NIM endpoint and that endpoint reports row-level connection or inference errors.

Previously, stages such as page-elements caught remote transport failures and stored them in payload columns like `page_elements_v3.error`, while `GraphIngestor(...).ingest()` returned the DataFrame as if ingestion had succeeded. That made unreachable explicit endpoints hard to detect in Python library mode.

Changes:
- Adds `GraphIngestionError`, raised after graph execution when explicit remote NIM-backed stages report structured row-level errors.
- Adds `error_policy="collect"` to `GraphIngestor` and `create_ingestor` for callers that intentionally want partial results with row-level error payloads.
- Centralizes stage error detection across result columns instead of patching only page-elements.
- Allows `get_error_rows()` to inspect in-process pandas DataFrames as well as Ray datasets.
- Documents strict vs. collect behavior for Python library mode.

Validation:
- `uv run --project nemo_retriever --extra dev pytest nemo_retriever/tests/test_ingest_interface.py nemo_retriever/tests/test_params_models.py nemo_retriever/tests/test_graph_pipeline_cli.py -q`
- `git diff --check`

## Checklist
- [x] If adjusting docker-compose.yaml environment variables have you ensured those are mimicked in the Helm values.yaml file. N/A: this PR does not adjust docker-compose environment variables.

### DIFF
--- a/nemo_retriever/src/nemo_retriever/graph_ingestor.py
+++ b/nemo_retriever/src/nemo_retriever/graph_ingestor.py
@@ -29,7 +29,7 @@ from __future__ import annotations
 import json
 import os
 import sys
-from typing import Any, Callable, Dict, List, Optional, Union
+from typing import Any, Callable, Dict, Iterable, Iterator, List, Optional, Union
 
 from nemo_retriever.graph import InprocessExecutor, RayDataExecutor
 from nemo_retriever.graph.ingestor_runtime import batch_tuning_to_node_overrides, build_graph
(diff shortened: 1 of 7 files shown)


nemo_retriever/src/nemo_retriever/graph_ingestor.py
```diff
@@ -459,46 +531,150 @@ def ingest(self, params: Any = None, **kwargs: Any) -> Any:
...
+        if key == "failed" and isinstance(value, bool):
+            return value
+        if isinstance(value, str):
+            return bool(value.strip())
+        if isinstance(value, (list, tuple, set, dict)):
+            return len(value) > 0
+        return bool(value)
+
+    @staticmethod
+    def _jsonish(value: str) -> Any:
```
jperez999: why is this _jsonish method required? it seems like your trying to read this from a string? Do you expect to find malformed or inconsistent error reporting?

jperez999 (approved):

nemo_retriever/src/nemo_retriever/graph_ingestor.py
charlesbluca: Going to chalk up to codex slop:

> It is not required. I originally added it defensively, but the graph stages return structured dict payloads, not JSON-encoded strings. String-valued errors are still handled when they are under a real "error" field. I removed _jsonish to keep the detection contract limited to structured row-level error payloads.

Dropped in https://github.com/NVIDIA/NeMo-Retriever/pull/1999/commits/5e9dc7ace41f6a34decc769eae4433189254af87, thanks for the catch

PR #2007, opened 2026-05-08
[bug-fix] Normalising missing input path errors in graph ingest across modes

mahikaw:
## Description

- Validated graph ingest input paths before file loading in both batch and inprocess modes.
- Raised a consistent NeMo-level FileNotFoundError for missing literal paths instead of leaking Ray/PyArrow errors (batch mode) or silently returning an empty DataFrame (in-process mode)

Addresses [6154864](https://nvbugspro.nvidia.com/bug/6154864) 

## Checklist


nemo_retriever/src/nemo_retriever/utils/input_files.py
```diff
@@ -12,6 +17,62 @@
...
+        pattern = str(Path(raw_path).expanduser())
+        matches = glob.glob(pattern, recursive=True)
+        if matches:
+            expanded.extend(sorted(matches))
+        elif is_explicit_glob_path(pattern):
+            expanded.append(pattern)
+        elif not Path(pattern).exists():
+            raise_input_path_not_found(pattern)
+        else:
+            expanded.append(pattern)
```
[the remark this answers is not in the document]
mahikaw: although directories are not expected as inputs here (should get converted to files or glob patterns at cli level),  adding directory rejection error path in case of direct usage of executors APIs.

nemo_retriever/src/nemo_retriever/utils/input_files.py
```diff
@@ -12,6 +17,80 @@
...
+    expanded: list[str] = []
+    for input_path in paths:
+        raw_path = fspath(input_path)
+        if _has_uri_scheme(raw_path):
+            expanded.append(raw_path)
+            continue
+
+        local_path = Path(raw_path).expanduser()
+        if local_path.is_dir():
+            raise IsADirectoryError(f"Input path is a directory, not a file or glob pattern: {local_path}")
```
jperez999: So i think we want to allow directory level input paths and we assume everything in that path is part of the input data.

nemo_retriever/src/nemo_retriever/utils/input_files.py
```diff
@@ -12,6 +17,80 @@
     "video": ("*.mp4", "*.mov", "*.mkv"),
 }
 
+InputPath = str | PathLike[str]
+
+
+def _has_uri_scheme(path: str) -> bool:
```
jperez999: What pattern were you expecting to leverage this urlparse?

nemo_retriever/src/nemo_retriever/utils/input_files.py
```diff
@@ -12,6 +17,80 @@
...
+    cause
+        The lower-level ``FileNotFoundError`` raised by the reader.
+
+    Raises
+    ------
+    FileNotFoundError
+        Always raised with a product-level message and ``cause`` chained.
+    """
+    if len(input_paths) == 1:
+        _raise_input_path_not_found(input_paths[0], cause)
```
jperez999: cant we just str(paths) and get the output here without the extra conversion, and why is the input paths with a list of one handled different from the others. It should be the same raise no?

jperez999 (changes requested):

nemo_retriever/src/nemo_retriever/utils/input_files.py
mahikaw: combined into single message in all scenarios

nemo_retriever/src/nemo_retriever/utils/input_files.py
mahikaw: removed uri checking. 
as discussed offline, we want to support remote urls in a later pr using `fsspec`

nemo_retriever/src/nemo_retriever/utils/input_files.py
mahikaw: for direct `Executor` calls strictly, existing code allows directory level input paths if they are passed as valid glob patterns into the executor class. this remains unchanged.
however, it silently skips "some_dir/some_sub_dir" style of directories in direct `Executor` calls. with this PR, an exception is raised prompting the user to use glob patterns in direct calls eg "some_dir/some_sub_dir/**.pdf"
converging __main__.py contract (eg: expanding non-glob dir paths as well) in direct `Executor` calls (bypassing cli scenario) will require further code refactoring (out of bug fix scope)

nemo_retriever/src/nemo_retriever/utils/input_files.py
```diff
@@ -12,6 +16,59 @@
...
+    for input_path in paths:
+        raw_path = fspath(input_path)
+        pattern = str(Path(raw_path).expanduser())
+        matches = [match for match in glob.glob(pattern, recursive=True) if Path(match).is_file()]
+        if matches:
+            expanded.extend(sorted(matches))
+        elif _is_explicit_glob_path(pattern):
+            expanded.append(pattern)
+        elif not Path(pattern).exists():
+            raise_input_path_not_found(pattern)
```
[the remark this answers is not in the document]
mahikaw: out of current pr scope: https://github.com/NVIDIA/NeMo-Retriever/pull/2007#discussion_r3211440012

jperez999 (approved):

PR #2020, opened 2026-05-12
Fix local query embedding regressions for recall evaluation and simplify beir pipeline command

ChrisJar:
### Summary

Simplifies BEIR recall evaluation and fixes local query embedding regressions.

- Adds first-class BEIR dataset resolution so known datasets like `bo767`, `bo10k`, `jp20`, `earnings`, `financebench`, and `vidore_v3_*` can be run with shorthand `--beir-dataset-name` values.
- Routes local query embedding through model-native `embed_queries()` when available, instead of the generic passage embedding path.
- Stabilizes HF `llama-nemotron-embed-1b-v2` query embeddings with fixed-length query padding.
- Enables native vLLM query embedding by allowing retrieval to read vectors from payload/direct embedding columns, not only `metadata["embedding"]`.

### Why

BO767 recall had dropped substantially, from the historical `recall@5 ~= 0.835` range to about `0.524`. The root cause was the HF query embedding path switching to batch-local dynamic padding, which made query vectors depend on batch grouping for `llama-nemotron-embed-1b-v2`.

Because recall quality is the main product signal here, query embeddings need to be stable and reproducible across batch sizes and execution paths. Fixed-length query padding restores the expected BO767 recall level.

The vLLM query path also could not complete evaluation reliably because its native output shape was not accepted by retrieval vector extraction.

Separately, the BEIR CLI was too easy to misconfigure because each dataset required loader, annotation path, doc-id field, and k values to be specified manually. The new resolver encodes those known dataset defaults while preserving explicit CLI overrides.

### Validation

- Added/updated unit coverage for BEIR dataset resolution, HF query padding behavior, runtime query routing, and vLLM query vector extraction.
- Ran focused tests:
  - `nemo_retriever/tests/test_retriever_queries.py`
  - `nemo_retriever/tests/test_beir_evaluation.py`
- Verified BO767 recall:
  - Before fix: `recall@5 ~= 0.524`
  - HF fixed-padding query backend: `recall@5 ~= 0.8378`
  - vLLM native query backend: `recall@5 ~= 0.8388`, `recall@10 ~= 0.8856`

## Checklist

### DIFF
--- a/nemo_retriever/src/nemo_retriever/model/local/llama_nemotron_embed_1b_v2_hf_embedder.py
+++ b/nemo_retriever/src/nemo_retriever/model/local/llama_nemotron_embed_1b_v2_hf_embedder.py
@@ -10,6 +10,7 @@ evaluation) while document ingestion uses vLLM elsewhere.
 
 from __future__ import annotations
 
+import logging
 import warnings
 from dataclasses import dataclass
 from typing import List, Optional, Sequence
@@ -18,6 +19,8 @@ import torch
 
 from nemo_retriever.utils.hf_cache import configure_global_hf_cache_base
 from nemo_retriever.utils.hf_model_registry import get_hf_revision
+
+logger = logging.getLogger(__name__)
 
 
 def _l2_normalize(x: torch.Tensor, eps: float = 1e-12) -> torch.Tensor:
@@ -34,6 +37,8 @@ class LlamaNemotronEmbed1BV2HFEmbedder:
     hf_cache_dir: Optional[str] = None
     normalize: bool = True
     max_length: int = 8192
+    # Fixed query padding length. Queries longer than this are truncated; larger
+    # values preserve more text but increase query embedding cost.
     query_max_length: int = 128
     model_id: Optional[str] = None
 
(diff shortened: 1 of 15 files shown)


nemo_retriever/src/nemo_retriever/model/local/llama_nemotron_embed_1b_v2_hf_embedder.py
```diff
@@ -129,6 +139,35 @@ def embed(self, texts: Sequence[str], *, batch_size: int = 64) -> torch.Tensor:
...
+
+            if isinstance(input_ids, list):
+                length = len(input_ids[0]) if input_ids and isinstance(input_ids[0], list) else len(input_ids)
+                if length > max_length:
+                    truncated += 1
+
+        if truncated:
+            logger.warning(
+                "Truncating %d/%d HF query embeddings to query_max_length=%d tokens; "
+                "increase query_max_length if long queries lose recall.",
```
charlesbluca: In practice, are we actually able to override the default 128 max length like this suggests? Agent flagged that there wasn't a means of overriding

charlesbluca (approved):

PR #2032, opened 2026-05-13
Change pipeline evaluation default to none and rename audio recall mode

ChrisJar:
## Checklist


nemo_retriever/src/nemo_retriever/harness/config.py
```diff
@@ -161,20 +161,22 @@ def validate(self) -> list[str]:
                 errors.append("service_max_concurrency must be >= 1")
             return errors
 
-        if self.evaluation_mode not in VALID_EVALUATION_MODES:
+        if self.evaluation_mode == "recall":
```
jperez999: I dont think we need this, we haven't actually released with this setup before so there is no reason to alert users to the change.

nemo_retriever/src/nemo_retriever/harness/config.py
```diff
@@ -161,20 +161,22 @@ def validate(self) -> list[str]:
...
+            errors.append("evaluation_mode=recall has been renamed to evaluation_mode=audio_recall")
+        elif self.evaluation_mode not in VALID_EVALUATION_MODES:
             errors.append(f"evaluation_mode must be one of {sorted(VALID_EVALUATION_MODES)}")
 
-        if self.evaluation_mode == "recall" and self.recall_required and not self.query_csv:
+        if self.evaluation_mode == "audio_recall" and self.recall_required and not self.query_csv:
```
jperez999, line 169: I feel like we can do all these checks inside one if block (multiple cascading if statements but they shouldn't need to all be checks at the same level, what do you think?

jperez999: talking about lines 175-179

nemo_retriever/src/nemo_retriever/harness/run.py
```diff
@@ -306,6 +306,8 @@ def _resolve_lancedb_uri(cfg: HarnessConfig, artifact_dir: Path) -> str:
 def _build_command(cfg: HarnessConfig, artifact_dir: Path, run_id: str) -> tuple[list[str], Path, Path, Path | None]:
     runtime_dir = artifact_dir / "runtime_metrics"
     runtime_dir.mkdir(parents=True, exist_ok=True)
+    if cfg.evaluation_mode == "recall":
```
jperez999: With your changes I don't think we should ever hit this, right?

nemo_retriever/src/nemo_retriever/pipeline/__main__.py
```diff
@@ -1200,13 +1204,15 @@ def run(
             raise ValueError(f"Unsupported --run-mode: {run_mode!r}")
         if audio_split_type not in {"size", "time", "frame"}:
             raise ValueError(f"Unsupported --audio-split-type: {audio_split_type!r}")
-        if evaluation_mode not in {"none", "recall", "beir", "qa"}:
-            raise ValueError(f"Unsupported --evaluation-mode: {evaluation_mode!r}")
         if evaluation_mode == "recall":
```
jperez999: Same as above about this check. I dont think it is necessary since we never released with this. But if you do think we need, let me know.

jperez999 (approved):

jperez999 (changes requested):
please address previous comments

jioffe502 (approved):

jperez999 (approved):

PR #2052, opened 2026-05-18
Gate service ffmpeg install at runtime

charlesbluca:
## Description

Make ffmpeg/ffprobe optional in the bundled service image by removing the custom source-build installer and replacing the old build-time path with a runtime install flow.

The service image now keeps ffmpeg/ffprobe out by default. When audio or video extraction is needed, users can set `INSTALL_FFMPEG=true` for the service container, or set `service.installFfmpeg=true` in Helm. The service entrypoint installs the Ubuntu `ffmpeg` package at startup through narrowly scoped passwordless `sudo`, which provides both `ffmpeg` and `ffprobe`.

This PR also:

- Splits media dependency checks into distinct `ffmpeg-python`, `ffmpeg`, and `ffprobe` checks so missing dependencies produce actionable install guidance.
- Uses the granular checks where possible, so audio extraction and frame extraction only require `ffmpeg-python` plus the `ffmpeg` CLI, while media splitting/probing still requires `ffprobe`.
- Adds a Helm `service.installFfmpeg` value and a duplicate-env guard so users cannot set both `service.installFfmpeg=true` and `INSTALL_FFMPEG` manually in `service.env`.
- Updates README, Helm, troubleshooting, prerequisites, audio/video, and release-note docs for the runtime install flow and caveats.

Validation:

- `env UV_CACHE_DIR=/localhome/charlesb/.cache/tmp/uv-cache uv run --directory nemo_retriever --extra dev pytest tests/test_container_ffmpeg_install.py tests/test_media_dependency_availability.py -q`
- `git diff --check`
- `bash -n docker/scripts/retriever_service_entrypoint.sh`
- `black --check` and `flake8` on touched Python files
- focused pre-commit hooks for whitespace, EOF, large files, AST, and debug statements

## Checklist


nemo_retriever/src/nemo_retriever/audio/media_interface.py
```diff
@@ -48,8 +96,10 @@ def _probe(
     timeout: Optional[float] = None,
     **kwargs: Any,
 ) -> Any:
-    if not _FFMPEG_AVAILABLE or ffmpeg is None:
-        raise RuntimeError("ffmpeg is required for media probing; install ffmpeg-python and system ffmpeg.")
+    if not is_ffmpeg_python_available():
```
jperez999: Can this be collapsed into an or statement? if not, why?

nemo_retriever/src/nemo_retriever/audio/media_interface.py
```diff
@@ -79,8 +129,10 @@ def _run_ffmpeg(stream: Any, *, label: str, input_path: str) -> None:
     tempfile instead — file writes never block, so ffmpeg always makes progress
     and the call returns. We only read stderr when ``returncode != 0``.
     """
-    if ffmpeg is None:
-        raise RuntimeError("ffmpeg-python is not installed.")
+    if not is_ffmpeg_python_available():
```
jperez999: same here as above.

nemo_retriever/src/nemo_retriever/audio/media_interface.py
```diff
@@ -26,11 +26,59 @@
 
 try:
```
jperez999, line 27: couldnt we do all these checks here, that way they are not peppered throughout the code modules. This way all the gating logic is always at the top of the module that is affected. Wdyt?

jperez999 (changes requested):

nemo_retriever/helm/README.md
```diff
@@ -68,12 +68,32 @@ then override `service.image.repository` / `service.image.tag`:
 ```bash
 # from the repo root:
 docker build \
-    -f nemo_retriever/Dockerfile \
+    -f Dockerfile \
```
jperez999: i think from repo root you automatically find this. its the only dockerfile we have.

nemo_retriever/helm/README.md
```diff
@@ -68,12 +68,32 @@ then override `service.image.repository` / `service.image.tag`:
...
+system packages at deploy time; it only selects which image Kubernetes runs.
+If you need audio or video extraction, build an ffmpeg-enabled image and point
+the chart at it:
+
+```bash
+# from the repo root:
+docker build \
+    -f Dockerfile \
+    --target service \
+    --build-arg INSTALL_FFMPEG=true \
```
jperez999: should not be a build-arg should be a`docker run --env INSTALL_FFMPEG=true ...` and that should activate the install before the service starts.

jperez999 (changes requested):
Seems like this PR does not actually allow for the user to activate the installation on running the container. We want this to actually pull down the built container and before it starts running the service, it should look for this env var and it should install ffmpeg if activated. `docker run --env INSTALL_FFMPEG=true ...`

nemo_retriever/src/nemo_retriever/audio/media_interface.py
charlesbluca: Think the biggest thing we'd want to be careful around is collapsing the import-check into just "are ffmpeg-python & ffmpeg/ffprobe available?" as it looks like there are methods of extraction (i.e. frame extraction) that exclusively make use of ffmpeg but not ffprobe

IMO, I would feel better about keeping this PR restricted mostly to media interface import contract stuff (making it non-default, ensuring users an informative error message when it's not available), and following up with a media interface clean-up PR post-release that cleanly centralizes the import checks

nemo_retriever/helm/README.md
charlesbluca: Taking a look at this now - want to point out that it seems like our current Docker/Helm setup may require changes to allow apt package installs at runtime (versus build-time):

>This is a real design pivot from the current build-time flag. The current service image runs as `USER nemo`, and Helm defaults to `runAsNonRoot` / `runAsUser: 1000`, so startup-time apt-get install will not work without changing the container entrypoint/security model.

sosahi:
Just a reminder to merge this into the 26.05 branch as well when ready.

jperez999 (approved):

PR #2063, opened 2026-05-19
feat(tabular): TabularLanceDB subclass for VDB-agnostic label/databas…

tomer-levin-nv:
…e_name filtering

Promote `label` and `database_name` to first-class top-level columns on the tabular VDB row via a new `TabularLanceDB(LanceDB)` subclass under `tabular_data/vdb/`, mirroring the Postgres tabular ingest schema.

Tabular semantic search now filters with plain SQL column equality (`WHERE label = 'Column' AND database_name = 'dor_prod'`), which runs identically on LanceDB (DataFusion) and Postgres / pgvector with no per-backend dialect translation — no `filter=` contract added to the reference VDB ABC.

- Add overridable `_build_schema()` / `_make_row()` hooks on `LanceDB` so subclasses extend the schema and per-row projection without duplicating the filter/count loop. Default behavior unchanged.
- Register `vdb_op="tabular_lancedb"` in the VDB factory; wire `debug_ingest.py` and `generate_sql.py` to use it.
- Replace the LIKE-on-JSON `_build_metadata_where_clause` with a plain SQL builder; surface `label` / `database_name` on hits.

## Checklist


nemo_retriever/src/nemo_retriever/vdb/lancedb.py
```diff
@@ -201,10 +201,29 @@ def _get_text_for_element(element):
         return metadata.get("content")
 
 
+def _default_row_builder(element, *, embedding, text, content_meta, source_meta) -> dict:
```
jperez999, line 193: why is all this extra code still here?

nemo_retriever/src/nemo_retriever/vdb/factory.py
```diff
@@ -11,11 +11,16 @@ def get_vdb_op_cls(vdb_op: str):
...
 
-    available_vdb_ops = ["lancedb"]
+    available_vdb_ops = ["lancedb", "tabular_lancedb"]
 
     if vdb_op == "lancedb":
         from nemo_retriever.vdb.lancedb import LanceDB
 
         return LanceDB
 
+    if vdb_op == "tabular_lancedb":
```
jperez999: Why do we need to do this, like this? Cant you just create the vdb_op ahead of the graph and pass it in as a vdb_op? This keeps current NRL customers from seeing code that might lead to confusion.

nemo_retriever/src/nemo_retriever/tabular_data/vdb/tabular_lancedb.py
```diff
@@ -0,0 +1,80 @@
...
+            content_meta=content_meta,
+            source_meta=source_meta,
+        )
+        if isinstance(content_meta, dict):
+            row["label"] = _coerce_str(content_meta.get("label"))
+            row["database_name"] = _coerce_str(content_meta.get("database_name"))
+        else:
+            row["label"] = None
+            row["database_name"] = None
+        return row
```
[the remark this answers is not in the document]
jperez999: I understand that this is a code smell, but it allows us to keep the current setup without further convoluting the codebase that users will reference from the docs. When we are ready to incorporate structured data, then we can work on combining logic and make something that works for both structured and non-structured data.

nemo_retriever/src/nemo_retriever/vdb/lancedb.py
jperez999: Just because greptile says so, does not make it so. In this case we are trying to hide this because NRL users will not leverage this, because we do not support structured data yet.

jperez999 (changes requested):

PR #2111, opened 2026-05-25
fix(tabular): raise SQLSyntaxError on unparseable SQL

DinaLaptii:
Previously `extract_tables_and_columns` swallowed sqlglot parse/token errors and returned an empty `ExtractionResult`, making truly invalid SQL indistinguishable from valid SQL that referenced no known tables. Both reached callers as the same silent "no tables found" signal, so the SQL-validation agent reported success on syntactically broken SQL.

Introduce a dedicated `SQLSyntaxError(ValueError)` and raise it when sqlglot rejects the input. Valid-but-unresolved SQL still returns the empty extraction as before, so callers can classify the two outcomes.

## Checklist


liavnave (approved):

jioffe502 (approved):
per liav approval

PR #2119, opened 2026-05-26
Sohail/re org skill

sosahi:
## Checklist


jperez999:
where are the eval metrics for runs using this setup? Please post here. Also why do you have commits going back 4 months for file changes that only started 2 weeks ago?

edknv:
This looks like a verbatim copy of the skill, only to a different location, so it probably doesn't need a new eval sweep, but if we *must* move the skill to `skills/` I think the best way to do that is to use `git mv` to move the files and create symlinks at `.claude` and `.agents` that link to the new directory.

PR #2135, opened 2026-05-27
docs: update Retriever constructor examples

jioffe502:
## Summary
- update public direct `Retriever(...)` examples to use `vdb_kwargs`, `embed_kwargs`, and `rerank`
- replace unsupported `graph_pipeline --lancedb-uri` examples with `--vdb-kwargs-json` in README/evaluation/workflow docs
- add focused docs/notebook regression tests for unsupported direct-constructor kwargs and graph_pipeline options
- add a 26.05 migration note for the grouped constructor surface, including `local_query_embed_backend` mapping

## Context
QA is validating docs from GitHub Pages built from `main`; release-branch PR #2134 is merged and the narrow 26.05 follow-up is #2136.

## Testing
- `python -m py_compile nemo_retriever/tests/test_src_documentation_snippets.py && git diff --check`
- `UV_CACHE_DIR=$(mktemp -d /tmp/uv-cache-nemo-main.XXXXXX) uv run --project nemo_retriever --with pytest --with neo4j pytest nemo_retriever/tests/test_src_documentation_snippets.py`
- constructor smoke test for updated README/CLI examples
- graph_pipeline CLI smoke test for rejected `--lancedb-uri` and accepted `--vdb-kwargs-json`

PR #2136, opened 2026-05-27
docs: fix graph_pipeline LanceDB examples

jioffe502:
## Summary
- replace unsupported `graph_pipeline --lancedb-uri` examples with `--vdb-kwargs-json` in README/evaluation/workflow docs
- update the evaluation README diagram to avoid documenting the removed option
- align the docs regression guard with #2135 and cover the workflow docs page too
- update the migration note with the `local_query_embed_backend` grouped-kwargs mapping

## Context
Follow-up to #2134 after review noticed ingestion docs still used an option rejected by the shipped `graph_pipeline` CLI. This is kept aligned with the main/GitHub Pages companion PR #2135.

## Testing
- `python -m py_compile nemo_retriever/tests/test_src_documentation_snippets.py && git diff --check`
- `UV_CACHE_DIR=$(mktemp -d /tmp/uv-cache-nemo-2605.XXXXXX) uv run --project nemo_retriever --with pytest --with neo4j pytest nemo_retriever/tests/test_src_documentation_snippets.py`
- CLI smoke: `--lancedb-uri` is rejected; `--vdb-kwargs-json` appears in help

### DIFF
--- a/docs/docs/extraction/releasenotes.md
+++ b/docs/docs/extraction/releasenotes.md
@@ -10,7 +10,7 @@ To upgrade the Helm charts for this release, refer to the [NeMo Retriever Helm c
 
 Highlights for the 26.05 release line include everything in [26.03](#2603-release-notes-2630) plus changes on `main` merged into the `26.05` branch. See the [Git compare view](https://github.com/NVIDIA/NeMo-Retriever/compare/26.03...26.05) for the full commit list.
 
-**Migration note:** Direct `Retriever(...)` construction uses grouped configuration dictionaries. Replace flat `lancedb_uri=`, `lancedb_table=`, `embedder=`, `embedding_endpoint=`, and `reranker=` arguments with `vdb_kwargs={...}`, `embed_kwargs={...}`, and `rerank=...`. Helper APIs that document their own flat kwargs keep their own compatibility layer.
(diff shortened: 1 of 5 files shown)


charlesbluca:
One remaining public docs page still has the removed `graph_pipeline` flag: `docs/docs/extraction/workflow-document-ingestion.md` shows `python -m nemo_retriever.examples.graph_pipeline ... --lancedb-uri lancedb`. Since this command rejects `--lancedb-uri`, can we update it to the same `--vdb-kwargs-json` form and include this file in `_PUBLIC_GRAPH_PIPELINE_DOCS`?

Also, for parity with the main companion PR #2135, it would be good to align `_UNSUPPORTED_DIRECT_RETRIEVER_KWARGS` here with the broader 2135 set (`vdb` and `local_query_embed_backend` in addition to the existing flat kwargs). Not required for the graph-pipeline fix itself, but it avoids the 26.05 and main regression guards drifting immediately.

jioffe502:
@charlesbluca addressed both items: the workflow ingestion page now uses `--vdb-kwargs-json` and is included in `_PUBLIC_GRAPH_PIPELINE_DOCS`; `_UNSUPPORTED_DIRECT_RETRIEVER_KWARGS` is aligned with #2135, including `vdb` and `local_query_embed_backend`.

I also carried over the same small guard cleanups and migration-note mapping so 26.05 and main stay in sync. Re-ran the docs snippet test with a clean uv cache: `58 passed, 2 skipped`.

charlesbluca (approved):

jioffe502:
Same Docker CI follow-up as #2135: the failure was from the missing-doc assertion in the package-only Docker image, not from remaining bad docs examples. The Docker image has `nemo_retriever/` under `/workspace` but not repo-root `README.md` / `docs/docs/...`.

I aligned the helper here too: skip repo-root-only docs in that image, keep asserting for missing package docs. Local docs snippet test still passes: `58 passed, 2 skipped`.

PR #2137, opened 2026-05-27
setting up the workflow for onboarding

sosahi:
## Checklist


skills/nemo-retriever/references/ingest.md
```diff
@@ -0,0 +1,124 @@
+# retriever ingest
```
jperez999, line 1: Your not just adding the workflow, your adding these files for progressive disclosure. Do you have evals that show this is correct? Can you only upload your team-request file separately. We don't know this is  a good setup. We have other devs working on this, and I feel better with those because they are being compared against eval metric.

jperez999 (changes requested):

PR #2148, opened 2026-05-28
add checks against service mode params

jperez999:
## Description
This PR fixes issues where the user is able to provide specific arguments via cli and they are ignored in the service run mode. Now, when an argument that is not accepted by the service is added to a cli command for service mode it will raise an error.

## Checklist


nemo_retriever/src/nemo_retriever/pipeline/__main__.py
```diff
@@ -98,6 +98,132 @@
...
+    ("--pdf-split-batch-size", "pdf_split_batch_size"),
+    ("--pdf-extract-batch-size", "pdf_extract_batch_size"),
+    ("--pdf-extract-tasks", "pdf_extract_tasks"),
+    ("--pdf-extract-cpus-per-task", "pdf_extract_cpus_per_task"),
+    ("--nemotron-parse-actors", "nemotron_parse_actors"),
+    ("--nemotron-parse-gpus-per-actor", "nemotron_parse_gpus_per_actor"),
+    ("--nemotron-parse-batch-size", "nemotron_parse_batch_size"),
+    # In-graph VDB / sidecar metadata (not wired through ServiceIngestor by the CLI)
+    ("--vdb-op", "vdb_op"),
+    ("--vdb-kwargs-json", "vdb_kwargs_json"),
```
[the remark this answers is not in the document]
charlesbluca: Does this need to be resolved?

charlesbluca (changes requested):

randerzander (approved):

nemo_retriever/src/nemo_retriever/pipeline/__main__.py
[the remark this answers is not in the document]
charlesbluca: > I don’t think we should resolve this yet. I did some functional CLI probing against the PR branch using a fake `ServiceIngestor`, plus the NVBug 6234495 repro context, and this is still a real gap.
> 
> The intended fix for the bug is to fail fast for service-mode CLI flags that look accepted but are not actually sent to the service request. `--no-vdb` still exits successfully in service mode even though `enable_in_graph_vdb_upload = run_mode != "service" and not no_vdb` means it has no effect.
> 
> I also found the same issue for `--vdb-overwrite/--vdb-append`: those flags only mutate local VDB kwargs, but service mode does not pass a VDB upload spec from this CLI path, so they are also silently ignored.
> 
> Suggested follow-up:
> - Add `("--no-vdb", "no_vdb")`
> - Add `("--vdb-overwrite/--vdb-append", "vdb_overwrite")`
> - Add a regression test covering these VDB flags in `--run-mode service`
> 
> The rest of the bug repro flags I tested (`--method`, `--text-chunk`, `--text-chunk-max-tokens`, `--embed-granularity`, extraction toggles, caption/store controls) do now reject correctly.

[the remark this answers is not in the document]
jperez999: I will add both your suggested changes and greptile's  "--vdb-kwargs-json" as well.

nemo_retriever/src/nemo_retriever/pipeline/__main__.py
```diff
@@ -98,6 +98,134 @@
...
+#   --embed-model-name, --embed-modality, --embed-invoke-url, --api-key
+#       (client-side query embedding during evaluation)
+#   --save-intermediate, --detection-summary-file
+#       (post-ingest local outputs)
+#   --audio-match-tolerance-secs
+#       (audio recall evaluation)
+#   everything in I/O, Service, Evaluation, Observability panels
+_SERVICE_INCOMPATIBLE_FLAGS: tuple[tuple[str, str], ...] = (
+    # Extract
+    ("--method", "method"),
```
charlesbluca, line 116: One more thing before this goes green: the CI fixes above make the tests behave as intended, but this reject list still looks too broad after the rebase.

Current `26.05` has coverage asserting service mode forwards `method`, `extract_text`, `dpi`, `split_config`, and `embed_granularity` into the `ServiceIngestor` payload.

So flags like `--method`, `--dpi`, `--text-chunk*`, and `--embed-granularity` should probably not be rejected anymore unless this PR is intentionally reverting that service override support. VDB flags still look valid to reject.

nemo_retriever/src/nemo_retriever/pipeline/__main__.py
```diff
@@ -98,6 +98,134 @@
...
+    Only flags whose click parameter source is ``COMMANDLINE`` or
+    ``ENVIRONMENT`` are treated as user-supplied — flags carrying their
+    declared default do not trigger the error.
+    """
+    from click.core import ParameterSource
+
+    user_set: list[str] = []
+    for cli_flag, param_name in _SERVICE_INCOMPATIBLE_FLAGS:
+        source = ctx.get_parameter_source(param_name)
+        if source in (ParameterSource.COMMANDLINE, ParameterSource.ENVIRONMENT):
```
charlesbluca: This looks like the root cause for the first three CI failures.

PR CI installs with `uv pip install -e nemo_retriever[all,dev]`, which currently resolves newer Click/Typer than the lockfile-backed local `uv run` path. In that environment, `ctx.get_parameter_source(...)` returns `typer._click.core.ParameterSource`, while this code compares against `click.core.ParameterSource`, so the enum identity comparison never matches and `_reject_service_incompatible_flags()` returns without raising.

I reproduced this under a CI-style env as `method ParameterSource.COMMANDLINE typer._click.core False COMMANDLINE`.

Can we make this version-robust by checking the source name instead, e.g. `getattr(source, "name", None) in {"COMMANDLINE", "ENVIRONMENT"}`?

nemo_retriever/tests/test_graph_pipeline_cli.py
```diff
@@ -473,3 +473,119 @@ def open_table(self, _name):
...
+
+def test_graph_pipeline_cli_service_mode_accepts_allowlisted_flags(tmp_path, monkeypatch) -> None:
+    import nemo_retriever.service_ingestor as service_ingestor_module
+
+    dataset_dir = tmp_path / "dataset"
+    dataset_dir.mkdir()
+    (dataset_dir / "sample.pdf").write_text("placeholder", encoding="utf-8")
+    save_dir = tmp_path / "save"
+
+    class _FakeServiceIngestor(list):
```
charlesbluca, line 559: This test double needs to match the rebased service-mode path.

Current `26.05` now wires service ingest through `.files(...).extract(...).embed(...).ingest(...)`, but this fake only implements `files()` and `ingest()`, which is why CI fails with `AttributeError("_FakeServiceIngestor object has no attribute extract")`.

Adding chainable `extract(...): return self` and `embed(...): return self` here should address the allowlisted-flags failure.

nemo_retriever/src/nemo_retriever/pipeline/__main__.py
charlesbluca: Think this is relevant still? Are the above flags not getting used in service-mode?

nemo_retriever/tests/conftest.py
```diff
@@ -11,8 +11,17 @@
 
 from __future__ import annotations
 
+import os
+
 from fastapi.testclient import TestClient
```
charlesbluca, line 16: This should address the ANSI-splitting failure, but I would keep the fix local to these CLI assertions rather than changing Rich/Typer color behavior for the entire test suite. A small helper in `test_graph_pipeline_cli.py` should be enough, e.g. `from click import unstyle`, then `output = unstyle(result.output)` before checking for `--method`, `--no-vdb`, etc. That resolves the GitHub Actions color issue without globally mutating `NO_COLOR`/`TERM` for unrelated tests.

jperez999: I just created a conftest variable that handles the no color issue.

PR #2182, opened 2026-05-31
docs: add async service ingestor examples

### LINKED ISSUE #1134: [DOC]: Add docs for async ingestor API jobs
### How would you describe the priority of this documentation request

None

### Please provide a link or source to the relevant docs

https://github.com/NVIDIA/nv-ingest/tree/main/client/client_examples

### Describe the problems in the documentation

We should create examples showing how to launch async ingestor jobs

### (Optional) Propose a correction or improvement

_No response_

colesmcintosh:
## Description
Adds service-mode API guide examples for async ingestor jobs.

This documents:
- when to use `ingest()`, `ingest_stream()`, and `aingest_stream()`
- a synchronous streaming example using `ingest_stream()`
- a true async example using `async for event in ingestor.aingest_stream()`
- the event shapes emitted by service-mode streaming ingest jobs

Closes #1134.

## Checklist

Validation:
- `git diff --check` (clean)
- `mkdocs build -f mkdocs.yml --strict` passes
- verified Markdown code fences and event table counts


colesmcintosh:
Thanks for the review. Both points addressed in 574e93b4:

1. **`asyncio.run()` in the async example** — good catch. `asyncio.run()` cannot be nested inside a running loop, which is exactly the Jupyter/FastAPI case the prose calls out. I kept `asyncio.run(main())` for the standalone-script path and added a note that inside an already-running loop you `await main()` (or inline the `async for` loop) instead.

2. **`job_started` not handled in the examples** — added a sentence to the Event shapes intro clarifying that the examples handle the events most callers act on and that any unhandled event (such as `job_started`) can be safely ignored. I left the example code as-is rather than adding no-op branches, since the note conveys the intent without cluttering the illustrations.

PR #2205, opened 2026-06-03
Share root ingest path with pipeline

jioffe502:
## Description

This PR consolidates the root ingest / pipeline parity work

- Makes local `retriever pipeline run` (`batch` / `inprocess`) reuse the same ingest planning and execution path as `retriever ingest`.
- Keeps pipeline-specific behavior intact: service mode, evaluation modes, runtime summaries, intermediate outputs, generic VDB controls, and dev-only flags.
- Expands root `retriever ingest` with the durable user-facing ingest controls needed for parity:
  - table/graphic extraction controls
  - embedding modality/granularity controls
  - text chunking controls
  - `--api-key`
  - `--dedup` / `--dedup-iou-threshold`
  - `--store-images-uri`
  - `--pdf-split-batch-size`
  - Nemotron Parse tuning via `--nemotron-parse-*`
- Adds a shared structured ingest execution result used by both root ingest and pipeline local ingest.
- Marks `retriever pipeline` as development/compatibility-facing while keeping `retriever ingest` and `retriever query` as the public path.

## Intentionally Not Moved To Root Ingest

These remain pipeline-only for now:

- Evaluation/reporting flags: `--evaluation-mode`, BEIR, QA, recall, reranker, query CSV, retrieval save path.
- Service mode flags: `--service-url`, `--service-concurrency`, `--service-api-token`.
- Development/observability flags: `--runtime-metrics-dir`, `--runtime-metrics-prefix`, `--detection-summary-file`, `--debug`, `--log-file`.
- Intermediate/export behavior: `--save-intermediate`.
- Generic VDB controls: `--vdb-op`, `--vdb-kwargs-json`, `--no-vdb`.
- Sidecar metadata: `--meta-dataframe`, `--meta-source-field`, `--meta-fields`, `--meta-join-key`.
- Advanced caption tuning: caption device, GPU utilization, temperature/top-p, max tokens, actor GPU overrides.
- `--input-type`, because root ingest keeps auto-routing as the public behavior.

Some pipeline flag names are covered by root equivalents instead of copied directly, e.g. `--embed-actors` maps to root `--embed-workers`, and `--vdb-overwrite/--vdb-append` maps to root `--overwrite/--append`.

## Testing

- `python3 -m py_compile ...`
- `git diff --check`
- Focused pytest suite: `107 passed, 3 skipped`

DGX eval evidence:

- `jp20` batch: passed, non-empty LanceDB table, index present, query hits returned.
[...]

PR #2209, opened 2026-06-04
Split root CLI ingest and query workflows

jioffe502:
## Description

This PR splits the root CLI workflow architecture now that #2205 has merged.

The main goal is to remove the large mixed `sdk_workflow.py` module and give ingest/query their own clear ownership boundaries without changing CLI behavior.

- Deletes `adapters/cli/sdk_workflow.py`.
- Adds `nemo_retriever.ingest` as the core ingest planning/execution package.
- Adds thin CLI workflow adapters for root `retriever ingest` and `retriever query`.
- Keeps `retriever pipeline run` as a legacy/development/eval wrapper, but makes its local `batch` / `inprocess` path translate into the same core ingest plan used by `retriever ingest`.
- Does not change flags, eval behavior, harness behavior, or core graph construction behavior.

## Why

After #2205, root `retriever ingest` and `retriever query` are the forward-looking first-class CLI workflows. The remaining issue was structure: ingest planning, ingest execution, query setup, dry-run formatting, and pipeline compatibility logic were still tangled through one large root CLI module.

This PR makes the architecture easier for humans and agents to reason about:

```text
retriever ingest
  -> adapters/cli/main.py parses flags
  -> ingest/plan.py resolves typed options into ResolvedIngestPlan
  -> ingest/execution.py executes that plan
  -> GraphIngestor still owns real graph construction

retriever query
  -> adapters/cli/main.py parses flags
  -> adapters/cli/query_workflow.py calls Retriever.query()
  -> Retriever still owns query graph/retrieval policy

retriever pipeline run
  -> legacy/dev/eval command
  -> defaults to inprocess for small-document initial testing
  -> batch remains explicit for high-throughput/eval workflows
  -> batch/inprocess translates old pipeline flags into ResolvedIngestPlan
  -> service mode remains ServiceIngestor-owned for a follow-up migration
```

### Core ingest package

Introduces `nemo_retriever.ingest`:

- `plan.py`
  - typed ingest request option groups
  - `ResolvedIngestPlan`
  - `resolve_ingest_plan()`
  - manifest/profile/media/chunk/embed/storage resolution

- `execution.py`
  - `IngestExecutionResult`
  - `build_ingest_pipeline(plan)`
  - `execute_ingest_plan(plan)`
  - row verification and empty-ingest handling

`ResolvedIngestPlan` is now the single executable ingest contract. The earlier partial stage overlay was removed because it made ownership ambiguous.

### CLI adapters

Adds focused CLI workflow adapters:

- `adapters/cli/ingest_workflow.py`
  - dry-run rendering
  - root ingest execution policy

- `adapters/cli/query_workflow.py`
  - root query setup
  - thin delegation into `Retriever.query()`

`adapters/cli/main.py` remains the Typer flag parser. It constructs typed request groups and delegates.

### Pipeline compatibility

`retriever pipeline run` remains available, but its local `batch` / `inprocess` path no longer owns a competing local ingest construction model.

Instead, it now resolves an `IngestPlanRequest`, applies only pipeline-specific legacy overrides, and calls:

```python
execute_ingest_plan(
    ingest_plan,
    verify_rows=False,
    raise_on_empty=False,
)
```

Service mode remains separate and still uses `ServiceIngestor`. A follow-up PR should move service-mode ownership behind the first-class `retriever ingest` path before further reducing `retriever pipeline run`.

## What This Does Not Change

- No new CLI flags.
- No CLI flag removals.
- No eval migration.
- No harness migration.
- No service-mode migration.
- No changes to core graph construction.
- `GraphIngestor` / graph runtime remain the owners of actual graph/operator construction.

## Intentional Default Policy

This PR keeps `retriever ingest --run-mode` Ray-free by default with `inprocess`; `batch` remains opt-in via `--run-mode batch`.

`retriever pipeline run` also defaults to `inprocess` because initial testing is expected to use small 1-3 document sets, where Ray batch startup hurts pages-per-second. `batch` remains explicit for high-throughput/eval workflows. The no-flag pipeline path is covered by a test that asserts it resolves `run_mode="inprocess"`.

The extraction defaults are intentionally aligned across root ingest and local pipeline ingest based on JP20 parity testing:

- `dpi=200`
- `extract_infographics=False`
- `extract_page_as_image=True`

`extract_page_as_image` remains typed as `Optional[bool]` to preserve the public parameter contract while changing the default value.

## Reviewer Guide

The most important review question is ownership:

- `adapters/cli/main.py` should only parse CLI flags and construct typed request objects.
- `nemo_retriever.ingest.plan` should own ingest option resolution.
- `nemo_retriever.ingest.execution` should own SDK ingest-chain execution.
- `adapters/cli/query_workflow.py` should stay a thin wrapper around `Retriever.query()`.
- `pipeline/__main__.py` should only bridge legacy pipeline flags into the core ingest path for local modes; service mode remains isolated until the follow-up service migration.

The intended diff shape is mostly moves, deletions, and import updates. The notable design choice is deleting `sdk_workflow.py` entirely instead of keeping a compatibility facade, because this is forward-looking for the next release and there is no backwards-compatibility requirement for that internal module.

## Checklist


nemo_retriever/src/nemo_retriever/adapters/cli/ingest_workflow.py
```diff
@@ -0,0 +1,94 @@
...
+        data = dict(params)
+    else:
+        return {"value": str(params)}
+    return _strip_secret_values(data)
+
+
+def _is_dry_run_secret_field(key: Any) -> bool:
+    normalized_key = str(key).lower().replace("-", "_")
+    return normalized_key.endswith("token") or any(
+        pattern in normalized_key for pattern in _DRY_RUN_SECRET_FIELD_PATTERNS
```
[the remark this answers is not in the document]
jperez999: We should make these private. These functions should only be leveraged through the cli tool, never on their own.

nemo_retriever/src/nemo_retriever/adapters/cli/ingest_workflow.py
```diff
@@ -0,0 +1,96 @@
...
+    )
+
+
+def _strip_secret_values(value: Any) -> Any:
```
jperez999, line 40: How do secrets end up in the model dump?

nemo_retriever/src/nemo_retriever/adapters/cli/main.py
```diff
@@ -575,88 +589,112 @@ def ingest_command(
...
+                    dedup=IngestDedupOptions(
+                        enabled=dedup,
+                        iou_threshold=dedup_iou_threshold,
+                    ),
+                    chunk=IngestChunkOptions(
+                        enabled=text_chunk,
+                        text_chunk_max_tokens=text_chunk_max_tokens,
+                        text_chunk_overlap_tokens=text_chunk_overlap_tokens,
+                    ),
+                    embed=IngestEmbedOptions(
```
jperez999, line 671: So I think there is overlap between some of these Ingest* classes and the query counterpart. We could probably generalize these param models, what do you think? This way we can leverage the same one in both ingest/query

nemo_retriever/src/nemo_retriever/adapters/cli/query_workflow.py
```diff
@@ -0,0 +1,91 @@
...
+
+from nemo_retriever.params import build_embed_option_kwargs
+from nemo_retriever.retriever import Retriever
+from nemo_retriever.utils.remote_auth import resolve_remote_api_key
+from nemo_retriever.vdb.records import RetrievalHit
+
+_LOCAL_VL_RERANK_MODEL = "nvidia/llama-nemotron-rerank-vl-1b-v2"
+
+
+def _build_rerank_kwargs(
```
jperez999, line 17: Are you going to eventually create a param model for these kwargs just like you did for the others?

nemo_retriever/src/nemo_retriever/adapters/cli/query_workflow.py
```diff
@@ -0,0 +1,91 @@
...
+    }
+    if embed_kwargs:
+        retriever_kwargs["embed_kwargs"] = embed_kwargs
+    if rerank:
+        rerank_kwargs = _build_rerank_kwargs(reranker_invoke_url, reranker_model_name, reranker_backend)
+        retriever_kwargs["rerank"] = True
+        if rerank_kwargs:
+            retriever_kwargs["rerank_kwargs"] = rerank_kwargs
+
+    retriever = Retriever(**retriever_kwargs)
```
jperez999, line 85: I like this here. But we definitely want to maintain the same process for both ingest and query.  Referring to Param models. It is cool, if you meant this PR to be intermediary and that refactor is coming. If that is the case, respond here with that. Otherwise we should fix this sooner rather than later.

nemo_retriever/src/nemo_retriever/ingest/execution.py
```diff
@@ -0,0 +1,213 @@
...
+
+@dataclass(frozen=True)
+class IngestExecutionResult:
+    """Structured result from executing a resolved root ingest plan."""
+
+    plan: ResolvedIngestPlan
+    result: object
+    n_rows: int | None
+    result_n_rows: int | None
+    initial_n_rows: int | None
```
jperez999, line 33: What is the difference between initial and result here? This will need a more detailed docstring explaining all the different parameters in the class.

nemo_retriever/src/nemo_retriever/ingest/execution.py
```diff
@@ -0,0 +1,213 @@
...
+    lancedb_uri: str
+    table_name: str
+    metadata: dict[str, Any]
```
jperez999: How is this separate from the individual chunks that are output? Why is it separated?

nemo_retriever/src/nemo_retriever/ingest/execution.py
```diff
@@ -0,0 +1,213 @@
...
+
+    @property
+    def documents(self) -> list[str]:
+        return list(self.plan.documents)
+
+    @property
+    def lancedb_target(self) -> str:
+        return f"{self.lancedb_uri}/{self.table_name}"
+
+    def to_summary_dict(self) -> dict[str, Any]:
```
jperez999, line 46: When I read to summary, I am thinking of something really light weight. All the documents being here... that seems real big. along with results. maybe this is a count of how many documents were ingested, and result is the number of chunks that came back. I dont see a reason to have this otherwise as I could pull this from the class directly. And same as above, what is n_rows and result_n_rows?

nemo_retriever/src/nemo_retriever/ingest/execution.py
```diff
@@ -0,0 +1,213 @@
...
+    """Execute a resolved ingest plan and return structured execution data.
+
+    Root ``retriever ingest`` uses row verification as its public success bar.
+    Development wrappers can disable verification while preserving the exact
+    shared plan/build/ingest path and layering their own reporting afterward.
+    """
+
+    lancedb_target = _resolve_lancedb_target(plan)
+    if verify_rows and lancedb_target is None:
+        raise ValueError("Row verification requires an effective VDB upload stage; pass verify_rows=False to skip it.")
```
jperez999: You dont need a vdb stage is you dont want one right? Like I could just go through the extraction pipeline and want the results. I get that the default will have one, but I dont think we should raise an error not allowing a pipeline to run without sending to a vdb. Thoughts?

nemo_retriever/src/nemo_retriever/ingest/execution.py
```diff
@@ -0,0 +1,213 @@
...
+
+def execute_ingest_plan(
+    plan: ResolvedIngestPlan,
+    *,
+    verify_rows: bool = True,
+    raise_on_empty: bool = True,
+) -> IngestExecutionResult:
+    """Execute a resolved ingest plan and return structured execution data.
+
+    Root ``retriever ingest`` uses row verification as its public success bar.
```
jperez999: What does row verification actually mean here? Is it I got back a set of chunks or is it these chunks are in the vdb? Is there a reason we can't support either?

nemo_retriever/src/nemo_retriever/ingest/execution.py
```diff
@@ -0,0 +1,213 @@
...
+
+    return IngestExecutionResult(
+        plan=plan,
+        result=result,
+        n_rows=n_rows,
+        result_n_rows=result_n_rows,
+        initial_n_rows=initial_n_rows,
+        lancedb_uri=lancedb_uri,
+        table_name=table_name,
+        metadata={
```
jperez999: In this metadata, can we also add the model names used for each stage that was run during ingestion? i.e. embed model, page elements, ocr, audio, etc...

nemo_retriever/src/nemo_retriever/ingest/execution.py
```diff
@@ -0,0 +1,182 @@
...
+    if n_rows > 0 and (initial_n_rows is None or n_rows > initial_n_rows):
+        return
+
+    if initial_n_rows is not None:
+        raise RuntimeError(
+            f"retriever ingest did not add rows to LanceDB {target}; row count stayed at {n_rows} "
+            f"for {len(documents)} input file(s). This usually means extraction or embedding failed before "
+            "any rows were written; check the captured stage logs above, and verify NVIDIA_API_KEY/NGC_API_KEY "
+            "or the configured local/remote endpoints."
+        )
```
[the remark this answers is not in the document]
jperez999: This seems like a gotcha we need to worry about. Thoughts?

nemo_retriever/src/nemo_retriever/pipeline/__init__.py
```diff
@@ -2,12 +2,11 @@
...
-"""End-to-end ingestion pipeline subcommand for the ``retriever`` CLI.
+"""Legacy end-to-end ingestion pipeline subcommand for the ``retriever`` CLI.
 
-This package wraps :class:`~nemo_retriever.graph_ingestor.GraphIngestor` with a
-Typer application that exposes every knob needed to run a full PDF / doc /
-txt / html / image / audio ingestion job, write results to LanceDB, and
-optionally evaluate recall or BEIR.
+This package owns the ``retriever pipeline`` Typer application. Local pipeline
+runs delegate ingest graph construction to the core ingest package; service
+runs still build the service client path directly.
```
jperez999, line 9: This will be changed later so that even service path will get piped through ingest cli sub command right?

nemo_retriever/src/nemo_retriever/pipeline/__main__.py
```diff
@@ -666,136 +584,60 @@ def _build_ingestor(
...
-                    api_key=caption_remote_api_key,
-                    model_name=caption_model_name,
-                    device=caption_device,
-                    context_text_max_chars=caption_context_text_max_chars,
-                    gpu_memory_utilization=caption_gpu_memory_utilization,
-                    temperature=caption_temperature,
-                    top_p=caption_top_p,
-                    max_tokens=caption_max_tokens,
-                )
+        ingestor = ingestor.caption(
```
jperez999, line 628: If we are truly using ingest for everything this calls here to ingestor should not exist. Unless they are specific to service and they are going away in a subsequent PR

jperez999 (changes requested):

nemo_retriever/src/nemo_retriever/adapters/cli/ingest_workflow.py
jioffe502: The dry-run path serializes the resolved Pydantic option models, and those models can include endpoint credentials like the NVIDIA API key. I added a short comment there and kept the redaction local to dry-run output.

nemo_retriever/src/nemo_retriever/adapters/cli/main.py
jioffe502: I agree there is conceptual overlap. I avoided introducing a shared generic model in this PR because ingest and query options still have different semantics, and I don’t want to create a premature common “options” layer. ATM its just sharing embedding.

My intended follow-up is for query to get typed params in nemo_retriever/query/, then extract only the truly shared contracts once both sides are clear.

nemo_retriever/src/nemo_retriever/adapters/cli/query_workflow.py
jioffe502: Yes, I view the current query workflow as intermediary. This PR keeps query as a thin adapter, and the next query-core work should introduce typed query params so ingest/query follow the same shape without forcing an awkward abstraction too early.

nemo_retriever/src/nemo_retriever/ingest/execution.py
jioffe502: Agreed this was unclear. I expanded the docstring: result_n_rows is the row count from the current ingest result when countable, while initial_n_rows/n_rows are LanceDB counts before/after execution when verification is enabled

nemo_retriever/src/nemo_retriever/ingest/execution.py
jioffe502: I renamed this to run_metadata to make clear it is execution-level metadata, not chunk/document metadata

nemo_retriever/src/nemo_retriever/ingest/execution.py
jioffe502: Agreed and Fixed: summary now only carries n_documents, target info, and row counts. The raw result and full document list stay on the execution object.

nemo_retriever/src/nemo_retriever/ingest/execution.py
jioffe502: The core path does support no-VDB execution when verify_rows=False; the error only applies when the caller asks for VDB row verification but the plan has no VDB target. There is also the option to skip vdb upload altogether

I clarified the docstring and error message

nemo_retriever/src/nemo_retriever/ingest/execution.py
jioffe502: Today it means: count the current ingest result when possible, then verify rows in the effective LanceDB upload target. For extraction-only plans, callers should disable VDB row verification.

nemo_retriever/src/nemo_retriever/pipeline/__init__.py
jioffe502: Yes. The remaining pipeline construction is service-specific for now and is ready to move in a stacked follow-up PR. This PR keeps local ingest/query first-class; the next one starts cutting service mode over.

nemo_retriever/src/nemo_retriever/pipeline/__main__.py
jioffe502: this is in the service mode investor and will be removed in follow on PR

nemo_retriever/src/nemo_retriever/ingest/execution.py
jioffe502: I agree with the goal. I’m hesitant to add this in this PR by scraping params.

 I think model provenance should be part of a real run trace/metadata contract, probably stage-owned, so extract/embed/caption/audio/video report the actual resolved model/backend consistently. 

Happy to track that as follow-up; I’d rather not bolt on incomplete model metadata in the CLI tool

nemo_retriever/src/nemo_retriever/adapters/cli/query_workflow.py
jioffe502: created an issue to track this and will be in very soon. 

https://github.com/NVIDIA/NeMo-Retriever/issues/2218

nemo_retriever/src/nemo_retriever/ingest/execution.py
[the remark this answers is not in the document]
jioffe502: resolved

jperez999 (approved):

PR #2211, opened 2026-06-05
Add ragas judge

jperez999:
## Description
This PR changes the logic used in LLMJudge operator from basic judging 1-5 via one call to a 2 judge setup with an average of both scores (0,2,4). This logic was taken from the ragas python codebase. This PR also adds the commit sha to all report building for the agent_eval.

Could not use ragas directly because:

The judge cannot import ragas in this repo's .venv — and it's not a PyPI-vs-fork issue. Both ragas 0.4.3 and the fork unconditionally do from langchain_community.chat_models.vertexai import ChatVertexAI in ragas/llms/base.py, but this venv's langchain-community (sunset 1.x) removed that submodule,
  and an older compatible one would clash with the langchain-core 1.x that langchain-nvidia-ai-endpoints needs. Those vertexai symbols are only used in an unrelated type tuple, so it's purely an import-time version skew, not a code problem.

  To actually run the judge you'll need either a venv where ragas's langchain-community still has chat_models.vertexai, or a ragas build that makes that import lazy (plus NVIDIA_API_KEY + network). I left the code degrading gracefully (a real judge() returns a clear judge_api_error naming the
  underlying import failure rather than crashing) and recorded the gotcha in memory.

  I did install the fork via --no-deps into the venv during verification — say the word if you'd like me to revert it to the PyPI 0.4.3 that was there before.

## Checklist


shan-nvidia:
LGTM. Could we add the details of why direct ragas integration is not working in the description? So we track the reason of this workaround.

agent_eval/build_report.py
```diff
@@ -512,8 +517,30 @@ def _render_tokens_md(tokens: dict, setup_metas: list[dict]) -> str:
...
+    report_info = gitinfo.git_commit()
+    lines = ["\n## Code version\n"]
+    if code_commit:
+        lines.append(f"- eval-time code commit: {gitinfo.format_commit(code_commit)}\n")
+    else:
+        lines.append("- eval-time code commit: _not recorded — this run predates commit logging_\n")
+    lines.append(f"- report generated from working tree: {gitinfo.format_commit(report_info)}\n")
+    if not code_commit:
+        lines.append(
+            "- ⚠️ eval-time commit unknown; the report-time version above may differ "
```
jioffe502, line 532: knit picky but are we allowed to have emojis in the code

```suggestion
            "- Eval-time commit unknown; the report-time version above may differ "
```

agent_eval/functional_report.py
```diff
@@ -58,6 +76,31 @@ def main(argv=None) -> int:
...
+            any_run_commit = True
+            L.append(f"| {label} | {gitinfo.format_commit(ci)} |")
+        else:
+            L.append(f"| {label} | _not recorded — predates commit logging_ |")
+    L.append(
+        f"\nReport generated from working tree at {gitinfo.format_commit(report_info)}."
+        + (
+            ""
+            if any_run_commit
+            else " ⚠️ No run recorded its eval-time commit; the line above is the **report-time** "
```
jioffe502, line 98:
```suggestion
            else "No run recorded its eval-time commit; the line above is the **report-time** "
```

jioffe502 (approved):

PR #2217, opened 2026-06-08
Add Zipkin tracing parity

charlesbluca:
## Description
This adds Zipkin parity and broader trace-context propagation across the Helm deployment and service execution paths.

- Adds Helm support for Zipkin collector deployment, service wiring, endpoint overrides, and chart-wide tracing env generation guards.
- Keeps tracing observability-only while carrying trace context through gateway proxy requests, service work, pipeline pools/executors, and NIM HTTP/internal inference calls.
- Surfaces trace IDs through job/service responses and dashboard views so users can correlate service work with collected traces.
- Documents the Zipkin tracing workflow and adds focused tests for Helm rendering, tracing helpers, NIM tracing, proxy/pool/executor propagation, ingest routing, job tracking, and dashboard trace UI.

Compatibility note: `topology.zipkin.enabled` intentionally remains `true` by default to preserve the managed trace-backend behavior shipped by the legacy 26.1.2 Helm chart. Upgrades with default values may create a chart-owned Zipkin Deployment and Service; set `topology.zipkin.enabled=false` before upgrading if your deployment uses an external backend or should not run chart-owned Zipkin. Context: https://github.com/NVIDIA/NeMo-Retriever/pull/2217#discussion_r3375615743

Validation: `uv run pytest tests/test_helm_tracing_zipkin.py tests/test_nim_tracing.py tests/test_service_tracing.py tests/test_service_proxy_tracing.py tests/test_service_pool_tracing.py tests/test_service_pipeline_tracing.py tests/test_service_ingest_router.py tests/test_service_ingest_async.py tests/test_service_job_tracker.py tests/test_dashboard_static_trace_ui.py` (152 passed).

## Checklist


nemo_retriever/helm/values.yaml
```diff
@@ -443,6 +449,57 @@ topology:
             processors: [batch]
             exporters: [debug]
 
+  # ---------------------------------------------------------------------------
+  # Zipkin
+  # ---------------------------------------------------------------------------
```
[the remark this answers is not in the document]
charlesbluca: Good point; think since we're aiming for collector + Zipkin which is what we had in 26.1, this would be a welcome infra change, but is worth calling out explicitly for those who only wanted OTel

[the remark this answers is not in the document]
charlesbluca: Keeping this default on is intentional for functional parity with the legacy 26.1.2 Helm chart, which shipped a managed Zipkin backend enabled by default. Pod trace export remains opt-in via `service.otel.enabled` and `nimOperator.otel.enabled`, so the default here is only the chart-owned backend availability.

I documented the upgrade behavior and opt-out in `nemo_retriever/helm/README.md` and `nemo_retriever/helm/values.yaml`, and added the same compatibility note to the PR description with a link back to this P1: https://github.com/NVIDIA/NeMo-Retriever/pull/2217#discussion_r3375615743

nemo_retriever/helm/values.yaml
```diff
@@ -178,6 +178,12 @@ service:
   # Do not also set INSTALL_FFMPEG manually in service.env.
   installFfmpeg: true
 
+  # OpenTelemetry env defaults for the retriever service container.
+  otel:
+    enabled: false
```
jdye64: Since Zipkin is enabled by default we should enable OTEL as well. Otherwise we should disable Zipkin by default as well.

jdye64 (changes requested):
looks good overall. Had single comment about defaults.

nemo_retriever/helm/values.yaml
charlesbluca: +1 - this comes down to my own misconception of OTel being enabled by default, will switch to match up with current defaults

charlesbluca: Revisiting this, because it's a bit more nuanced than I originally thought

My understanding is that by default, we've got:
- OTel/Zipkin charts enabled (getting installed on overarching chart installation)
- NIM & NRL service OTel instrumentation disabled (this flag is referring to NRL instrumentation enablement)

So, in terms of OTel/Zipkin chart installation, things are consistent right now - is that what you were talking about? Or were you thinking about flipping the default for NRL / NIM instrumentation, so that by default all running services have OTel instrumentation enabled, and are generating traces to be collected?

PR #2265, opened 2026-06-24
Add LanceDB retrieval mode autodetection

jioffe502:
## Summary
- Add LanceDB table capability detection so local query auto-routes dense, hybrid, and sparse tables without mode flags.
- Add expert `--retrieval-mode auto|dense|hybrid|sparse` for explicit query overrides; default `auto` preserves artifact-driven routing.
- Add public ingest `--index-mode dense|hybrid|sparse` with dense as the default, plus hidden deprecated `--hybrid` and `--sparse` ingest aliases for compatibility.
- Keep deprecated query `--hybrid` as a hidden compatibility alias for `--retrieval-mode hybrid`.
- Add opt-in sparse LanceDB ingest/query support while preserving query output shape.
- Clean up the retrieval-mode diff by removing an unused capability wrapper and deduplicating repeated test fakes.

## Validation
- Focused changed-test set: `UV_PROJECT_ENVIRONMENT=/tmp/nemo-retriever-retrieval-mode-venv uv run --project nemo_retriever pytest nemo_retriever/tests/test_root_cli_workflow.py nemo_retriever/tests/test_root_query_cli.py nemo_retriever/tests/test_query_workflow_options.py nemo_retriever/tests/test_lancedb_capabilities.py nemo_retriever/tests/test_retriever_queries.py` -> 133 passed.
- Live CLI e2e with temp text docs, real `retriever ingest local`, real LanceDB tables, and a local OpenAI-compatible embedding endpoint covered: `--index-mode dense` with auto/dense query, `--index-mode hybrid` with auto/hybrid/dense/sparse query, and `--index-mode sparse` with auto/sparse query. Each returned `alpha safety manual` as the top hit and detected the expected table capabilities.
- Ruff passed on PR-changed Python files.
- Flake8 passed on PR-changed Python files.
- Pre-commit passed on PR-changed files: trailing whitespace, EOF, large files, Python AST, debug statements, black, flake8.
- `git diff --check` passed.

### DIFF
--- a/nemo_retriever/src/nemo_retriever/cli/ingest/options.py
+++ b/nemo_retriever/src/nemo_retriever/cli/ingest/options.py
@@ -209,11 +209,18 @@ OverwriteOption = Annotated[
 HybridOption = Annotated[
     bool,
     typer.Option(
-        "--hybrid/--no-hybrid",
+        "--hybrid",
         help=(
             "Also build a full-text (BM25) index over ingested text so query --hybrid can fuse "
             "lexical and vector retrieval. Disabled by default."
         ),
+    ),
+]
+SparseOption = Annotated[
+    bool,
+    typer.Option(
+        "--sparse",
+        help="Skip dense embedding and build a LanceDB full-text-search-only table. Disabled by default.",
     ),
 ]
 RayAddressOption = Annotated[
--- a/nemo_retriever/src/nemo_retriever/cli/ingest/options.py
+++ b/nemo_retriever/src/nemo_retriever/cli/ingest/options.py
@@ -10,6 +10,7 @@ import typer
 
 from nemo_retriever.ingest.plan import (
     AudioSplitTypeValue,
+    IngestIndexModeValue,
     IngestProfileValue,
     LocalIngestEmbedBackendValue,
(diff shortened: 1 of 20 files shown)


nemo_retriever/src/nemo_retriever/cli/ingest/options.py
```diff
@@ -209,13 +209,20 @@
 HybridOption = Annotated[
     bool,
     typer.Option(
-        "--hybrid/--no-hybrid",
+        "--hybrid",
```
edknv, line 223: Removing `--no-hybrid` makes sense, but should we add `--dense` to mirror `--sparse`? Maybe there's a use case where the user wants to force dense-only retrieval on a hybrid index.

edknv (approved):

nemo_retriever/src/nemo_retriever/cli/ingest/options.py
jioffe502: @edknv I actually just removed both --hybrid and --sparse too and added a --index-mode that defaults to dense and a --retrieval-mode that defaults to auto but can also allow us to select any of them.

PR #2273, opened 2026-06-26
docs: add agentic retrieval README examples

mahikaw:
Document CLI and Python examples for agentic retrieval with hosted NVIDIA endpoints so users can smoke test against an existing LanceDB index.

## Checklist

### DIFF
--- a/nemo_retriever/README.md
+++ b/nemo_retriever/README.md
@@ -309,6 +309,84 @@ Answer:
 Cat is the animal whose activity (jumping onto a laptop) matches the location of the typos, so the cat is responsible for the typos in the documents.
 ```
 
+### Run agentic retrieval
+
+Agentic retrieval runs an LLM-driven ReAct loop over an existing LanceDB index.
+It does not ingest documents; first build the index with one of the ingestion
+flows above, then query the same `lancedb_uri`, `table_name`, and embedding
+model.
+
+For [build.nvidia.com](https://build.nvidia.com/) hosted inference, set
+`NVIDIA_API_KEY`. The hosted API uses `integrate.api.nvidia.com` endpoints:
+
+```bash
+export NVIDIA_API_KEY=nvapi-...
+
+retriever query "Given their activities, which animal is responsible for the typos in my documents?" \
+  --agentic \
+  --agentic-llm-model nvidia/llama-3.3-nemotron-super-49b-v1.5 \
+  --agentic-invoke-url https://integrate.api.nvidia.com/v1/chat/completions \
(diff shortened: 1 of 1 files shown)


nemo_retriever/README.md
```diff
@@ -309,6 +309,87 @@ Answer:
...
+results = agentic_query_documents(
+    QueryRequest(
+        query="What is RAG?",
+        retrieval=QueryRetrievalOptions(top_k=10),
+        storage=QueryStorageOptions(
+            lancedb_uri="lancedb",
+            table_name="nemo-retriever",
+        ),
+        embed=QueryEmbedOptions(
+            embed_invoke_url="https://integrate.api.nvidia.com/v1/embeddings",
```
randerzander: why do we need to set the embed url explicitly?
```suggestion
            
```

it's not necessary to set for it to work out of the box, right?

nemo_retriever/README.md
```diff
@@ -309,6 +309,87 @@ Answer:
...
+            embed_model_name="nvidia/llama-nemotron-embed-1b-v2",
+        ),
+        agentic=QueryAgenticOptions(
+            enabled=True,
+            llm_model="nvidia/llama-3.3-nemotron-super-49b-v1.5",
+            invoke_url="https://integrate.api.nvidia.com/v1/chat/completions",
```
randerzander: same q

nemo_retriever/README.md
```diff
@@ -309,6 +309,87 @@ Answer:
...
+            lancedb_uri="lancedb",
+            table_name="nemo-retriever",
+        ),
+        embed=QueryEmbedOptions(
+            embed_invoke_url="https://integrate.api.nvidia.com/v1/embeddings",
+            embed_model_name="nvidia/llama-nemotron-embed-1b-v2",
+        ),
+        agentic=QueryAgenticOptions(
+            enabled=True,
+            llm_model="nvidia/llama-3.3-nemotron-super-49b-v1.5",
```
randerzander, line 383: same q

shouldn't there be a builtin default?

randerzander (changes requested):

nemo_retriever/README.md
mahikaw: that's correct, removing explicit flag setting

nemo_retriever/README.md
mahikaw: as of right now, it's a required flag. here's the draft pr where its getting removed: #2274 
will update this too with it.

randerzander (approved):

PR #2274, opened 2026-06-26
Agentic retriever enhancements

mahikaw:
## Description

  Adds agentic retrieval support for root query and harness BEIR evaluation, with opt-in structured tracing for debugging agent runs.

  Key changes:

  - Wires agentic retrieval through retriever query --agentic and harness query.agentic.
  - Adds shared agentic config construction from QueryRequest.
  - Supports BEIR agentic evaluation in harness with shared load/score artifacts.
  - Preserves retrieved hit metadata in agentic results.
  - Adds --agentic-trace / query.agentic_trace for opt-in JSONL traces.
  - Keeps default operator logs sanitized at lifecycle boundaries.
  - Normalizes retrieval knobs consistently across dense and agentic paths.

## Checklist


nemo_retriever/src/nemo_retriever/query/agentic_trace.py
```diff
@@ -0,0 +1,86 @@
...
+            event = {
+                "time": utc_now(),
+                "sequence": self._sequence,
+                **dict(payload),
+            }
+            with self.path.open("a", encoding="utf-8") as handle:
+                handle.write(json.dumps(_jsonable(event), sort_keys=False) + "\n")
+
+
+def _jsonable(value: Any) -> Any:
```
jperez999, line 121: Are you really wanting to dump "None"  into your logs?

nemo_retriever/src/nemo_retriever/query/agentic_trace.py
```diff
@@ -0,0 +1,86 @@
...
+        if len(items) > _TRACE_COLLECTION_LIMIT:
+            out["_truncated_items"] = len(items) - _TRACE_COLLECTION_LIMIT
+        return out
+    if isinstance(value, (list, tuple, set)):
+        items = list(value)
+        out = [_jsonable(item) for item in items[:_TRACE_COLLECTION_LIMIT]]
+        if len(items) > _TRACE_COLLECTION_LIMIT:
+            out.append({"_truncated_items": len(items) - _TRACE_COLLECTION_LIMIT})
+        return out
+    return str(value)
```
jperez999, line 141: This entire files seems more logging specific than agentic logging specific. Perhaps this should be in move to tools so that it can be re-used in other places where we want to do dumps?

nemo_retriever/src/nemo_retriever/query/workflow.py
```diff
@@ -35,6 +35,12 @@ def _strategies_for_retrieval_mode(mode: LanceRetrievalMode | None) -> list[str]
     return ["semantic"]
 
 
+def _normalize_content_types(content_types: Any) -> str | None:
```
jperez999: please reuse functions that exist already. Make sure you have your agent check for helpers that already exist, this seems to do what is already here https://github.com/NVIDIA/NeMo-Retriever/blob/d29ddafd31a666cc3f94019cbd99eef02ef96410/nemo_retriever/src/nemo_retriever/common/vdb/lancedb_schema.py#L135. Probably would be good to move all these normalizations to a specific file.

jperez999 (changes requested):

nemo_retriever/src/nemo_retriever/query/agentic_trace.py
mahikaw: `tools` seems more eval/benchmarks focused. would `nemo_retriever/src/nemo_retriever/common/io` or similar be a better home for this eventually when this kind of logging is needed in another spot? right now things like truncation, event policies and out paths are mostly agentic specific.

nemo_retriever/src/nemo_retriever/query/workflow.py
mahikaw: refactoring to utilize existing normalisations

nemo_retriever/src/nemo_retriever/harness/beir_runner.py
```diff
@@ -219,6 +219,16 @@ def _agentic_retrieve(
     from nemo_retriever.query.agentic_options import agentic_target_top_k
     from nemo_retriever.query.workflow import build_agentic_config
 
+    if query_request.agentic.trace_enabled:
+        trace_path = query_request.agentic.trace_path or str(writer.path("agentic_trace.jsonl"))
+        query_request = replace(query_request, agentic=replace(query_request.agentic, trace_path=trace_path))
+        writer.event(
```
jperez999: Is every line going to have this prepended? I think really the only thing necessary here from the three hardcoded pieces is query. I feel the other two are obvious? Thoughts? Am I missing something as to why you added agentic_trace_enabled and writing agentic trace?

nemo_retriever/src/nemo_retriever/operators/graph_ops/react_agent_operator.py
```diff
@@ -442,6 +429,25 @@ def _build_extra_body(self) -> Optional[Dict[str, Any]]:
             extra["reasoning_effort"] = self._reasoning_effort
         return extra or None
 
+    def _emit_trace(self, event: str, **payload: Any) -> None:
```
jperez999: You have multiple of these, this needs to be a decorator you add to the functions where you want it, because the logic is the same its just the report location that is different. We should share logic.

nemo_retriever/src/nemo_retriever/operators/graph_ops/selection_agent_operator.py
```diff
@@ -229,6 +223,25 @@ def __init__(
...
 
+    def _emit_trace(self, event: str, **payload: Any) -> None:
+        if self._trace_event is None:
+            return
+        try:
+            self._trace_event({"event": event, "operator": "selection_agent", **payload})
+        except Exception as exc:
+            logger.warning("SelectionAgentOperator: trace event failed: %s", exc, exc_info=True)
+
+    def _trace_documents(self, docs: List[Dict[str, Any]]) -> List[Dict[str, Any]]:
```
jperez999: same here. Also, posing question, can this be done using the python logging module?

nemo_retriever/src/nemo_retriever/query/agentic_trace.py
```diff
@@ -0,0 +1,86 @@
...
+
+def make_agentic_trace_path(root: str | Path | None = None) -> Path:
+    """Return a unique artifact trace path for one root CLI agentic query run."""
+    trace_root = Path(root).expanduser() if root is not None else _TRACE_ROOT
+    stamp = datetime.now(timezone.utc).strftime("%Y%m%dT%H%M%S_%fZ")
+    run_dir = f"{stamp}_{os.getpid()}_{uuid4().hex[:8]}"
+    return trace_root / run_dir / TRACE_FILE_NAME
+
+
+class AgenticTraceWriter:
```
jperez999: How does this differ from the capabilities the logging module affords us via python?

jperez999 (changes requested):

nemo_retriever/src/nemo_retriever/harness/beir_runner.py
mahikaw: fair enough, dropping this writer.event block

mahikaw: since we need all 3 if we use that event structure.

nemo_retriever/src/nemo_retriever/query/agentic_trace.py
mahikaw: refactored this into dedicated logger per run.

nemo_retriever/src/nemo_retriever/operators/graph_ops/react_agent_operator.py
mahikaw: makes sense. shared logic in a helpers since these are used in multiple spots

nemo_retriever/src/nemo_retriever/operators/graph_ops/selection_agent_operator.py
mahikaw: added helpers and switched to logging module

PR #2279, opened 2026-06-29
Reduce service ingest result payloads

ChrisJar:
## Checklist


nemo_retriever/src/nemo_retriever/operators/graph_ops/store_operator.py
```diff
@@ -104,7 +182,8 @@ def _store_row_images(
     strip_base64: bool = True,
 ) -> pd.DataFrame:
     """Return a copy of *df* with ``_stored_image_uri`` set for stored rows."""
-    if df.empty or ("_image_b64" not in df.columns and "page_image" not in df.columns):
+    image_columns = ("page_image", "images", "tables", "charts", "infographics", "table", "chart", "infographic")
+    if df.empty or ("_image_b64" not in df.columns and not any(column in df.columns for column in image_columns)):
```
charlesbluca: Is it possible that video frames with a row shape like
```
  {
      "image_b64": "...",
      "_content_type": "video_frame",
  }
```

Could get missed in this check since we aren't transforming them to have a `_image_b64` column in [`build_graph`](https://github.com/ChrisJar/nv-ingest/blob/1ced5cf23b54b246333c40bc0be66dd49009e151/nemo_retriever/src/nemo_retriever/graph/ingestor_runtime.py#L894-L905)?

ChrisJar: Yep you're right just patched

nemo_retriever/src/nemo_retriever/ingestor/results.py
```diff
@@ -38,19 +67,254 @@ def sanitize_cell_value(val: Any) -> Any:
...
+    if key in _EMBEDDING_FIELD_NAMES:
+        return None
+    if key in _EMBEDDING_PAYLOAD_COLUMNS and not isinstance(val, dict):
+        return None
+    if isinstance(val, dict):
+        return {str(k): _sanitize_result_value(str(k), v) for k, v in val.items()}
+    if isinstance(val, list):
+        return [_sanitize_result_value("", item) for item in val]
+    if isinstance(val, tuple):
+        return [_sanitize_result_value("", item) for item in val]
```
charlesbluca: Should we pass these results into `sanitize_cell_value` instead of returning early here? Seems like there's some list/tuple truncation happening there that falls in line with the memory reduction this function is aiming for

charlesbluca (approved):

nemo_retriever/src/nemo_retriever/ingestor/results.py
ChrisJar: Good call

PR #2283, opened 2026-06-30
LLM Based Task

KyleZheng1284:
## Description

- Adds a reusable TextGenerationOperator built on the existing operator framework.
- Introduces typed tasks for QA, summarization, and generic prompt generation.
- Preserves existing QA APIs, schemas, prompts, and client compatibility.
- Improves secure graph serialization without persisting API credentials.
- Preserves sampling overrides and DataFrame value types across execution.
- Keeps embedding and captioning as separate specialized operator families.
- Adds tests for generation, persistence, validation, and backward compatibility

## Checklist


nemo_retriever/src/nemo_retriever/common/params/models.py
```diff
@@ -48,29 +189,95 @@ class _ParamsModel(BaseModel):
       so no downstream consumer needs changes.
     """
 
-    model_config = ConfigDict(extra="forbid")
+    model_config = ConfigDict(extra="forbid", hide_input_in_errors=True)
+
+    # Keep the explicit no-auth intent after NO_API_KEY is normalized to
```
mahikaw, line 194: nit: do we want to update it here too for the api_key=None behaviour and gen error codes `nemo_retriever/src/nemo_retriever/tools/evaluation/README.md`

nemo_retriever/src/nemo_retriever/common/params/models.py
```diff
@@ -618,6 +849,131 @@ def _check_timeout(cls, v: float) -> float:
...
+        for name in self.model_fields_set:
+            value = getattr(self, name)
+            values[name] = value
+        return LLMInferenceParams(**values)
+
+
+_SAMPLING_UNSET = object()
+
+
+class TextGenerationParams(_ParamsModel):
```
mahikaw, line 916: another nit: do you think it might be useful for non-secret generation config like temperature, top_p, max_tokens, and reasoning_enabled be emitted into eval metadata/traces so runs are easier to compare (if that's the goal with this infra)?

nemo_retriever/src/nemo_retriever/models/llm/tasks/base.py
```diff
@@ -0,0 +1,206 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-26, NVIDIA CORPORATION & AFFILIATES.
+# All rights reserved.
```
mahikaw, line 2: if this is envisioned to be generation specific base let's rename it to something more reflective like TextGenerationTask

mahikaw (approved):
looks good, added some comments that might lead to minor refactoring.

jioffe502 (approved):

PR #2286, opened 2026-06-30
[harness] Preserve harness artifact values and write exit codes

jioffe502:
## Summary

- preserve legitimate token-limit settings such as `caption_max_tokens` and `text_chunk_overlap_tokens` while continuing to redact API keys, tokens, credentials, and webhook values
- classify `run.log` and TREC artifact I/O failures through the harness's existing artifact-write error path and exit code 30
- rely on the artifact writer's startup cleanup instead of repeating raw `query_results.jsonl` unlinks

## Why this remains after #2290

#2290 absorbed the broader artifact-safety work: atomic JSON publication, preflight-before-cleanup, portable relative manifests, and exit-30 handling for JSON and session writes. This PR intentionally keeps that contract unchanged.

Two concrete gaps remained on merged `main`:

1. Redaction matched the substring `token`, so reproducibility values such as `max_tokens` were serialized as `"<redacted>"`.
2. Raw text writes could bypass the artifact-write classifier. A TREC/run-log `OSError` could therefore surface as internal exit code 70 instead of artifact exit code 30.

## Scope boundaries

This PR does not change the artifact schema, session layout, output-directory reuse policy, manifest format, Slack reporting, or benchmark execution behavior introduced by #2290. It adds no new writer abstraction.

## Validation

- `uv run pytest -q tests/test_harness*.py tests/test_beir_evaluation.py` — 72 passed
- pre-commit on all four changed files — passed
- real `jp20_smoke --dry-run --json` artifact check:
  - exit code 0
  - reranker secret absent from stdout and emitted JSON
  - `caption_max_tokens: 77` and `text_chunk_max_tokens: 123` remained numeric in resolved/planning artifacts


nemo_retriever/src/nemo_retriever/harness/artifact_writer.py
```diff
@@ -121,12 +127,22 @@ def capture_output_to_log(path: Path, *, label: str):
...
-    return any(token in normalized for token in _SENSITIVE_KEY_PARTS)
+    snake_case = re.sub(r"([a-z0-9])([A-Z])", r"\1_\2", str(value))
+    normalized = re.sub(r"[^a-z0-9]+", "_", snake_case.lower()).strip("_")
+    if any(marker in normalized for marker in _SENSITIVE_KEY_MARKERS):
+        return True
+    parts = normalized.split("_")
+    if "token" in parts:
+        return True
+    return any(
+        part == "tokens" and index > 0 and parts[index - 1] in _TOKEN_CREDENTIAL_QUALIFIERS
```
charlesbluca, line 143: This only checks the segment immediately before plural `tokens`, so compound credential keys such as `service_account_tokens` and `oauth_app_tokens` remain unredacted; bare `tokens` does as well. Since these artifacts may be replayed or shared, could we make plural-token handling fail closed while explicitly preserving known budget fields such as `max_tokens` and `overlap_tokens`? Please add compound-key and structured-override regressions.

ChrisJar (approved):
lgtm apart from existing comments

PR #2290, opened 2026-07-01
feat(harness): add portable runfile sessions and replayable Slack reports

jioffe502:
## TL;DR

Add a portable, artifact-first workflow for running one or more checked-in Retriever benchmarks and optionally reporting the completed artifacts to Slack.

Users and agents can copy the dataset path-map example, point it at datasets available on their machine, run one runfile or the four-dataset library suite, inspect a compact terminal summary, and preview or post the same artifacts afterward. Harness execution never implicitly contacts Slack. This PR does not add recurring scheduling or machine-owned nightly infrastructure.

### Portable benchmark sessions

- Add `retriever harness run-files` for one or more JSON/YAML runfiles.
- Apply an untracked, machine-local `--dataset-paths` map after runfile defaults and before CLI overrides.
- Give each session a stable layout with `expanded_runs.json`, `session_summary.json`, and one child artifact directory per run.
- Preserve each runfile's mode: JP20 remains local; the larger BO767, Earnings, FinanceBench, and ViDoRe runfiles use batch ingest.
- Preflight every run once before the session starts, then complete the remaining runs after an individual runtime failure while returning the first nonzero exit code.
- Keep dry-run behavior session-wide so one session cannot mix planned and executed children.

### Smaller artifact contract

- Publish JSON artifacts atomically so readers never observe a partially written payload.
- Poll `status.json` while a run is active.
- Read `results.json` when one run is terminal.
- Read `session_summary.json` when a multi-run session is terminal.
- Keep resolved plans, environment, full logs, per-query results, BEIR outputs, and LanceDB as separate drill-down evidence referenced by relative paths.
- Stop emitting a duplicate `summary_metrics.json` for new runs while retaining read compatibility in `diff` and Slack replay.
- Keep terminal failures concise and full tracebacks in `run.log`.
- Validate configuration before replacing known outputs in a reused artifact directory.

### Post-hoc Slack reporting

- Add `retriever harness post-slack` for a completed session or one or more run artifacts.
- Add `--preview`, which renders the exact payload without reading `SLACK_WEBHOOK_URL` or making a network request.
- Keep webhook credentials in the process environment, redact secret-like override values, and avoid exposing webhook values in transport failures.
- Omit runner-local artifact paths unless `--artifact-paths` is explicitly requested.
- Reject empty or malformed session summaries instead of silently posting a misleading report.
- Give raw-artifact replay a deterministic identity so preview and post render the same payload.
- Send a new message per invocation without mutating or rerunning the source artifacts.

### Repeated-query reliability

- Reuse the cached local query graph and its lazily resolved operator delegates so BEIR queries retain one embedding delegate instead of rebuilding the embedder and re-contacting Hugging Face for every query.
- Include the underlying embedding failure when vector extraction fails so query errors remain actionable.

## User-facing boundaries

- Use `retriever ingest` and `retriever query` for direct product workflows.
- Use `retriever harness run-files` for portable benchmark/evaluation sessions.
- Use `retriever harness post-slack` only as optional artifact post-processing.
- Treat `--dry-run` as configuration/planning preflight, not runtime evidence; use a real small benchmark when model startup or execution behavior changes.
- Scheduling, deployment, locking, retries, retention, and secret distribution remain outside this PR.
- Benchmark corpora and qrels are not distributed or downloaded by the harness.

## Suggested review order

1. **CLI and session resolution:** `harness/cli.py`, `harness/dataset_paths.py`, `harness/runsets.py`, and the checked-in runfiles.
2. **Artifact contract:** `harness/artifact_writer.py`, `harness/execution.py`, `harness/environment.py`, and `harness/diff.py`.
3. **Slack replay:** `harness/slack.py` and `test_harness_slack.py`.
4. **Query reuse and error propagation:** `graph/pipeline_graph.py`, `graph/retriever.py`, `operators/vdb.py`, and their focused tests.
5. **User contract:** `harness/README.md`, `EXPECTED_RESULTS.md`, and `HANDOFF.md`.

The commits already follow that dependency/order boundary; no history rewrite is needed.

## Validation

- `177 passed`: artifact/session/Slack contract, runfile sessions, graph execution, and repeated-query caching.
- `128 passed`: agentic, query workflow, root query CLI, comparator, and service-query regressions.
- Pre-commit passes on the final changed-file set.
- A dry run against tracked `data/multimodal_test.pdf` wrote status/result/planning artifacts, created no LanceDB, run log, or query output, and produced a valid Slack preview with no webhook configured.
- A four-run portable dry-run applied local mode to JP20 and batch mode to the three larger datasets, produced relative manifests, and previewed Slack with `SLACK_WEBHOOK_URL` unset.
- A real JP20 batch smoke processed 20 files / 1,940 pages, wrote a 22 MB LanceDB index, and emitted the compact terminal contract.
- Missing-input and metric-gate failures emitted concise terminal results.
- Current readers previewed a pre-change session and diffed current results against legacy `summary_metrics.json` artifacts.
- An earlier full eight-GPU all-batch suite completed with `success=true`, `all_passed=true`, and exit code `0`:
  - JP20: 115/115 queries
  - BO767: 991/991 queries
  - Earnings: 628/628 queries
  - FinanceBench: 150/150 queries
- That completed suite posted one real Slack report successfully.

## Known follow-up

Two oversized BO767 chunks exceeded the embedder context limit during the full run. The session and query gates passed; input truncation/splitting is separate follow-up work.

## Checklist

### DIFF
--- a/nemo_retriever/src/nemo_retriever/harness/artifact_writer.py
+++ b/nemo_retriever/src/nemo_retriever/harness/artifact_writer.py
@@ -17,7 +17,7 @@ import traceback
 from typing import Any, Mapping
 
 from nemo_retriever.harness.contracts import FailurePayload, PHASE_VALUES, STATUS_VALUES
-from nemo_retriever.harness.json_io import jsonable, write_json
+from nemo_retriever.harness.json_io import artifact_write_error, jsonable, write_json
 
 _ARTIFACT_NAMES = {
     "status": "status.json",
@@ -41,9 +41,12 @@ def utc_now() -> str:
 
 
 def append_jsonl(path: Path, payload: Mapping[str, Any]) -> None:
-    path.parent.mkdir(parents=True, exist_ok=True)
-    with path.open("a", encoding="utf-8") as handle:
-        handle.write(json.dumps(jsonable(payload), sort_keys=False) + "\n")
+    try:
+        path.parent.mkdir(parents=True, exist_ok=True)
+        with path.open("a", encoding="utf-8") as handle:
+            handle.write(json.dumps(jsonable(payload), sort_keys=False) + "\n")
+    except Exception as exc:
+        raise artifact_write_error(exc) from exc
 
 
 def append_text(path: Path, text: str) -> None:
--- a/nemo_retriever/src/nemo_retriever/harness/artifact_writer.py
+++ b/nemo_retriever/src/nemo_retriever/harness/artifact_writer.py
@@ -88,12 +88,19 @@ def capture_output_to_log(path: Path, *, label: str):
             os.close(saved_stdout)
 
 
+_SENSITIVE_KEY_PARTS = ("api_key", "password", "secret", "credential", "token")
+
+
+def _is_sensitive_key(value: Any) -> bool:
+    normalized = str(value).lower()
+    return any(token in normalized for token in _SENSITIVE_KEY_PARTS)
+
+
 def redact(value: Any) -> Any:
     if isinstance(value, dict):
         out: dict[str, Any] = {}
         for key, nested in value.items():
-            normalized = str(key).lower()
-            if any(token in normalized for token in ("api_key", "password", "secret", "credential", "token")):
+            if _is_sensitive_key(key):
                 out[key] = "<redacted>" if nested else nested
             else:
                 out[key] = redact(nested)
@@ -102,6 +109,10 @@ def redact(value: Any) -> Any:
         return [redact(item) for item in value]
     if isinstance(value, tuple):
         return [redact(item) for item in value]
+    if isinstance(value, str) and "=" in value:
+        key, _separator, _raw = value.partition("=")
+        if _is_sensitive_key(key):
+            return f"{key}=<redacted>"
     return value
 
 
--- a/nemo_retriever/src/nemo_retriever/harness/artifact_writer.py
+++ b/nemo_retriever/src/nemo_retriever/harness/artifact_writer.py
@@ -10,12 +10,30 @@ import io
 import json
 import os
 from pathlib import Path
+import shutil
 import sys
 import tempfile
+import traceback
 from typing import Any, Mapping
 
 from nemo_retriever.harness.contracts import FailurePayload, PHASE_VALUES, STATUS_VALUES
 from nemo_retriever.harness.json_io import jsonable, write_json
+
+_ARTIFACT_NAMES = {
+    "status": "status.json",
+    "events": "events.jsonl",
+    "environment": "environment.json",
+    "runfile": "runfile.json",
+    "resolved_benchmark": "resolved_benchmark.json",
+    "ingest_plan": "ingest_plan.json",
+    "query_plan": "query_plan.json",
+    "log": "run.log",
+    "query_results": "query_results.jsonl",
+    "beir_metrics": "beir_metrics.json",
+    "beir_run": "beir_run.trec",
+    "lancedb": "lancedb",
+}
+_LEGACY_ARTIFACT_NAMES = ("summary_metrics.json",)
 
 
 def utc_now() -> str:
@@ -43,12 +61,19 @@ def capture_output_to_log(path: Path, *, label: str):
         stdout_fd, stderr_fd = sys.stdout.fileno(), sys.stderr.fileno()
     except (AttributeError, OSError, ValueError, io.UnsupportedOperation):
         append_text(path, "stdio capture unavailable in this runtime\n")
-        yield
-        append_text(path, f"## {utc_now()} {label} complete\n")
+        try:
+            yield
+        except BaseException:
+            append_text(path, traceback.format_exc())
+            append_text(path, f"## {utc_now()} {label} failed\n")
+            raise
+        else:
+            append_text(path, f"## {utc_now()} {label} complete\n")
         return
 
     saved_stdout = saved_stderr = buf = None
     failed = False
+    failure_traceback: str | None = None
     try:
         saved_stdout = os.dup(stdout_fd)
         saved_stderr = os.dup(stderr_fd)
@@ -65,6 +90,7 @@ def capture_output_to_log(path: Path, *, label: str):
                 os.dup2(saved_stderr, stderr_fd)
         except BaseException:
             failed = True
+            failure_traceback = traceback.format_exc()
             raise
         finally:
             if buf is not None:
@@ -74,6 +100,10 @@ def capture_output_to_log(path: Path, *, label: str):
                     if captured:
                         handle.write(captured)
                         if not captured.endswith(b"\n"):
+                            handle.write(b"\n")
+                    if failure_traceback:
+                        handle.write(failure_traceback.encode("utf-8", errors="replace"))
+                        if not failure_traceback.endswith("\n"):
                             handle.write(b"\n")
                     handle.write(f"## {utc_now()} {label} {'failed' if failed else 'complete'}\n".encode("utf-8"))
                 if failed and captured:
@@ -123,9 +153,13 @@ class ArtifactWriter:
         self.benchmark = benchmark
         self.started_at = utc_now()
         self.artifact_dir.mkdir(parents=True, exist_ok=True)
+        for name in (*_ARTIFACT_NAMES.values(), *_LEGACY_ARTIFACT_NAMES, "results.json"):
+            path = self.artifact_dir / name
+            if path.is_dir() and not path.is_symlink():
+                shutil.rmtree(path)
+            elif path.exists() or path.is_symlink():
+                path.unlink()
         self.events_path = self.artifact_dir / "events.jsonl"
-        if self.events_path.exists():
-            self.events_path.unlink()
 
     def path(self, name: str) -> Path:
         return self.artifact_dir / name
@@ -150,7 +184,7 @@ class ArtifactWriter:
         status: str,
         phase: str,
         failure: FailurePayload | None = None,
-        summary_metrics_path: Path | None = None,
+        results_path: Path | None = None,
     ) -> dict[str, Any]:
         if status not in STATUS_VALUES:
             raise ValueError(f"Invalid status: {status}")
@@ -164,7 +198,7 @@ class ArtifactWriter:
             "started_at": self.started_at,
             "updated_at": utc_now(),
             "artifact_dir": str(self.artifact_dir),
-            "summary_metrics_path": str(summary_metrics_path) if summary_metrics_path is not None else None,
+            "results_path": str(results_path.relative_to(self.artifact_dir)) if results_path is not None else None,
             "failure": failure.to_dict() if failure is not None else None,
         }
         write_json(self.path("status.json"), payload)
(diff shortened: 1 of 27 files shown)


nemo_retriever/src/nemo_retriever/harness/artifact_writer.py
```diff
@@ -102,6 +139,10 @@ def redact(value: Any) -> Any:
         return [redact(item) for item in value]
     if isinstance(value, tuple):
         return [redact(item) for item in value]
+    if isinstance(value, str) and "=" in value:
```
charlesbluca, line 145: P2: Structured override values can bypass this redaction. Runfile mapping values are JSON-serialized into strings, so `notifications={"webhook_url":"...SECRET..."}` reaches this branch with the non-sensitive key `notifications` and is written verbatim to `expanded_runs.json` before override validation. I reproduced this with the actual parser and redactor. Please redact structured right-hand values recursively, or apply redaction before flattening overrides into `KEY=VALUE` strings.

nemo_retriever/src/nemo_retriever/harness/runsets.py
```diff
@@ -80,34 +159,139 @@ def run_runset(
...
+                mode=str(expanded["mode"]),
+            )
+        )
+        if exit_code == EXIT_SUCCESS and outcome.exit_code != EXIT_SUCCESS:
+            exit_code = outcome.exit_code
+
+    session_summary = _session_summary(
+        session_type="runfiles",
+        exit_code=exit_code,
+        dry_run=any(bool(expanded["dry_run"]) for expanded in expanded_runs),
```
charlesbluca: P2: This collapses per-run dry-run state into a session-wide flag using `any(...)`. If one runfile is a dry run but another executes and fails, the session has `dry_run=true`; `build_slack_payload` then labels the overall report and every child as DRY RUN, hiding the real failure. Please preserve `dry_run` in each run summary and only mark the entire session dry-run when all children are dry-run (or render mixed sessions per run).

nemo_retriever/src/nemo_retriever/harness/slack.py
```diff
@@ -257,26 +365,30 @@ def _two_column_row_bold(left: str, right: str) -> list[dict[str, Any]]:
...
-    )
+    if report.dry_run:
+        overall_status = f"DRY RUN ({total_count} planned)"
+    elif report.all_passed:
+        overall_status = f"PASS ({passed_count}/{total_count})"
+    else:
+        overall_status = f"FAIL ({passed_count}/{total_count} passed)"
     first_metadata = next((run.run_metadata for run in report.results if run.run_metadata), {})
 
     rows: list[list[dict[str, Any]]] = []
```
charlesbluca, line 383: P2: This row list is unbounded and is ultimately emitted as one Slack table. Slack limits a table block to 100 rows (https://docs.slack.dev/reference/block-kit/blocks/table-block/), while `run-files` accepts an arbitrary number of runs; a sufficiently large session (often about nine runs with all default metrics) will therefore produce a payload Slack rejects. Please split large reports across bounded tables/messages or truncate with an explicit continuation summary.

nemo_retriever/tests/test_harness_runfiles.py
```diff
@@ -0,0 +1,191 @@
+import json
+from pathlib import Path
+
+import pytest
```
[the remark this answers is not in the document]
jioffe502: Fixed in `57fe0630` by adding the standard NVIDIA SPDX header to this new test file.

nemo_retriever/tests/test_harness_slack.py
```diff
@@ -0,0 +1,179 @@
+import json
+from pathlib import Path
+
+import pytest
+import requests
```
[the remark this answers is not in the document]
jioffe502: Fixed in `57fe0630` by adding the standard NVIDIA SPDX header to this new test file.

nemo_retriever/tests/test_harness_artifacts.py
```diff
@@ -0,0 +1,164 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-26, NVIDIA CORPORATION & AFFILIATES.
+# SPDX-License-Identifier: Apache-2.0
```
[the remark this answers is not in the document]
jioffe502: Fixed in `57fe0630` by adding the missing `All rights reserved.` line to the SPDX header.

nemo_retriever/src/nemo_retriever/harness/execution.py
```diff
@@ -141,18 +182,22 @@ def run_benchmark(
...
         parse_metric_gates(requirements)
         resolved = resolve_benchmark(benchmark, mode=mode, overrides=overrides)
-        dataset_path, _query_path = validate_dataset_inputs(resolved, dry_run=dry_run)
+        write_json(writer.path("resolved_benchmark.json"), redact(resolved))
         write_json(writer.path("environment.json"), collect_environment())
+        dataset_path, _query_path = validate_dataset_inputs(resolved, dry_run=dry_run)
 
         writer.status(status="running", phase="ingest_plan")
         ingest_request = build_ingest_request(resolved, dataset_path, writer.artifact_dir)
+        write_json(writer.path("resolved_benchmark.json"), redact(resolved))
```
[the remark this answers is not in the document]
jioffe502: Addressed in `57fe0630` by documenting why both writes are intentional: the first preserves the resolved config for validation failures, and the second records the final artifact-local storage paths after `build_ingest_request()`.

nemo_retriever/src/nemo_retriever/harness/artifact_writer.py
jioffe502: Fixed in `57fe0630`. Override redaction now parses structured right-hand values and recursively redacts nested sensitive keys before writing `expanded_runs.json`; I also added regression coverage through the actual runfile mapping serialization path.

nemo_retriever/src/nemo_retriever/harness/runsets.py
jioffe502: Addressed by the prior session dry-run fix. Dry-run ownership now stays at the session/run summary level so mixed sessions do not label real executions as dry runs; this thread is now outdated against the latest diff.

nemo_retriever/src/nemo_retriever/harness/slack.py
jioffe502: Fixed in `57fe0630` by bounding Slack table output to Slack’s 100-row limit and adding an explicit `TRUNCATED` row that points reviewers to the session artifacts for omitted rows. Added large-session coverage for the payload shape.

nemo_retriever/src/nemo_retriever/harness/runsets.py
```diff
@@ -80,34 +167,146 @@ def run_runset(
...
+                "artifact_dir": f"{index:03d}_{run_name}",
+                "mode": effective_mode,
+                "dataset_paths": dataset_paths.to_dict() if dataset_paths is not None else None,
+                "overrides": list(effective_overrides),
+                "requirements": list(effective_requirements),
+                "dry_run": bool(dry_run),
+            }
+        )
+
+    session_dir.mkdir(parents=True, exist_ok=True)
```
charlesbluca: P2: Reusing a session directory leaves the previous terminal summary visible during the new run.

After preflight succeeds, this code writes the new `expanded_runs.json` but does not remove an existing `session_summary.json`. Pollers or `post-slack` can therefore read the previous session success while the replacement session is still running; if execution later aborts, that stale summary remains indefinitely. This conflicts with `session_summary.json` being the authoritative terminal result.

Please remove or invalidate the old summary after all-session preflight succeeds and before starting child runs. A regression test should seed an old summary, start a rerun, and verify that it is not visible until the new terminal summary is written.

nemo_retriever/src/nemo_retriever/harness/execution.py
```diff
@@ -269,12 +330,13 @@ def run_benchmark(
         )
         return RunOutcome(exit_code=exc.exit_code, artifact_dir=writer.artifact_dir, results=result)
     except Exception as exc:
+        append_text(writer.path("run.log"), f"\n## {traceback.format_exc()}")
```
charlesbluca: P2: Artifact-write failures do not produce the documented exit code 30.

Exceptions while writing artifacts enter this generic internal-error path and are classified as exit code `70`; if writing the failure result also fails, they escape entirely. As a result, the documented `EXIT_ARTIFACT_WRITE_FAILURE = 30` does not appear reachable through this execution path.

Please either classify recoverable artifact-write failures as exit code `30` or remove/update the unsupported contract.

nemo_retriever/src/nemo_retriever/harness/runsets.py
```diff
@@ -80,34 +167,146 @@ def run_runset(
...
+                "dataset_paths_file": str(dataset_paths_file.expanduser().resolve()) if dataset_paths_file else None,
+                "runfiles": expanded_runs,
+            }
+        ),
+    )
+
+    run_results: list[dict[str, Any]] = []
+    exit_code = EXIT_SUCCESS
+    for expanded, request in zip(expanded_runs, requests, strict=True):
+        outcome = run_benchmark(
```
charlesbluca: P1: A raised child-run failure aborts the entire session.

`run_benchmark()` repeats `preflight_benchmark()` before its exception handler, so an input disappearing between the session preflight and execution raises here instead of returning a failed `RunOutcome`. The exception skips all remaining runs and prevents `session_summary.json` from being written, contradicting the advertised continue-remaining-runs and terminal-artifact behavior. The same problem affects `run_runset()`, and other setup failures before the `run_benchmark()` try block can follow this path.

Please either reuse the session preflight result or catch and convert per-run exceptions at the session-loop boundary. Add coverage where the initial preflights succeed but a later execution-time preflight raises, asserting that subsequent runs execute and the session summary is written.

nemo_retriever/src/nemo_retriever/harness/runsets.py
jioffe502: Fixed in `15150d19`. After all session preflight succeeds, reused session directories now remove any stale `session_summary.json` before child runs start; added regression coverage that seeds an old summary and asserts it is not visible during the replacement run.

nemo_retriever/src/nemo_retriever/harness/execution.py
jioffe502: Fixed in `15150d19`. Artifact-write failures during terminal result writing now return `EXIT_ARTIFACT_WRITE_FAILURE` (`30`), including the case where writing the failure result itself also fails. Added direct `run_benchmark()` regression coverage.

nemo_retriever/src/nemo_retriever/harness/runsets.py
jioffe502: Fixed in `15150d19` by catching per-child execution exceptions at the session-loop boundary and converting them into failed child outcomes, so remaining runs continue and `session_summary.json` is still written. Added coverage for both `run-files` and `runset`.

charlesbluca (approved):

PR #2292, opened 2026-07-02
Stabilize local vLLM DeepGEMM warmup startup

jioffe502:
## Summary

- Default NRL local vLLM startup to `VLLM_DEEP_GEMM_WARMUP=skip` via `os.environ.setdefault(...)`.
- Apply that default before every local NRL `vllm.LLM` constructor: embedding, VL rerank, captioning, and Nemotron Parse.
- Do not set `VLLM_USE_DEEP_GEMM=0` and do not hard-code `CUDA_HOME`; users can still opt into another vLLM warmup mode explicitly.

## Why

JP20 local harness runs have been failing during ingest before any rows are written. The failing artifact points to local embedding/vLLM startup, not caption or rerank:

- `results.json`: `exit_code: 10`, `failed_phase: ingest`
- `ingest_plan.json`: `caption: null`, local `nvidia/llama-nemotron-embed-1b-v2`, `local_ingest_embed_backend: "vllm"`
- `query_plan.json`: `rerank: false`

Original trace:

```text
RuntimeError: DeepGEMM backend is not available or outdated. Please install or update the `deep_gemm` to a newer version to enable FP8 kernels.
RuntimeError: Engine core initialization failed. See root cause above. Failed core proc(s): {}

File ".../llama_nemotron_embed_1b_v2_embedder.py", line 69, in _ensure_loaded
    self._llm = create_vllm_llm(
File ".../models/inference/vllm.py", line 87, in create_vllm_llm
    return LLM(**kwargs)
```

Downstream symptom:

```text
lancedb_write summary: total=3147 accepted=0 dropped_bad_length=3147 expected_dim=2048
```

## When DeepGEMM warmup can help

This change skips DeepGEMM's ahead-of-time kernel warmup; it does **not** disable runtime DeepGEMM. With warmup skipped, an eligible kernel is JIT-compiled when first encountered during model execution.

Warmup is useful when all of the following are true:

- The workload runs on Hopper or Blackwell with a compatible DeepGEMM installation.
- The model uses FP8 layers that vLLM actually routes through DeepGEMM.
- First-request latency or stable benchmark latency matters enough to move JIT compilation into startup.

NRL's default local embedding, VL rerank, Nemotron Parse, and caption paths use BF16, so they are not expected to benefit. The realistic NRL candidates are the optional local FP8 caption profiles, especially the FP8 Nemotron 3 Omni MoE model. For those deployments, operators can opt into `VLLM_DEEP_GEMM_WARMUP=relax` first, or `full` when every possible GEMM shape must be compiled before accepting work.

## Review question

Is skipping optional DeepGEMM warmup the right default for NRL local startup reliability, while letting Hopper/Blackwell performance owners opt into `VLLM_DEEP_GEMM_WARMUP=full` or another upstream-supported mode?

## Validation

- `uv run --project nemo_retriever pytest nemo_retriever/tests/test_vllm_embed.py nemo_retriever/tests/test_nemotron_rerank_vl_v2.py nemo_retriever/tests/test_caption_model_profiles.py nemo_retriever/tests/test_nemotron_parse_v1_2.py -q`
  - `100 passed, 2 warnings`
- Rebased onto current `main` with Typer 0.27 help support; combined CLI-contract and DeepGEMM-focused suite: `105 passed, 2 warnings`
- `python -m compileall` on changed Python files
- `git diff --check`
- Live local embedding smoke with CUDA/DeepGEMM env vars unset:

```text
warmup skip
shape (1, 2048)
dtype torch.float32
```


jioffe502:
To discuss with Julio and Jeremy

jdye64:
I have no problem with this.

jdye64 (approved):

PR #2296, opened 2026-07-02
Add legacy result payload flags

ChrisJar:
## Checklist

### DIFF
--- a/nemo_retriever/src/nemo_retriever/ingestor/results.py
+++ b/nemo_retriever/src/nemo_retriever/ingestor/results.py
@@ -86,6 +86,33 @@ def _sanitize_returned_payload(val: Any) -> Any:
     return sanitize_cell_value(val)
 
 
+def _contains_requested_payload(val: Any, *, return_embeddings: bool, return_images: bool) -> bool:
+    if isinstance(val, dict):
+        for key, nested in val.items():
+            str_key = str(key)
+            if return_images and str_key in _RAW_IMAGE_FIELD_NAMES:
+                return True
+            if return_embeddings and str_key in _EMBEDDING_FIELD_NAMES:
+                return True
+            if _contains_requested_payload(
+                nested,
+                return_embeddings=return_embeddings,
+                return_images=return_images,
+            ):
+                return True
+        return False
+    if isinstance(val, (list, tuple)):
+        return any(
+            _contains_requested_payload(
+                item,
+                return_embeddings=return_embeddings,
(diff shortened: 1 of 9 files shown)


nemo_retriever/src/nemo_retriever/service/service_ingestor.py
```diff
@@ -516,6 +516,9 @@ def _pipeline_payload(self, *, result_schema: ResultSchema = "legacy") -> dict[s
         """
         spec = dict(self._pipeline_spec)
         spec["result_schema"] = result_schema
+        if result_schema == "legacy":
+            spec.setdefault("return_embeddings", False)
```
charlesbluca: **Expose these flags through the public client API**

`setdefault` preserves these values only if something has already written them into the private `_pipeline_spec` dictionary. Currently, neither `IngestExecuteParams` nor `_resolve_execute_flags` accepts `return_embeddings` or `return_images`, so `ingest(return_images=True)` is silently ignored. The new test works around this by mutating `_pipeline_spec` directly.

Could we expose these as supported execute-time options, thread them into `_pipeline_payload`, and test the public `ingest(...)` path? This is blocking because the primary client cannot currently use the feature this PR introduces.

nemo_retriever/src/nemo_retriever/ingestor/results.py
```diff
@@ -76,17 +101,47 @@ def _sanitize_result_value(key: str, val: Any) -> Any:
...
+                str(k),
+                v,
+                return_embeddings=return_embeddings,
+                return_images=return_images,
+            )
+            for k, v in val.items()
+        }
     if isinstance(val, list):
-        return sanitize_cell_value([_sanitize_result_value("", item) for item in val])
+        return sanitize_cell_value(
```
charlesbluca: **Requested nested payloads can still be summarized here**

After recursively preserving `image_b64`, the outer list is passed to `sanitize_cell_value`, which replaces lists longer than 20 items with `"<list len=N>"`. For example, an `images` column containing 21 image dictionaries still loses every image even with `return_images=True`.

Could we avoid container-size summarization when the container holds explicitly requested payloads, and add a test with more than 20 nested images?

charlesbluca (changes requested):
Requesting changes because the new payload-return options are not reachable through the supported ServiceIngestor API. I also found a nested-collection case where explicitly requested images are still discarded by size summarization.

charlesbluca (approved):

PR #2319, opened 2026-07-08
Add configurable auto retrieval to the service query path

mahikaw:
## Description
Service /v1/query (and MCP, which proxies to it) now auto-detects dense vs hybrid retrieval from the LanceDB table’s existing indexes — same contract as library/CLI auto mode. A table with vector + FTS runs hybrid (vector + BM25), vector-only runs dense. No vectordb.retrieval_mode config — query is read-only and never builds or mutates indexes. FTS must exist from ingestion.

Evidence responses report strategies_used: ["dense"] or ["hybrid"] based on what actually ran.

## Changes
Query path: inspect table via inspect_lancedb_table_object, resolve mode, route through LanceDB.retrieval() with hybrid when FTS is present
Write path: persist rows only; use list_tables() for create-vs-append to avoid accidental overwrite on retry
Health: expose effective_retrieval_mode (dense / hybrid / unknown) when a table exists
Config: add vectordb stanza to bundled retriever-service.yaml (URI, table, query embed model, URL) — no retrieval-mode knob
Docs: update MCP query tool description and gateway /query summary string
Tests: unit tests (mocked) + @pytest.mark.integration tests for auto-detect dense/hybrid from FTS presence
Design notes (post-review)
Removed initial dense / hybrid / auto server config and post-ingest FTS backfill — index format is an ingestion-time contract
vectordb.embed_model in YAML is deploy-time config for the VectorDB pod’s query embedder (overridden via Helm in production), not per-request

## Checklist


nemo_retriever/src/nemo_retriever/service/vectordb_app.py
```diff
@@ -161,6 +170,69 @@ def embed_mode(self) -> str:
...
+        table = self._db.open_table(self.table_name)
+        return inspect_lancedb_table_object(table)
+
+    def resolve_effective_retrieval_mode(self) -> LanceRetrievalMode:
+        """Resolve the configured retrieval mode against table capabilities."""
+        if not self._table_exists:
+            return "dense"
+
+        caps = self._table_capabilities()
+        assert caps is not None
```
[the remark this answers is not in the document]
jioffe502: +1

nemo_retriever/src/nemo_retriever/service/retriever-service.yaml
```diff
@@ -102,6 +104,20 @@ pipeline:
   batch_workers: 16
   batch_queue_size: 4096
 
+# Dedicated LanceDB pod used by standalone/gateway modes for /v1/query.
+# Workers POST embedding rows to vectordb_url; the gateway proxies queries.
+vectordb:
```
jperez999, line 107: Should we be setting the model here? Isnt that something that should be dictated during the run?

nemo_retriever/src/nemo_retriever/service/cli.py
```diff
@@ -14,6 +15,14 @@
 app = typer.Typer(help="Operate the Retriever service. Use `retriever ingest service` to submit documents.")
 
 
+class VectorDbRetrievalModeCliOption(str, Enum):
+    """CLI choices for ``--vectordb-retrieval-mode``."""
+
+    dense = "dense"
+    hybrid = "hybrid"
+    auto = "auto"
```
jperez999: In what situation would auto not be acceptable? Like it seems I should not be allowed to choose hybrid when the ingestion was dense and vice versa, no? Shouldn't always be auto so that no matter what, as long as you have the table will know what type it should be.

nemo_retriever/src/nemo_retriever/service/vectordb_app.py
```diff
@@ -137,18 +143,33 @@ def __init__(
...
+        # explicitly configured: hybrid promises vector + FTS, so we make good
+        # on that promise even for a table this process did not create. `auto`
+        # is intentionally excluded here — it performs non-destructive detection
+        # of an existing table and must not mutate one it did not just write
+        # (a caller who wants FTS built should configure hybrid, or ingest under
+        # auto, which builds it on write; see `write_rows`/`_ensure_hybrid_indexes`).
+        # Kept outside the open-table probe above so a build failure fails fast
+        # at startup instead of being silently swallowed and leaving the service
+        # advertising hybrid while queries return HTTP 422.
+        if self._table_exists and self.retrieval_mode == "hybrid":
```
jperez999: Based on the comment above, are you allowing someone to change a table from dense to hybrid if they so choose?

nemo_retriever/src/nemo_retriever/service/vectordb_app.py
```diff
@@ -161,6 +182,134 @@ def embed_mode(self) -> str:
...
+        if mode == "hybrid" and (not caps.has_vector or not caps.has_fts):
+            raise ValueError(
+                f"LanceDB table {self.table_name!r} at {self.lancedb_uri!r} cannot run hybrid retrieval: "
+                "both a vector column and FTS index are required."
+            )
+        if mode == "sparse":
+            raise ValueError("Sparse retrieval is not supported by the VectorDB service. Use dense, hybrid, or auto.")
+        return mode
+
+    def _ensure_hybrid_indexes(self, *, refresh_after_append: bool = False) -> None:
```
jperez999: I like the part about incrementally updating the BM25 index, but I am not sure I should just be able to add that. Would that even work if you dont have the vocabulary?

jperez999 (dismissed):

nemo_retriever/src/nemo_retriever/service/retriever-service.yaml
mahikaw: yes, the YAML line is just the bundled default; production overrides it at deploy time.

nemo_retriever/src/nemo_retriever/service/vectordb_app.py
mahikaw: falling back to `auto` only with pre-req of pre-built index

nemo_retriever/src/nemo_retriever/service/cli.py
mahikaw: falling back to `auto` only with pre-req of pre-built index

jioffe502 (approved):

Issue #2320, opened 2026-07-09
Unify normalized text chunking for HTML and plain-text inputs


jioffe502:
## Context

HTML ingestion converts source content to an in-memory Markdown string with MarkItDown, then builds token chunks and ingestion rows in `_markdown_to_chunks_df()`. Native `.txt`, `.md`, `.json`, and `.sh` inputs perform nearly identical chunking and row construction in the text module.

Current behavior is correct, but the duplicated implementation could drift when tokenizer settings, chunk metadata, or empty-input handling change.

## Goal

Introduce one shared in-process text-to-chunks implementation at the seam after format normalization:

```text
HTML -> MarkItDown --+
                     +-> shared text-to-chunks module
Text/Markdown -------+
```

## Acceptance criteria

- HTML and native text adapters call one shared text-to-chunks implementation.
- Existing text, metadata, source paths, and chunk boundaries remain unchanged.
- Representative HTML, Markdown, TXT, JSON, and shell inputs retain equivalent behavior.
- Tests verify behavior through the existing ingestion interface.
- Redundant implementation-level tests are removed where equivalent public behavior coverage exists.

## Non-goals

- New Markdown, JSON, or shell parsers.
- A new chunking strategy.
- Retrieval-quality or performance changes.

This is low-priority maintainability work and should not block extension-support fixes.

PR #2321, opened 2026-07-09
Support documented Markdown, JSON, and shell text inputs

jioffe502:
## Summary

- route `.md`, `.json`, and `.sh` through the existing plain-text extraction path
- support the formats across root ingest, GraphIngestor, buffers, directory/glob expansion, and service uploads
- preserve case-insensitive directory discovery and the existing `text/plain` service behavior
- update the bundled Retriever skill format guidance

## Why

The public documentation already lists Markdown, JSON, and shell files as supported text inputs, but the shared local and service extension registries only accepted `.txt`. As a result, `retriever ingest` and GraphIngestor rejected documented formats before extraction.

This change only restores the documented routing behavior. It does not add parsers, new chunking behavior, or retrieval changes.

## Validation

- focused test suite: `241 passed, 2 deselected`
- configured pre-commit hooks: passed
- Retriever skill validation: passed
- live hybrid ingest with `nvidia/llama-nemotron-embed-1b-v2`: 3 files (`.md`, `.json`, `.sh`) -> 3 LanceDB rows
- two-axis Standards and Spec review: no findings

## Follow-up

- #2320 tracks the low-priority opportunity to unify normalized text-to-chunks row construction for HTML and native text inputs.

### DIFF
--- a/docs/docs/extraction/api-keys.md
+++ b/docs/docs/extraction/api-keys.md
@@ -21,7 +21,7 @@ export NVIDIA_API_KEY="nvapi-..."
 
 On Windows PowerShell you can use `$env:NVIDIA_API_KEY = "nvapi-..."`.
 
-For a full list of related variables, see [Environment configuration variables](environment-config.md).
+For a full list of related variables, refer to [Environment configuration variables](environment-config.md).
 
 !!! note
 
--- a/docs/docs/extraction/prerequisites-support-matrix.md
+++ b/docs/docs/extraction/prerequisites-support-matrix.md
@@ -63,6 +63,8 @@ Ensure your deployment environment meets these specifications before running the
 
 The NeMo Retriever Library extraction core pipeline features run on a single A10G or better GPU.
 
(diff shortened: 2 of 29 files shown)


mahikaw (approved):

jioffe502:
Greptile follow-up: `glob.glob` in `resolve_input_patterns` is indeed case-sensitive on Linux, but the reported affected surface is not accurate. That helper is used by the legacy `retriever pipeline run` directory path; `GraphIngestor.files()` accepts file paths or glob patterns and does not expand directory literals. Explicit SDK inputs are classified through `input_type_for_path()`, which lowercases the suffix, and I verified an actual `GraphIngestor` extraction of `README.MD` succeeds with the original Markdown text. The legacy pipeline limitation also predates this PR for `.TXT` and other uppercase extensions. The root `retriever ingest` directory surface changed here is case-insensitive and covered by a focused test. Fixing legacy pipeline directory discovery would broaden this compatibility change, so no code change is planned for that note.

PR #2334, opened 2026-07-13
Remove obsolete staged CLI and graphic-elements code

jioffe502:
## Summary

This is a focused dead-code cleanup: **124 files changed, 92 additions, and 10,685 deletions**.

It removes three obsolete implementation families that survived the move to the modern operator graph:

- the isolated numbered-stage system;
- format-specific CLI wrappers that duplicated or bypassed `retriever ingest`;
- the fully deprecated Nemotron graphic-elements path.

The supported product and SDK paths remain: `retriever ingest`, `retriever query`, `retriever harness`, `retriever service`, `GraphIngestor`, `ServiceIngestor`, the service pipeline executor, and the compatibility `retriever pipeline` command.

## Recommended review order

Please review this PR commit by commit. Each commit is one independently motivated deletion slice.

| Commit | Review unit | Size | Why it is safe to delete |
| --- | --- | ---: | --- |
| `1cf10324` | Remove obsolete numbered-stage system | 23 files, -2,795 lines | `cli/local/stages`, `graph/stages`, stage artifact helpers, and the stage VDB uploader had no modern production, SDK, or service caller. The active graph is `GraphIngestor` + `graph/ingestor_runtime.py` + `operators/`. |
| `c401bd03` | Remove obsolete format-specific CLI wrappers | 42 files, -3,359 lines | The wrappers were root-hidden or unregistered and were not called by modern ingest. Their supported capabilities remain in the modern operators and `GraphIngestor`. Orphans were deleted only after their wrapper/test consumers disappeared. |
| `95783bc0` | Remove deprecated graphic-elements extraction | 66 files, -4,519 lines | Graphic-elements was already deprecated and removed from the public CLI. This removes its remaining SDK, service, config, model, resource-planning, dependency, and operator branches. Normal chart extraction remains page-elements detection plus OCR. |
| `083303ec` | Remove obsolete PDF fallback from the Retriever skill | 1 file | The only remaining repository reference to `retriever pdf` was an agent recovery fallback. It bypassed indexing and encouraged per-PDF corpus processing. Recovery now stays on `retriever ingest` / `retriever query` and reports an ingestion failure instead of silently changing workflows. |

For the fastest review, the first two commits are primarily reachability-backed deletions. The third commit is the behavioral hotspot: verify that graphic-elements-specific branches are gone while page elements, OCR, table structure, and chart extraction remain. The fourth commit closes the only active documentation reference to a deleted wrapper.

## What remains intentionally

[...]

PR #2364, opened 2026-07-15
Remove orphaned adapters and retired PDF DataFrame API

jioffe502:
## Summary

This is the next focused dead-code cleanup after #2334: **42 files changed, 19 additions, and 7,603 deletions**.

It removes:

- orphaned internal image/table/infographic extraction, embedding, schema, logging, and OCR adapters;
- an unregistered `retriever compare` shell and unused chunk-bundle helper;
- the retired nv-ingest direct-DataFrame PDF interface and alternate-engine adapters;
- one private-fixture test and the root `pytest.ini` inherited from the old repository layout.

Supported CLI, SDK, service, graph, media-ingestion, and modern extraction paths remain unchanged.

## Recommended review order

Please review this PR commit by commit. Each commit is one independently motivated deletion slice.

| Commit | Review unit | Size | Why it is safe to delete |
| --- | --- | ---: | --- |
| `7a47a3d8` | Remove orphaned internal adapter stacks | 17 files, +14/-4,266 | The removed modules had no production, SDK, service, CLI, or test caller. The small surviving call sites use their modern equivalents. |
| `150c3ac4` | Remove unreachable compare and chunk helpers | 5 files, +1/-112 | The compare command had no registered subcommands or usable root entry point, and `chunk_bundle.py` had no caller. |
| `089e16a4` | Remove retired nv-ingest PDF DataFrame interface | 17 files, -3,184 | #2334 removed its last CLI consumer. The direct-DataFrame API and alternate-engine wrappers are not supported interfaces. |
| `d1d1a95b` | Remove obsolete and unwired tests | 4 files, -186 | Removes old-root configuration, one hard-coded private-fixture test, and three integration tests with no automated execution path. |
| `52d11b7c` | Restore active integration coverage | 2 files, +149 | Restores the Parakeet compatibility test unchanged and updates two remote-error tests to exercise the current chart/page-elements route. |

## Compatibility boundary

This removes unsupported internal nv-ingest adapters, not the supported Retriever interfaces. `retriever ingest`, `retriever query`, `GraphIngestor`, `ServiceIngestor`, and the modern operator graph remain.

The PDF slice removes the retired direct-DataFrame API, not PDF support. `pypdfium2`, shared page rasterization, `PDFSplitActor`, `PDFExtractionActor`, and the modern OCR/page-elements/table-structure paths remain unchanged.

The final test cleanup removes only the private-path JP20 test. Active Helm, ffmpeg/audio/video, Parakeet, remote-error, and scheduled multimodal PDF coverage remains.

## Validation

- Full package suite: **2,507 passed, 110 skipped, 4 deselected; 26 subtests passed**.
- Restored remote-error integrations: **2 passed**; the restored Parakeet test remains opt-in and skipped without its model/fixture environment.
- A checked-in PDF rasterized successfully through both retained PDFium render paths.
- All applicable pre-commit hooks passed.

## Reviewer decision requested

Please confirm that the removed internal nv-ingest paths are not compatibility surfaces we intend to support; the active modern operator and PDF rasterization paths are intentionally preserved.


jperez999 (changes requested):
Did you need to remove those tests? If so, why? From the names of the tests it seems like they were good to have, to check that functionality works as expected.

jioffe502:
@jperez999 You're right to question these separately:

- `test_pdfium_output_can_have_empty_text_without_ocr_flag`: removed because it depends on the private absolute path `/raid/data/jp20/1312679.pdf` and skips everywhere that fixture is unavailable.
- The two `GraphIngestor` remote-stage tests: I removed them because no automated workflow runs the `integration` marker. Looking at their history, they were added in #1999 to guard active remote error handling, so that is not a strong enough reason to remove them.
- The Parakeet compatibility test: it is opt-in and needs an external audio fixture/model, but it was added in #1929 to test the active local ASR path. I do not think we should remove it without an audio-specific reason.

So the JP20 test removal is justified. I think the other three tests should be restored.

jioffe502:
Restored the three active-path tests in `52d11b7c`:

- The Parakeet compatibility test is restored unchanged.
- The two remote-error tests are restored and now use `extract_charts=True`, which exercises the current `PageElementDetectionActor` route. The old image-only trigger no longer scheduled that stage, so the restored tests initially exposed stale setup.

Both remote-error integrations now pass when run explicitly. The private-path JP20 test remains removed because it cannot run without `/raid/data/jp20/1312679.pdf`.

jperez999 (approved):

PR #2368, opened 2026-07-16
fix(service-mode): add support for TXT and HTML chunking in service mode

KyleZheng1284:
## Summary

Fixes TXT and HTML ingestion in service mode without adding `transformers`, PyTorch, or model weights.

The service now loads the revision-pinned `tokenizer.json` associated with the default embedding model using lightweight `tokenizers` and `huggingface-hub` dependencies.

## What changed

- Added `DOWNLOAD_DEFAULT_TOKENIZER=True` to optionally pre-cache the default tokenizer during image builds.
- Applied pre-caching to both `service` and `service-gpu` images.
- Kept `service` offline at runtime while allowing `service-gpu` to download its other Hugging Face models.
- Preserved per-document error isolation while surfacing missing-tokenizer failures clearly.
- Updated supported build workflows, Compose examples, documentation, and tests.

## Validation

- 23 focused regression tests passed after final cleanup.
- Validated `service` and `service-gpu` builds with pre-caching enabled and disabled.
- Offline non-root TXT and HTML chunking passed without `transformers`.
- End-to-end TXT and HTML HTTP ingestion returned chunks successfully.
- Compose, Helm, formatting, lint, lockfile, and diff checks passed.

## Checklist

- [x] The documentation is up to date.


Dockerfile
```diff
@@ -116,7 +117,8 @@ SHELL ["/bin/bash", "-c"]
 RUN --mount=type=cache,target=/root/.cache/pip \
     --mount=type=cache,target=/root/.cache/uv \
     . /opt/retriever_runtime/bin/activate \
-    && uv pip install -e "./nemo_retriever[service]"
+    && uv pip install -e "./nemo_retriever[service]" \
+    && python -c "from nemo_retriever.common.modality.txt.tokenizer_provider import load_chunk_tokenizer; load_chunk_tokenizer('nvidia/llama-nemotron-embed-vl-1b-v2')"
```
charlesbluca: Could we restore the `DOWNLOAD_LLAMA_TOKENIZER` build-argument gate that the existing Docker build and release tooling still passes, and use it for a shared pre-cache step in both `install` and `install-service-gpu`? The step should resolve `DEFAULT_TOKENIZER_MODEL_ID` instead of hardcoding the model ID, keeping the cached artifact aligned with the runtime chunker and its pinned revision. Currently the historical Dockerfile gate has been removed, so the tooling knob is ignored, `service` downloads unconditionally, and `service-gpu` does not pre-cache the tokenizer.

Dockerfile
```diff
@@ -158,6 +160,7 @@ CMD ["/bin/bash"]
 FROM install AS service
 
 ENV NEMO_RETRIEVER_SERVICE_CONFIG=/etc/nemo-retriever/retriever-service.yaml
+ENV HF_HUB_OFFLINE=1
```
charlesbluca, line 172: Please keep `HF_HUB_OFFLINE=1` scoped to the `service` image when adding the tokenizer pre-cache to `service-gpu`. The GPU image is expected to download local Hugging Face model weights at runtime, so it should reuse the pre-cached tokenizer without globally disabling Hub access.

charlesbluca (commented):
Two Docker image consistency points from the Helm deployment review:

charlesbluca (approved):

Issue #371, opened 2025-01-23
[BUG]: Base python client example fails (paddle model error)


joaopcm1996:
### Version

24.10.1

### Which installation method(s) does this occur on?

Docker

### Describe the bug.

Deployed nvingest with the docker compose from main branch, on H100's.

Tried to run the base python client example, with the default document in the [notebook](https://github.com/NVIDIA/nv-ingest/blob/main/client/client_examples/examples/python_client_usage.ipynb).

Immediately got an error in the handling of paddle NIM output.
Error logs extracted from nvingest runtime container.

### Minimum reproducible example

```shell
https://github.com/NVIDIA/nv-ingest/blob/main/client/client_examples/examples/python_client_usage.ipynb
```

### Relevant log output

```shell
INFO:     172.18.0.7:38424 - "GET /v1/health/ready HTTP/1.1" 200 OK
2025-01-22 11:04:55,866 - ERROR - Inference failed for model paddle: 'int' object has no attribute 'decode'
Unhandled error calling image inference model: Inference failed for model paddle: 'int' object has no attribute 'decode'
Traceback (most recent call last):
  File "/opt/conda/envs/nv_ingest/lib/python3.10/site-packages/nv_ingest/util/nim/helpers.py", line 120, in _call_image_inference_grpc_client
    return " ".join([output[0].decode("utf-8") for output in result.as_numpy("output")])
  File "/opt/conda/envs/nv_ingest/lib/python3.10/site-packages/nv_ingest/util/nim/helpers.py", line 120, in <listcomp>
    return " ".join([output[0].decode("utf-8") for output in result.as_numpy("output")])
AttributeError: 'int' object has no attribute 'decode'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/conda/envs/nv_ingest/lib/python3.10/site-packages/nv_ingest/stages/nim/table_extraction.py", line 78, in _update_metadata
    paddle_result = call_image_inference_model(paddle_client, "paddle", image_array, trace_info=trace_info)
  File "/opt/conda/envs/nv_ingest/lib/python3.10/site-packages/nv_ingest/util/tracing/tagging.py", line 185, in wrapper_inject_trace_info
    result = func(*args, **kwargs)
  File "/opt/conda/envs/nv_ingest/lib/python3.10/site-packages/nv_ingest/util/nim/helpers.py", line 103, in call_image_inference_model
[...]

PR #2378, opened 2026-07-17
feat: add scoped collection and document management APIs

### LINKED ISSUE #2353: [FEA]: Enable VDB collection management via Python API

New Feature

Significant improvement

To be filled out as needed

To be filled out as needed

### Additional context

_No response_

KyleZheng1284:
## Description

This PR adds collection and document lifecycle management to the NeMo Retriever service. It provides a logical, scope-isolated API for creating collections, ingesting and replacing documents, tracking asynchronous jobs, querying collection content, and cleaning up resources without exposing physical LanceDB table names or storage locations to clients.

### What changed

- Added scope-isolated APIs for managing collections, documents, ingestion jobs, and retrieval.
- Added stable document identities, idempotent ingestion, asynchronous status tracking, and retryable lifecycle cleanup.
- Kept authentication, physical VectorDB identifiers, storage locations, and owned artifacts within the service boundary.
- Added Python client support with API compatibility handling and legacy SSE fallback.
- Added deployment support through Docker Compose and Helm, including non-root artifact storage.
- Added regression tests, API documentation, configuration examples, and operational guidance.

### Why

Applications that want to integrate NeMo Retriever (like aiq) need a stable service boundary for managing durable knowledge collections. Previously, consumers had to depend on ingestion-specific behavior or physical VectorDB details.

This change establishes a logical API boundary where:

- clients work with collection names and stable document IDs;
- authentication determines the authorized scope;
- internal table names and storage locations remain service-owned;
- ingestion attempts can be retried without changing the logical document identity;
- collection cleanup and expiration remain recoverable after service restarts.

### Validation

Focused regression testing was run from commit `edfed55da`:

```text
35 passed in 26.06s
```

The focused coverage includes:

- collection CRUD and pagination;
- scope authorization and cross-scope isolation;
- internal VectorDB authentication;
- idempotent job replay;
- stable document IDs and attempt-ID tracking;
- collection and document cleanup behavior;
- API-version compatibility errors;
- legacy SSE fallback behavior;
- unique OpenAPI operation IDs;
- non-root artifact-directory ownership; and
- collection-management CI and Compose contracts.

The collection workflow was also exercised end to end with collection creation, asynchronous ingestion, job polling, document listing, retrieval, deletion, cleanup, and persistence across service recreation.

### Out of scope

The separate TXT/HTML tokenizer fix is intentionally not included in this PR. The integrated validation environment contained that independent patch, but this PR does not modify the tokenizer implementation.

## Checklist


KyleZheng1284:
@greptileai

nemo_retriever/dev/compose/secrets/artifact-storage-options.json.example
```diff
@@ -0,0 +1 @@
+{}
```
jperez999: is this file absolutely necessary? An empty json...

nemo_retriever/dev/compose/secrets/internal-vdb-token.example
```diff
@@ -0,0 +1 @@
+development-only-internal-token
```
jperez999: Is this necessary?

nemo_retriever/dev/compose/secrets/scope-tokens.json.example
```diff
@@ -0,0 +1 @@
+{"tokens":[{"token":"development-only-change-me","scopes":["default"]}]}
```
jperez999: same as above, what is the purpose?

nemo_retriever/dev/compose/collection-management.compose.yaml
```diff
@@ -0,0 +1,74 @@
...
+    secrets:
+      - internal_vdb_token
+      - artifact_storage_options
+    healthcheck:
+      test: ["CMD", "python", "-c", "import urllib.request; urllib.request.urlopen('http://localhost:7671/v1/health')"]
+      interval: 10s
+      timeout: 5s
+      retries: 12
+
+secrets:
```
jperez999: since we dont actually support docker compose, can we just inline these values as defaults, either here or in the service and provide a warning if they are not overridden?

nemo_retriever/dev/compose/collection-management.service.yaml
```diff
@@ -0,0 +1,19 @@
+mode: standalone
```
jperez999: Isn't this for helm?

nemo_retriever/helm/templates/configmap.yaml
```diff
@@ -139,8 +139,11 @@ resources:
   max_upload_bytes: {{ $maxUploadBytes }}
 
 auth:
-  api_token: {{ .Values.serviceConfig.auth.apiToken | default "null" }}
+  api_token: null
```
jperez999: is this actually a good idea, dont we need this piping?

nemo_retriever/helm/templates/configmap.yaml
```diff
@@ -151,6 +154,11 @@ vectordb:
   embed_model: {{ .Values.serviceConfig.vectordb.embedModel | quote }}
   embed_model_provider_prefix: {{ if .Values.serviceConfig.vectordb.embedModelProviderPrefix }}{{ .Values.serviceConfig.vectordb.embedModelProviderPrefix | quote }}{{ else }}null{{ end }}
   vectordb_url: "http://{{ .vectordbSvc }}:{{ .vectordbPort }}"
+  internal_api_token: null
```
jperez999: shouldnt this have a pipe from the values file? How do you envision this being set?

nemo_retriever/helm/templates/deployment-vectordb.yaml
```diff
@@ -163,6 +198,11 @@ spec:
       volumes:
         - name: tmp
           emptyDir: {}
+        {{- if $storageSecretEnabled }}
+        - name: artifact-storage-options
```
jperez999: same name twice? Please elaborate here.

nemo_retriever/helm/values.yaml
```diff
@@ -661,12 +661,39 @@ serviceConfig:
     tableName: "nemo_retriever"
     embedModel: "nvidia/llama-nemotron-embed-vl-1b-v2"
     embedModelProviderPrefix: ""
+    # Dedicated service-to-VectorDB credential. Prefer an existing Secret.
+    internalTokenSecret:
```
jperez999: The utility of this within our design... I dont think it has a place here. Please explain use case where this is a good idea.

nemo_retriever/src/nemo_retriever/common/schemas/collections.py
```diff
@@ -0,0 +1,145 @@
...
+
+    @field_validator("expires_at", mode="before")
+    @classmethod
+    def _validate_expiry(cls, value: str | None) -> str | None:
+        return _normalize_expires_at(value)
+
+
+class CollectionInfo(RichModel):
+    name: str
+    scope: str
```
jperez999, line 84: what does scope mean here?

nemo_retriever/src/nemo_retriever/common/schemas/collections.py
```diff
@@ -0,0 +1,145 @@
...
+    str,
+    StringConstraints(
+        min_length=1,
+        max_length=128,
+        pattern=r"^[A-Za-z0-9][A-Za-z0-9._-]*$",
+    ),
+]
+
+
+def _normalize_expires_at(value: str | None) -> str | None:
```
jperez999, line 47: This is just about timing. this helper function should in tools with the other helpers.

nemo_retriever/src/nemo_retriever/common/schemas/collections.py
```diff
@@ -0,0 +1,145 @@
...
+    try:
+        parsed = datetime.fromisoformat(value.replace("Z", "+00:00"))
+    except ValueError as exc:
+        raise ValueError("expires_at must be RFC3339") from exc
+    if parsed.tzinfo is None or parsed.utcoffset() is None:
+        raise ValueError("expires_at must include a timezone offset")
+    return parsed.astimezone(timezone.utc).isoformat()
+
+
+class CollectionCreateRequest(RichModel):
```
jperez999, line 53: Are all of these models, strictly for responses from API calls?

nemo_retriever/src/nemo_retriever/common/schemas/collections.py
```diff
@@ -0,0 +1,145 @@
...
+    content_type: str | None = None
+    source: Any = None
+    source_id: str | None = None
+    stored_image_uri: str | None = None
+    bbox: Any = None
+    metadata: dict[str, Any] = Field(default_factory=dict)
+
+    @field_validator("page_number", mode="before")
+    @classmethod
+    def _normalize_page(cls, value: Any) -> int | None:
```
jperez999: This is really normalize value. Did you look to see if this was available in the tools sub module already? If not, it should still be moved there.

nemo_retriever/src/nemo_retriever/common/schemas/requests.py
```diff
@@ -55,3 +64,36 @@ class JobCreateRequest(RichModel):
...
     )
+    collection_name: str | None = Field(default=None, min_length=1, max_length=128)
+    operation: IngestOperation = "append"
+    target_document_id: DocumentId | None = None
+    idempotency_key: str | None = Field(default=None, min_length=1, max_length=256)
+    document_manifest: list[DocumentManifestEntry] = Field(default_factory=list)
+
+    @model_validator(mode="before")
+    @classmethod
+    def _reject_raw_storage_keys(cls, value: Any) -> Any:
```
jperez999: In what case is this used?

nemo_retriever/src/nemo_retriever/common/vdb/records.py
```diff
@@ -37,6 +37,12 @@ class RetrievalHit(TypedDict, total=False):
     bbox_xyxy_norm: list[float]
     _distance: float
     _score: float
+    score: float
```
jperez999: what is the difference between score and _score?

nemo_retriever/src/nemo_retriever/common/vdb/records.py
```diff
@@ -256,11 +262,29 @@ def _normalize_hit(hit: dict[str, Any]) -> RetrievalHit:
         "pdf_basename": pdf_basename,
         "pdf_page": f"{pdf_basename}_{page_number}" if pdf_basename and page_number is not None else "",
     }
+    chunk_id = entity.get("chunk_id") or hit.get("chunk_id")
```
jperez999: This is about normalizing hits... shouldnt you just make sure the hit returns the parts of the schema you want? Why do you have dive into the entity? shouldn't the field you want be in a specific place and not multiple places?

jperez999 (changes requested):
Lets do this in multiple passes

nemo_retriever/helm/templates/configmap.yaml
KyleZheng1284: will follow up w/ helm changes as separate PR

nemo_retriever/helm/templates/configmap.yaml
KyleZheng1284: will follow up w/ helm changes as separate PR

nemo_retriever/helm/templates/deployment-vectordb.yaml
KyleZheng1284: will follow up w/ helm changes as separate PR

nemo_retriever/helm/values.yaml
KyleZheng1284: will follow up w/ helm changes as separate PR

nemo_retriever/dev/compose/collection-management.service.yaml
KyleZheng1284: I'm not sure I'm following on this exactly question exactly. For this I was looking at it as its own separate compose bring up as opposed to the pre-existing service compose with the added on functionality from this PR (auth and collection lifecycle settings)

Would you prefer if we moved this into the pre-existing compose then, I was suspecting it might be better to just keep that one for backwards compatibility?

nemo_retriever/dev/compose/collection-management.compose.yaml
KyleZheng1284: went ahead and removed the smaller example secret files, now requiring tokens at runtime

nemo_retriever/src/nemo_retriever/common/schemas/collections.py
KyleZheng1284: Its not only for responses but rather for both the request + response that the service and sdk can use for the api contract for collection management

nemo_retriever/src/nemo_retriever/common/schemas/collections.py
KyleZheng1284: scope represents the tenant or workspace that the caller is allowed to access. It uses the caller’s token and scope to keep collections, documents, ingestion jobs, and queries isolated. 

within the scope, each (scope, collection_name) maps to a private LanceDB table

for aiq the session UUID becomes the collection name, so the scope will ensure that other tenants or deployments cannot access that session’s data which is something that would be valuable for when we deploy to astra

nemo_retriever/src/nemo_retriever/common/schemas/collections.py
KyleZheng1284: moved it to `common/api/util/converters/datetools.py`

nemo_retriever/src/nemo_retriever/common/schemas/collections.py
KyleZheng1284: it didn't seem to be in the tools sub module already, went ahead and moved it under `common/api/util/converters/pagetools.py`. we are only calling on it when converting retrieval results into citation pages.

nemo_retriever/src/nemo_retriever/common/vdb/records.py
KyleZheng1284: _score is a backend-native FTS/BM25 diagnostic, while the previous public score was normalized by nrl

for this portion I actually removed the normalization on the retriever side and it was only implemented because aiq required a `chunk.score` between [0,1]

That conversion probably belongs more on the AIQ adapter rather than nrl so I have altered it back to the native behavior, since other consumers may need the native ranking values and should not be forced into the aiq specific contract.

nemo_retriever/src/nemo_retriever/common/schemas/requests.py
KyleZheng1284: This prevents clients from passing internal storage details like table_name or lancedb_uri through the public API

Without this check, pydantic would silently ignore those fields which makes the request look successful even though nrl never used them

nemo_retriever/src/nemo_retriever/common/api/util/converters/datetools.py
```diff
@@ -11,6 +11,40 @@
 from nemo_retriever.common.api.util.exception_handlers.converters import datetools_exception_handler
 
 
+def normalize_timezone_aware_iso8601_to_utc(date_string: str) -> str:
```
jperez999, line 14: Did you hit this problem? Or do you expect to? Please describe the real world actual scenario where this would be important.

nemo_retriever/src/nemo_retriever/common/api/util/converters/pagetools.py
```diff
@@ -0,0 +1,43 @@
...
+
+"""Converters for human-facing document page numbers."""
+
+from __future__ import annotations
+
+from decimal import Decimal, InvalidOperation
+from typing import Any
+
+
+def normalize_one_based_page_number(value: Any) -> int | None:
```
jperez999: What is the problem with using zero based and just adding one? Please describe an example where you hit an issue because of this?

nemo_retriever/src/nemo_retriever/common/vdb/records.py
```diff
@@ -7,10 +7,48 @@
...
 from pathlib import Path
 from typing import Any, TypedDict
 
+from nemo_retriever.common.api.util.converters.pagetools import (
+    normalize_one_based_page_number,
+)
+from nemo_retriever.common.vdb.lancedb_capabilities import LanceRetrievalMode
+
+
+class RetrievalContractError(RuntimeError):
```
jperez999, line 33: Under what circumstances is this not a value error? What is this actually supposed to bubble up? Why do we need a new error type for this, what do we gain from this?

nemo_retriever/src/nemo_retriever/common/vdb/records.py
```diff
@@ -37,6 +80,13 @@ class RetrievalHit(TypedDict, total=False):
     bbox_xyxy_norm: list[float]
     _distance: float
     _score: float
+    _relevance_score: float
```
jperez999: What is the difference between score and relevance score here?

nemo_retriever/src/nemo_retriever/common/vdb/records.py
```diff
@@ -7,10 +7,48 @@
...
+        "source_id",
+        "source_metadata",
+        "text",
+    }
+)
+
+_NATIVE_SCORE_FIELDS = frozenset({"_distance", "_score", "_relevance_score"})
+
+
+class RetrievalRanking(TypedDict):
```
jperez999: Why is this required, rankings come in sorted (score) order, and if not they can be sorted by score?

nemo_retriever/src/nemo_retriever/common/vdb/records.py
```diff
@@ -216,31 +276,46 @@ def _mapping(value: Any) -> dict[str, Any]:
     return parsed if isinstance(parsed, dict) else {}
 
 
+def _flatten_legacy_entity_hit(hit: dict[str, Any]) -> dict[str, Any]:
```
jperez999, line 364: Your taking away nesting here? Like no more json dicts?

jperez999: can you provide and example of a legacy and a "new" hit? Also why they are needed?

nemo_retriever/src/nemo_retriever/common/vdb/records.py
```diff
@@ -254,16 +329,71 @@ def _normalize_hit(hit: dict[str, Any]) -> RetrievalHit:
...
+    if retrieval_mode == "dense":
+        field = "_distance"
+        kind = "vector_distance"
+        higher_is_better = False
+    elif retrieval_mode == "hybrid":
+        field = "_relevance_score"
+        kind = "hybrid_relevance"
+        higher_is_better = True
+    else:
+        raise RetrievalContractError(f"unsupported collection retrieval mode: {retrieval_mode}")
```
jperez999: All of these are just value errors?

nemo_retriever/src/nemo_retriever/service/routers/ingest.py
```diff
@@ -632,6 +725,68 @@ async def create_job(request: Request, response: Response, body: JobCreateReques
     tracker = get_job_tracker()
     if tracker is None:
         raise HTTPException(status_code=503, detail="Job tracker not available")
+    from nemo_retriever.service.auth import authorized_scope, internal_auth_headers
+
+    scope = authorized_scope(request)
+    if body.collection_name:
```
jperez999, line 845: This is now a first class feature, the collection management api, so it should be apart of all deployments? not just an option right?

jperez999 (changes requested):

nemo_retriever/src/nemo_retriever/service/app.py
```diff
@@ -251,26 +274,29 @@ def create_app(config: ServiceConfig) -> FastAPI:
 
     app.add_middleware(_RequestIdMiddleware)
 
-    if config.auth.api_token:
-        from nemo_retriever.service.auth import BearerAuthMiddleware
+    from nemo_retriever.service.auth import BearerAuthMiddleware
```
jperez999, line 278: This only happened if there was a config auth api token, now it happens all the time, why the change?

nemo_retriever/src/nemo_retriever/common/api/util/converters/datetools.py
KyleZheng1284: aiq currently relies on its ingestion backend (frag, llamaindex etc.) to keep track of its time as metadata because we do a TTL of 24 hrs for all collections. this would be useful in the case perhaps where aiq or a different application have users in multiple time zones like utc to pst to est and without this there is a possiblity that it causes the cleanup comparsion to fail

AIQ sends: 2030-01-01T01:00:00+01:00
NRL stores: 2030-01-01T00:00:00Z

nemo_retriever/src/nemo_retriever/common/api/util/converters/pagetools.py
KyleZheng1284: page_number doesn't always mean document page right, this one in particular doesn't seem to be necessary for aiq but was surfaced during codex broad review.

It seems that current it can exist as this

 PDF page 1, 2, 3.
Audio chunk index 0, 1, 2.
Video frame index 0, 1, 2.
0 or -1 as an unknown/error marker.

Perhaps blindly adding one would break valid PDF pages and turn media indexes into fake citations

nemo_retriever/src/nemo_retriever/common/vdb/records.py
KyleZheng1284: would like to keep this would be useful for other applications if we can distinguish between scenarios A: client makes a bad request vs B: the backend returned an invalid result

KyleZheng1284: for instance if invalid top_k from the client → HTTP 422, calling on dense retrieval result missing _distance → HTTP 500.

nemo_retriever/src/nemo_retriever/common/vdb/records.py
KyleZheng1284: _score is lancedbs native FTS/BM25 score, while _relevance_score is for native hybrid ranking value

Dense search only uses _distance as its field. Should keep this for compatibility, and the public collection response exposes only the value that actually ordered the result.

I guess long term would there be perhaps a better way of normalizing this into one particular score or would that not be worth the lift?

nemo_retriever/src/nemo_retriever/common/vdb/records.py
KyleZheng1284: there is still nesting, the flattening is only occuring so that the collection hit has a better way of handling both the legacy and new hit shape.

currently it doesn't seem like consumers can support both the legacy and new hit shapes without writing some custom impl of it so flattening it should resolve that portion.

KyleZheng1284: from what I could gather this seems to be the old backend hit:
```
{
  "entity": {
    "text": "retrieved chunk",
    "source": {
      "source_id": "document.pdf"
    },
    "content_metadata": {
      "page_number": 2
    }
  },
  "_distance": 0.12
}
```

this is the new internal nrl hit
```
{
  "text": "retrieved chunk",
  "source": {
    "source_id": "document.pdf"
  },
  "metadata": {
    "page_number": 2
  },
  "_distance": 0.12
}
```

and for the collection hit this is how it exists right now:
```
{
  "text": "retrieved chunk",
  "source": "document.pdf",
  "metadata": {
    "page_number": 2
  },
  "chunk_id": "chunk-123",
  "document_id": "document-456",
  "filename": "document.pdf",
  "ranking": {
    "value": 0.12,
    "kind": "vector_distance",
    "higher_is_better": false
  }
}
```

nemo_retriever/src/nemo_retriever/common/vdb/records.py
KyleZheng1284: I mean yes essentially it is, but mapping it to be the new `RetrievalContractError` seems to be perhaps a better UI/UX as we can clearly show if its client vs nrl backend error better. Not sure if breakages on nrl backend is possible but still having these guards set up in case it does I think would be beneficial.

nemo_retriever/src/nemo_retriever/service/config.py
```diff
@@ -214,13 +215,18 @@ class ResourceLimitsConfig(RichModel):
...
 class AuthConfig(RichModel):
-    """Optional bearer-token authentication."""
+    """Bearer authentication and authorization for logical workspace scopes."""
 
     model_config = ConfigDict(extra="forbid")
 
     api_token: str | None = None
+    default_scope: str = "default"
+    scope_token_file: str | None = None
+    allow_unscoped_dev: bool = True
```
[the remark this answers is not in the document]
charlesbluca: Confirmed—this remains valid on `69581680` and in the current local successor state. Authentication probes return 401 only when credentials are explicitly configured; omitting them still enables the unscoped development path by default. Please make this fail closed and require an explicit opt-in for auth-disabled development.

nemo_retriever/src/nemo_retriever/common/vdb/records.py
charlesbluca: `_score` is a backend-native FTS/BM25 value, while `_relevance_score` is a hybrid reranker value. Neither is a normalized public confidence score. The collection API replaces those backend-specific fields with `ranking` metadata; the runtime payload confirms that this is the actual wire contract.

nemo_retriever/dev/compose/secrets/internal-vdb-token.example
charlesbluca: The internal gateway-to-VectorDB credential is necessary once VectorDB authentication is enabled, although this example file does not have to be the final packaging mechanism. The published PR no longer provides equivalent Helm secret mounts/configuration, while runtime validation confirms that the VectorDB endpoint rejects unauthenticated requests with 401. The deployment needs a supported way to inject the same credential into both sides.

nemo_retriever/src/nemo_retriever/common/vdb/records.py
charlesbluca: Confirmed—`ranking` is required by the paired contract. Runtime responses contain native ranking metadata and no legacy public `score`. Ordering alone loses the native value type and whether higher or lower is better; AIQ needs `value`, `kind`, and `higher_is_better` to derive its required query-relative score. The 1-based rank also lets consumers validate that the response order is consistent.

docs/docs/reference/collection-management-api.md
```diff
@@ -0,0 +1,176 @@
...
+`RetrieverServiceConflictError`, `RetrieverServiceValidationError`, or the base
+`RetrieverServiceError`. Resources are isolated by `scope`; cross-scope reads
+return 404. `expires_at` can be set at collection creation or update time for an
+operator cleanup process. Deletion is retryable and `if_exists=True` makes
+repeated deletion safe. Delete results report `existed`, `deleted`, `status`,
+and `cleanup_pending`; synchronous completion returns HTTP 200 and a retryable
+pending cleanup may return HTTP 202.
+
+Production deployments map bearer tokens to allowed workspace scopes. A valid
+token requesting an unauthorized scope receives 404 so resource existence is
```
charlesbluca: The implemented contract no longer returns 404 for a valid token with an unauthorized scope. `authorize()` intentionally returns 401 for both invalid credentials and wrong-scope credentials, and runtime validation confirms 401. Please update this paragraph—and any matching assertions—to document 401.

charlesbluca (changes requested):
Requesting changes for the remaining deployment blockers: authentication must fail closed by default, the published service image needs the TXT/HTML tokenizer dependency, and the published Helm path needs supported internal VectorDB credential wiring. The wrong-scope response documentation should also be aligned with the implemented 401 contract.

The focused NRL suite passed 140 tests. The relevant full run passed 2,791 tests with 4 skips, 10 deselections, and one environment-dependent shared-temp-path failure.

nemo_retriever/src/nemo_retriever/common/api/util/converters/datetools.py
KyleZheng1284: would like to personally keep this if possible since its a very minimal change. I feel like creating a consistent persisted format would help us avoid changing the storage/API behavior later.

nemo_retriever/dev/compose/service-mode.compose.yaml
```diff
@@ -291,7 +294,6 @@ configs:
         extract:
           enabled: ${LOCAL_EXTRACT_ENABLED:-true}
           use_table_structure: ${LOCAL_EXTRACT_USE_TABLE_STRUCTURE:-true}
-          use_graphic_elements: ${LOCAL_EXTRACT_USE_GRAPHIC_ELEMENTS:-true}
           ocr_version: "${LOCAL_EXTRACT_OCR_VERSION:-v2}"
```
KyleZheng1284, line 301: @charlesbluca is this safe to remove from my understanding it seems like use_graphic_elements is a stale configuration key. 

It seems like `use_graphic_elements` is no longer part of that schema, 
and is preventing the service from starting with an “extra field not permitted” validation error.

nemo_retriever/src/nemo_retriever/common/schemas/requests.py
KyleZheng1284: Have removed this in latest commits

nemo_retriever/src/nemo_retriever/common/schemas/collections.py
charlesbluca: Re-validated on pinned head `fe33d24b`: `scope` is the authenticated tenant/workspace boundary for collections, documents, jobs, and queries, and `(scope, collection_name)` resolves to service-owned storage. The API documentation describes this contract, and the focused cross-scope CRUD/isolation test passes.

nemo_retriever/src/nemo_retriever/common/schemas/collections.py
charlesbluca: Confirmed on pinned head `fe33d24b`: these are shared public wire models, not response-only models. Create/update models validate requests; collection, document, and query models validate responses used by both the service and SDK. The module documentation now states that contract explicitly.

nemo_retriever/src/nemo_retriever/common/api/util/converters/datetools.py
charlesbluca: Re-validated on pinned head `fe33d24b`: `2030-01-01T01:00:00+01:00` normalizes to `2030-01-01T00:00:00+00:00`, preventing equivalent instants from comparing differently during TTL reconciliation. Both the timezone-normalization test and the expired-collection reconciliation test pass on the exact PR head.

nemo_retriever/src/nemo_retriever/common/schemas/collections.py
```diff
@@ -0,0 +1,157 @@
...
+    cleanup_pending: bool = False
+
+
+class QueryHit(RichModel):
+    """Citation-ready hit returned to agentic applications."""
+
+    chunk_id: str
+    document_id: DocumentId
+    text: str
+    distance: float = Field(
```
charlesbluca, line 153: **P1: The current `QueryHit.distance` contract is incompatible with AIQ PR #371**

On pinned NRL head `fe33d24b`, `QueryHit` requires native `distance`. The live head of AIQ PR #371 is still `1c73ba49`, where `QueryHitWire` instead requires `score`. Fresh validation of a representative exact-head NRL payload containing `distance: 0.125` fails with `score: missing`, so collection-query integration fails at wire validation before the AIQ adapter can normalize the result.

For clarity, an unpublished local AIQ successor is not the state of PR #371: it targets the earlier NRL `ranking` contract and rejects the current `distance` payload with `ranking: missing`. It therefore does not resolve the published cross-PR incompatibility.

Please land one coordinated wire contract across NRL and AIQ. If NRL keeps `distance`, update AIQ PR #371 and its compatibility tests to consume native distance and derive any required bounded `Chunk.score` at the adapter boundary. If NRL changes the wire contract again, coordinate that change with the AIQ model, adapter, and tests in the same integration round.

nemo_retriever/src/nemo_retriever/common/vdb/records.py
charlesbluca: Superseded by the current exact-head `QueryHit.distance` compatibility finding: https://github.com/NVIDIA/NeMo-Retriever/pull/2378#discussion_r3684641298. The original score/ranking anchor no longer describes the published NRL payload.

nemo_retriever/src/nemo_retriever/common/vdb/records.py
charlesbluca: Superseded by the current exact-head `QueryHit.distance` compatibility finding: https://github.com/NVIDIA/NeMo-Retriever/pull/2378#discussion_r3684641298. The original score/ranking anchor no longer describes the published NRL payload.

nemo_retriever/src/nemo_retriever/common/vdb/records.py
charlesbluca: Superseded by the current exact-head `QueryHit.distance` compatibility finding: https://github.com/NVIDIA/NeMo-Retriever/pull/2378#discussion_r3684641298. The original score/ranking anchor no longer describes the published NRL payload.

nemo_retriever/src/nemo_retriever/common/schemas/collections.py
charlesbluca: Final-state clarification for `fe33d24b`: the generic page normalizer and `pagetools.py` were removed rather than retained as a shared helper. Collection projection now handles modality explicitly, and `QueryHit` validates canonical one-based document pages instead of repairing arbitrary values. The focused page-contract tests pass.

nemo_retriever/src/nemo_retriever/service/vectordb_app.py
```diff
@@ -216,82 +644,451 @@ def write_rows(self, rows: list[dict[str, Any]]) -> int:
...
+                    self.expiration_cleanup_enabled
+                    and row.get("status") == "active"
+                    and row.get("expires_at")
+                    and datetime.fromisoformat(str(row["expires_at"])) <= now
+                ):
+                    started = _now()
+                    row.update(
+                        {
+                            "status": "deleting",
+                            "deletion_phase": "drop_table",
```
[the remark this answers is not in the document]
charlesbluca: Confirmed fixed on `fe33d24b`: catalog scans run outside `_write_lock`, with the lock acquired only around individual lifecycle mutations and metric updates. `test_reconcile_catalog_scan_does_not_hold_the_write_lock` passes on the exact PR head.

nemo_retriever/src/nemo_retriever/service/vectordb_app.py
```diff
@@ -216,82 +649,453 @@ def write_rows(self, rows: list[dict[str, Any]]) -> int:
...
+            )
+            retrieval_kwargs: dict[str, Any] = {"top_k": top_k, "hybrid": hybrid}
+            if hybrid:
+                retrieval_kwargs["query_texts"] = query_texts
+            if caps is not None and caps.vector_column and caps.vector_column != "vector":
+                retrieval_kwargs["vector_column_name"] = caps.vector_column
+
+            raw_results = vdb.retrieval(vectors, **retrieval_kwargs)
+            return (
+                normalize_retrieval_results(raw_results, retrieval_mode=mode),
```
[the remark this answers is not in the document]
charlesbluca: Confirmed fixed on `fe33d24b`: collection retrieval leases and resolves the table under the lifecycle lock, then releases it before backend retrieval. `test_collection_searches_run_concurrently` and `test_search_and_collection_delete_share_lifecycle_lock` both pass on the exact PR head.

nemo_retriever/src/nemo_retriever/common/api/util/converters/pagetools.py
charlesbluca: Final-state clarification for `fe33d24b`: the public contract does not blindly add one. Positive document pages remain one-based; audio/video indices and invalid or unknown page provenance produce `page_number=null`, with modality-specific metadata preserved. The direct projection and `QueryHit` validation tests pass.

nemo_retriever/dev/compose/service-mode.compose.yaml
charlesbluca: Confirmed safe on `fe33d24b`: Compose no longer emits `use_graphic_elements`; `ExtractParams` no longer defines it or `graphic_elements_invoke_url`; and `extract_charts` remains supported. `test_graphic_elements_controls_are_removed` passes on the exact PR head.

nemo_retriever/src/nemo_retriever/common/schemas/collections.py
jperez999: why do you need this wrapper then?

nemo_retriever/src/nemo_retriever/common/schemas/collections.py
KyleZheng1284: Would make more sense to have this normalized on AIQ side instead of requiring retriever to do this

KyleZheng1284: looking into adding this fix into the existing 371 PR for AIQ

nemo_retriever/src/nemo_retriever/service/routers/ingest.py
```diff
@@ -204,7 +208,10 @@ def _register_document_under_job(
     document_id: str,
     job_id: str,
     filename: str | None = None,
-) -> None:
+    content_sha256: str | None = None,
```
jperez999, line 241: ### Proposal: centralize collection ingestion context resolution

  The collection-aware ingestion logic is currently repeated across the gateway, worker, and standalone branches of both submit_document_to_job() and submit_whole_document_to_job().

  Each branch independently determines:

  - The authorized scope and collection
  - The associated job
  - Append versus replacement behavior
  - Manifest validation and replay detection
  - The processing attempt ID
  - The stable collection document ID
  - The effective pipeline specification
  - The fields forwarded to workers and VectorDB

  I suggest introducing a typed CollectionIngestContext:
```
  @dataclass(frozen=True, slots=True)
  class CollectionIngestContext:
      job_id: str
      job: JobAggregate | None

      scope: str
      collection_name: str | None
      operation: Literal["append", "replace"]

      attempt_id: str
      stable_document_id: str

      content_sha256: str
      manifest_entry_id: str | None
      pipeline_spec: PipelineSpec | None

      is_replay: bool = False
```
  The identity fields have distinct meanings:

  - attempt_id identifies this pipeline execution. It is used for polling, callbacks, and worker tracking.
  - stable_document_id is the persistent logical document identity inside the collection.
  - manifest_entry_id deterministically identifies one file in an idempotent submission.
  - content_sha256 verifies that a replay contains the same file.
  - operation identifies whether the attempt appends or replaces a document.

  The resolver should account for the service topology:

  - In gateway mode, the job comes from the gateway JobTracker. Scope and collection come from the authorized job. The attempt ID is either a new UUID or the ID from an existing replay record.
  - In standalone mode, the job comes from the local JobTracker, with the same identity and authorization rules as the gateway.
  - In worker mode, the job context, scope, collection, stable ID, and attempt ID come from trusted gateway-injected internal headers. The worker should not generate a new stable ID or trust a public scope header.

  Conceptually:
```
  def resolve_collection_ingest_context(
      request: Request,
      *,
      job_id: str,
      filename: str,
      content_sha256: str,
      manifest_entry_id: str | None,
      requested_pipeline: PipelineSpec | None,
  ) -> CollectionIngestContext:
      if _is_worker(request):
          return _context_from_gateway_headers(
              request,
              job_id=job_id,
              filename=filename,
              content_sha256=content_sha256,
          )

      job = _require_job(job_id, request)
      _validate_manifest_entry(
          job,
          manifest_entry_id,
          filename,
          content_sha256,
      )
      _validate_collection_pipeline_spec(job, requested_pipeline)

      return _context_from_local_job(
          job,
          content_sha256=content_sha256,
          manifest_entry_id=manifest_entry_id,
          pipeline_spec=requested_pipeline,
      )
```
  Replay registration should remain an explicit second step because it mutates JobTracker:
```
  context = resolve_collection_ingest_context(...)
  acceptance = register_ingest_attempt(context)
```
  That operation could return:
```
  @dataclass(frozen=True, slots=True)
  class IngestAcceptance:
      context: CollectionIngestContext
      record: DocumentRecord
      created: bool
```
  When created=False, the route returns the original acceptance without enqueueing duplicate processing.

  The context can then construct a WorkItem consistently:
```
  def make_work_item(
      context: CollectionIngestContext,
      *,
      payload: bytes,
      filename: str,
      callback_url: str | None,
  ) -> WorkItem:
      return WorkItem(
          id=context.attempt_id,
          payload=payload,
          filename=filename,
          job_id=context.job_id,
          scope=context.scope,
          collection_name=context.collection_name,
          operation=context.operation,
          content_sha256=context.content_sha256,
          storage_document_id=context.stable_document_id,
          pipeline_spec=(
              context.pipeline_spec.model_dump(mode="json")
              if context.pipeline_spec
              else None
          ),
          callback_url=callback_url,
      )
```
  Both ingestion routes could then follow the same high-level flow:
```
  file_bytes = await file.read()
  content_sha256 = hashlib.sha256(file_bytes).hexdigest()

  context = resolve_collection_ingest_context(...)
  acceptance = register_ingest_attempt(context)

  if not acceptance.created:
      return response_from_existing(acceptance.record)

  item = make_work_item(
      acceptance.context,
      payload=file_bytes,
      filename=file.filename,
      callback_url=...,
  )
  await enqueue(item)

  return accepted_response(acceptance)
```
  Endpoint-specific behavior would remain limited to:

  - File classification
  - Realtime versus batch routing
  - Response model
  - Page-count routing
  - Metrics

  This would provide:

  - One stable-document identity policy
  - One manifest and replay policy
  - One trusted gateway-to-worker boundary
  - Typed propagation instead of loosely related arguments and headers
  - Smaller ingestion routes
  - Focused unit tests for identity and topology resolution
  - Less risk that /document, /whole, gateway, and standalone behavior diverge

This should help reduce 150-200 Lines of code.

nemo_retriever/src/nemo_retriever/service/routers/ingest.py
```diff
@@ -616,7 +699,7 @@ def _aggregate_to_response(agg, *, documents: list[dict[str, Any]] | None = None
     status_code=201,
     summary="Create a new ingestion job aggregate",
 )
-async def create_job(request: Request, response: Response, body: JobCreateRequest) -> JobCreatedResponse:
+async def create_job(request: Request, response: Response, body: JobCreateRequest) -> JobCreatedResponse | Response:
```
jperez999, line 826: ### Proposal: extract collection validation and idempotency fingerprint construction

  create_job() currently combines several responsibilities:

  - Resolving the authorized scope
  - Verifying that VectorDB is enabled
  - Looking up the target collection
  - Confirming that the collection is active
  - Looking up the target document for replacement
  - Translating VectorDB errors into gateway responses
  - Constructing the idempotency fingerprint
  - Creating tracing state
  - Registering the job

  The collection validation and fingerprint logic occupy a substantial part of create_job() (https://github.com/NVIDIA/NeMo-Retriever/blob/fe33d24b5410737f7660fd4b2adb79d7b2cc644a/nemo_retriever/src/nemo_retriever/service/routers/ingest.py#L702-L831) and make the job-registration behavior difficult
  to read in isolation.

  I suggest extracting two separate operations:
```
  await validate_collection_job_target(
      request,
      scope=scope,
      job_request=body,
  )

  fingerprint = build_job_idempotency_fingerprint(body)
```
  #### 1. Extract collection target validation

  The validation helper should own all remote VectorDB resource checks:
```
  async def validate_collection_job_target(
      request: Request,
      *,
      scope: str,
      job_request: JobCreateRequest,
  ) -> None:
      if job_request.collection_name is None:
          return

      config = request.app.state.config
      if not config.vectordb.enabled:
          raise HTTPException(
              status_code=404,
              detail="VectorDB is not enabled in the service configuration",
          )

      headers = {
          "X-NRL-Scope": scope,
          **internal_auth_headers(config.vectordb.internal_api_token),
      }

      base_url = config.vectordb.vectordb_url.rstrip("/")
      collection_url = (
          f"{base_url}/v1/collections/{job_request.collection_name}"
      )

      async with httpx.AsyncClient(timeout=30.0) as client:
          collection = await _get_vectordb_resource(
              client,
              collection_url,
              headers=headers,
              unavailable_detail="Failed to validate collection",
          )

          if collection.get("status") != "active":
              raise HTTPException(
                  status_code=409,
                  detail="Collection is not active",
              )

          if (
              job_request.operation == "replace"
              and job_request.target_document_id is not None
          ):
              await _get_vectordb_resource(
                  client,
                  (
                      f"{collection_url}/documents/"
                      f"{job_request.target_document_id}"
                  ),
                  headers=headers,
                  unavailable_detail="Failed to validate replacement document",
              )
```
  The repeated HTTP request and response translation can also be centralized:
```
  async def _get_vectordb_resource(
      client: httpx.AsyncClient,
      url: str,
      *,
      headers: dict[str, str],
      unavailable_detail: str,
  ) -> dict[str, Any]:
      try:
          response = await client.get(url, headers=headers)
      except httpx.HTTPError as exc:
          raise HTTPException(
              status_code=502,
              detail=unavailable_detail,
          ) from exc

      if response.status_code >= 400:
          detail = response.text[:1000] or "VectorDB request failed"
          raise HTTPException(
              status_code=response.status_code,
              detail=detail,
          )

      try:
          return response.json()
      except ValueError as exc:
          raise HTTPException(
              status_code=502,
              detail="VectorDB returned malformed JSON",
          ) from exc
```
  This has several benefits:

  - Collection and replacement-document validation use one AsyncClient.
  - Scope and internal authentication headers are constructed once.
  - Transport errors are handled consistently.
  - Malformed VectorDB responses are handled explicitly.
  - create_job() no longer contains internal VectorDB URL construction.
  - Additional job operations can reuse the same validation boundary.

  The helper name should make clear that this is validation, not authorization. Authorization has already happened when the scope is attached to the request.

  #### 2. Extract fingerprint construction

  The current fingerprint is assembled inline inside create_job():
```
  fingerprint = hashlib.sha256(
      json.dumps(
          {
              ...
          },
          sort_keys=True,
          separators=(",", ":"),
      ).encode()
  ).hexdigest()
```
  This is security- and correctness-sensitive logic because it defines when two uses of the same idempotency key are considered the same request.

  It should have a dedicated function:
```
  _JOB_FINGERPRINT_VERSION = 1


  def build_job_idempotency_fingerprint(
      request: JobCreateRequest,
  ) -> str:
      payload = {
          "version": _JOB_FINGERPRINT_VERSION,
          "request": request.model_dump(
              mode="json",
              exclude={"idempotency_key"},
          ),
      }

      canonical = json.dumps(
          payload,
          sort_keys=True,
          separators=(",", ":"),
          ensure_ascii=False,
      )

      return hashlib.sha256(canonical.encode("utf-8")).hexdigest()
```
  Using model_dump() instead of manually selecting fields has an important advantage: new semantically relevant request fields are less likely to be accidentally omitted from the fingerprint.

  The idempotency key itself should be excluded because it identifies the stored fingerprint; it is not part of the request equivalence comparison.

  A fingerprint version is useful because changing the canonical representation later would otherwise silently change idempotency behavior.

  #### Current fingerprint omission

  The current inline fingerprint includes:

  - expected_documents
  - collection_name
  - operation
  - target_document_id
  - metadata
  - document_manifest

  It does not include:

  - label
  - retain_results

  retain_results affects job behavior, so reusing an idempotency key with a different value should probably conflict rather than silently return the original job.

  Whether label participates is a product-contract decision. If it is intentionally excluded because it is only descriptive, that should be explicit:
```
  _FINGERPRINT_EXCLUDED_FIELDS = {
      "idempotency_key",
      "label",  # Descriptive only; does not change processing semantics.
  }
```
  An explicit exclusion is preferable to maintaining a manual allowlist that can silently omit future fields.

  #### Resulting create_job() flow

  After extraction, the core route becomes much easier to follow:
```
  async def create_job(
      request: Request,
      response: Response,
      body: JobCreateRequest,
  ) -> JobCreatedResponse:
      tracker = get_job_tracker()
      if tracker is None:
          raise HTTPException(
              status_code=503,
              detail="Job tracker not available",
          )

      scope = authorized_scope(request)

      await validate_collection_job_target(
          request,
          scope=scope,
          job_request=body,
      )

      fingerprint = build_job_idempotency_fingerprint(body)

      job_id = uuid.uuid4().hex
      trace_context, trace_id = create_job_trace_context(
          request,
          expected_documents=body.expected_documents,
      )

      job = tracker.register_job(
          job_id,
          expected_documents=body.expected_documents,
          label=body.label,
          metadata=body.metadata,
          retain_results=body.retain_results,
          trace_id=trace_id,
          trace_context=trace_context,
          collection_name=body.collection_name,
          scope=scope,
          operation=body.operation,
          target_document_id=body.target_document_id,
          idempotency_key=body.idempotency_key,
          idempotency_fingerprint=fingerprint,
          document_manifest=[
              entry.model_dump(mode="json")
              for entry in body.document_manifest
          ],
      )

      return job_created_response(job, requested_job_id=job_id)
```
  This separates three concepts that are currently interleaved:

  1. The target collection and document exist and are usable.
  2. The request has a deterministic idempotency identity.
  3. The job is registered and traced.

this extraction could remove approximately 30–60 lines directly

nemo_retriever/src/nemo_retriever/service/routers/ingest.py
jperez999, line 1102: ### Proposal: consolidate the shared document-submission flow

  submit_document_to_job() and submit_whole_document_to_job() currently repeat much of the same collection-aware submission sequence across their gateway and standalone/worker branches.

  The repeated sequence includes:

  - Reading the uploaded file
  - Calculating its size and SHA-256 digest
  - Validating the file against the job manifest
  - Resolving or allocating the processing attempt ID
  - Resolving the stable collection document ID
  - Registering the attempt with JobTracker
  - Detecting idempotent replays
  - Constructing a WorkItem
  - Enqueueing the work
  - Constructing the accepted response

  I suggest extracting this into a shared preparation, registration, and work-item construction flow.

  #### 1. Prepare the upload once

  File reading, size calculation, and hashing should happen in one helper:
```
  @dataclass(frozen=True, slots=True)
  class PreparedUpload:
      payload: bytes
      filename: str
      content_sha256: str
      size_bytes: int


  async def prepare_upload(file: UploadFile) -> PreparedUpload:
      payload = await file.read()

      return PreparedUpload(
          payload=payload,
          filename=file.filename or "",
          content_sha256=hashlib.sha256(payload).hexdigest(),
          size_bytes=len(payload),
      )
```
  This gives the remaining ingestion flow one canonical representation of the uploaded file.

  It also avoids independently calculating:

  - file_bytes
  - file_size
  - content_sha256
  - The effective filename

  in each gateway and worker branch.

  Upload-size enforcement should still happen before fully buffering the file when the framework provides the size. The helper is intended to centralize the post-validation read and hash, not bypass the existing upload-size guard.

  #### 2. Resolve the ingestion context

  After preparing the file, resolve the job, scope, collection, operation, pipeline, and identity information:
```
  context = resolve_collection_ingest_context(
      request,
      job_id=job_id,
      filename=upload.filename,
      content_sha256=upload.content_sha256,
      manifest_entry_id=manifest_entry_id,
      requested_pipeline=validated_spec,
  )
```
  This context should establish:

  - The authoritative job
  - The authorized scope
  - The target collection
  - Append versus replacement
  - The attempt ID
  - The stable document ID
  - The effective pipeline specification
  - Whether the request came from a gateway, worker, or standalone service

  The topology rules should remain explicit:

  - A gateway or standalone service resolves identity from its JobTracker.
  - A worker uses identity and scope values supplied through the authenticated internal gateway headers.
  - A worker must not generate a different stable document ID.
  - A public scope header must not override the scope already authorized by the gateway.

  #### 3. Validate the manifest in one place

  The manifest check should be part of context resolution for gateway and standalone requests:
```
  _validate_manifest_entry(
      job,
      manifest_entry_id,
      upload.filename,
      upload.content_sha256,
  )
```
  That check should ensure:

  - A manifest entry is supplied when the job declared a manifest.
  - The entry exists in the job manifest.
  - The filename matches.
  - The content hash matches.
  - Reuse of the entry with different content produces a conflict.

  Worker requests should not independently reinterpret the public manifest. The gateway has already validated the upload before forwarding it.

  If the worker needs the manifest entry for diagnostics, the gateway can forward it as authenticated internal metadata.

  #### 4. Allocate attempt and stable document IDs consistently

  The shared identity logic should preserve the distinction between an ingestion attempt and a persistent collection document.

  For an ordinary collection append:

  attempt_id         = new processing-attempt UUID
  stable_document_id = new collection-document UUID

  For replacement:

  attempt_id         = new processing-attempt UUID
  stable_document_id = target document ID

  For legacy non-collection ingestion:

  attempt_id         = new processing-attempt UUID
  stable_document_id = attempt ID

  For a worker:

  attempt_id         = gateway-provided attempt ID
  stable_document_id = gateway-provided stable document ID

  This logic currently appears in multiple branches. Centralizing it ensures /document and /whole cannot assign different identities for the same operation.

  #### 5. Register the attempt and detect replay atomically

  Registration should be a separate explicit operation because it mutates the job tracker:
```
  acceptance = register_ingest_attempt(
      context,
      filename=upload.filename,
      content_sha256=upload.content_sha256,
      manifest_entry_id=manifest_entry_id,
  )
```
  The result could be represented as:
```
  @dataclass(frozen=True, slots=True)
  class IngestAcceptance:
      context: CollectionIngestContext
      record: DocumentRecord | None
      created: bool
```
  created=True means a new attempt was registered and should be enqueued.

  created=False means the manifest entry was already accepted. In that case, the route should return the original acceptance without:

  - Consuming additional job capacity
  - Generating another stable document ID
  - Creating another WorkItem
  - Starting duplicate pipeline processing

  For worker requests, record can be None because the gateway owns the authoritative job tracker.

  #### 6. Construct WorkItem through one helper

  The gateway, standalone service, /document, and /whole routes currently propagate the same collection fields separately.

  A shared constructor can make this consistent:
```
  def make_ingest_work_item(
      acceptance: IngestAcceptance,
      upload: PreparedUpload,
      *,
      pool_type: PoolType,
      callback_url: str | None,
      callback_headers: dict[str, str],
      retain_results: bool,
  ) -> WorkItem:
      context = acceptance.context

      return WorkItem(
          id=context.attempt_id,
          payload=upload.payload,
          filename=upload.filename,
          callback_url=callback_url,
          callback_headers=callback_headers,
          job_id=context.job_id,
          pipeline_spec=(
              context.pipeline_spec.model_dump(mode="json")
              if context.pipeline_spec is not None
              else None
          ),
          retain_results=retain_results,
          scope=context.scope,
          collection_name=context.collection_name,
          operation=context.operation,
          content_sha256=upload.content_sha256,
          storage_document_id=context.stable_document_id,
      )
```
  This prevents errors where one branch forgets to propagate fields such as:

  - scope
  - collection_name
  - operation
  - content_sha256
  - storage_document_id
  - pipeline_spec
  - Internal callback authentication

  It also gives tests one place to verify the gateway-to-worker contract.

  #### 7. Share accepted-response construction

  /document and /whole return different Pydantic response models, but the identity and acceptance data used to construct those responses are largely the same.

  The shared flow can produce an endpoint-neutral result:
```
  @dataclass(frozen=True, slots=True)
  class AcceptedDocument:
      job_id: str
      attempt_id: str
      stable_document_id: str
      content_sha256: str
      filename: str
      size_bytes: int
      created_at: str
      replayed: bool
```
  Each endpoint can then perform a thin response conversion:
```
  def document_accepted_response(
      accepted: AcceptedDocument,
  ) -> IngestAccepted:
      return IngestAccepted(
          document_id=accepted.stable_document_id,
          attempt_id=accepted.attempt_id,
          job_id=accepted.job_id,
          content_sha256=accepted.content_sha256,
          status="accepted",
          created_at=accepted.created_at,
      )

  def whole_document_accepted_response(
      accepted: AcceptedDocument,
  ) -> DocumentIngestAccepted:
      return DocumentIngestAccepted(
          document_id=accepted.stable_document_id,
          attempt_id=accepted.attempt_id,
          filename=accepted.filename,
          file_size_bytes=accepted.size_bytes,
          content_sha256=accepted.content_sha256,
          status="accepted",
          created_at=accepted.created_at,
      )
```
  This preserves the existing public response models while centralizing the source data.

  #### Resulting route structure

  After extraction, submit_document_to_job() could be reduced conceptually to:
```
  async def submit_document_to_job(...):
      meta = parse_ingest_metadata(metadata)
      _check_upload_size(file, request)

      classification = classify_and_validate(file, meta)
      validated_spec = _resolve_pipeline_spec(request, meta)
      upload = await prepare_upload(file)

      route = _route_by_page_count(
          upload.payload,
          meta,
          file_category=classification.category,
      )

      context = resolve_collection_ingest_context(
          request,
          job_id=job_id,
          filename=upload.filename,
          content_sha256=upload.content_sha256,
          manifest_entry_id=manifest_entry_id,
          requested_pipeline=validated_spec,
      )

      acceptance = register_ingest_attempt(
          context,
          filename=upload.filename,
          content_sha256=upload.content_sha256,
          manifest_entry_id=manifest_entry_id,
      )

      if not acceptance.created:
          return document_accepted_response(
              accepted_document_from_replay(
                  acceptance,
                  upload,
              )
          )

      item = make_ingest_work_item(
          acceptance,
          upload,
          pool_type=route,
          callback_url=...,
          callback_headers=...,
          retain_results=...,
      )

      await enqueue_ingest_work(request, route, item)

      accepted = accepted_document_from_new_attempt(
          acceptance,
          upload,
      )
      record_ingest_metrics(request, classification, accepted)

      return document_accepted_response(accepted)
```
  submit_whole_document_to_job() would follow the same flow but would:

  - Always use PoolType.BATCH
  - Return DocumentIngestAccepted
  - Retain its endpoint-specific classification and metrics labels

  #### What should remain endpoint-specific

  The extraction should not force all ingestion routes into one oversized generic function.

  The following behavior should remain in the endpoint:

  - Parsing the endpoint-specific form data
  - File classification
  - Media dependency enforcement
  - Page-count routing for /document
  - Fixed batch routing for /whole
  - The public response model
  - Endpoint-specific metrics
  - Endpoint-specific tracing span names

  The shared helper should own identity, manifest, job registration, work-item propagation, and replay behavior—not every detail of the endpoint.

  #### Expected benefit

  This change would provide:

  - One file hashing implementation
  - One manifest-validation policy
  - One stable-ID allocation policy
  - One replay-registration path
  - One WorkItem propagation contract
  - One source of accepted-response identity data
  - Less duplication across /document, /whole, gateway, worker, and standalone modes
  - Less risk that one branch omits a collection or authentication field
 
this portion of the ingestion consolidation could remove approximately 80–140 production lines on its own. Combined with the proposed CollectionIngestContext resolver and shared pipeline forwarding, the total ingestion-related reduction could reach approximately 150–250 lines.

nemo_retriever/src/nemo_retriever/common/vdb/adt_vdb.py
```diff
@@ -235,6 +295,131 @@ def put(self, records: list, **kwargs: Any) -> dict[str, Any]:
             "in-place stable-key puts are not supported by this VDB backend."
         )
 
+    def create_collection(
```
jperez999, line 299: without backwards compatibility don't these need to be abstract methods now as well?

nemo_retriever/src/nemo_retriever/service/routers/ingest.py
jperez999: we can handle this refactor in subsequent PRs

jperez999 (changes requested):
This looks good to me, aside from the greptile comments that have not been resolved. I had a few questions about the abstract_methods decorator not being used for the new function in adt_vdb. We should probably require those, so everything is on the new setup.

nemo_retriever/src/nemo_retriever/common/schemas/collections.py
KyleZheng1284: changes in 371

nemo_retriever/src/nemo_retriever/common/schemas/collections.py
KyleZheng1284: The wrapper adapts the optional expires_at API field to the shared timestamp converter and is reused by both create and update validation. Keeping that schema-specific optional handling here avoids duplicating it or broadening the converter’s contract

nemo_retriever/src/nemo_retriever/service/routers/ingest.py
KyleZheng1284: collection and document lookups were byte-identical httpx calls, so they now share one scoped-read helper

nemo_retriever/src/nemo_retriever/common/vdb/adt_vdb.py
KyleZheng1284: moved to abstract methods

jperez999 (approved):

charlesbluca (approved):

PR #2380, opened 2026-07-20
Support local in-process vLLM LLMs for agentic retrieval

### LINKED ISSUE #2341: [FEA]: Enable agentic retrieval graph to automatically pull and start local LLM

New Feature

Currently preventing usage

To be populated as needed

To be populated as needed

### Additional context

_No response_

mahikaw:
## Description
  - Add an in-process vLLM-backed LLM adapter for agentic retrieval.
  - Default agentic retrieval to local vLLM with the supported Nemotron 8B model.
  - Wire local LLM options through module config and harness overrides while keeping the public query CLI minimal.
  - Restrict in-process support to known internal Nemotron model aliases for now.
  - Keep custom/self-hosted LLMs on the existing OpenAI-compatible endpoint path.
  - Update docs and tests for the new defaults and validation behavior.

  ## Validation
  - `uv run pytest tests/test_agentic_local_llm.py tests/test_agentic_eval.py tests/test_harness_agentic_eval.py
  tests/test_root_query_cli.py -q`
  - `uvx pre-commit run --all-files`
  - Bounded JP20 local vLLM run completed successfully on 15 queries.

## Checklist


nemo_retriever/src/nemo_retriever/models/local/agent_llm.py
```diff
@@ -0,0 +1,578 @@
...
+
+    def __init__(
+        self,
+        model_path: str,
+        *,
+        device: Optional[str] = None,
+        hf_cache_dir: Optional[str] = None,
+        gpu_memory_utilization: float = 0.8,
+        tensor_parallel_size: int = 1,
+        max_model_len: Optional[int] = None,
```
jperez999: Why not pass an LLMConfig here to standardize on the LLM options that are passed around the codebase? https://github.com/NVIDIA/NeMo-Retriever/blob/20420ba341b4791cab45383a574a5755fc2656d5/nemo_retriever/src/nemo_retriever/service/config.py#L177

nemo_retriever/src/nemo_retriever/models/local/agent_llm.py
```diff
@@ -0,0 +1,578 @@
...
+            return
+
+        try:
+            import torch
+        except ImportError:
+            return
+        if torch.cuda.is_available():
+            torch.cuda.empty_cache()
+
+    def _restore_cuda_visible_devices(self) -> None:
```
jperez999: This seems dangerous, why are you setting this here? And overwriting it... Can you explain the rationale for this logic?

nemo_retriever/src/nemo_retriever/models/local/agent_llm.py
```diff
@@ -0,0 +1,578 @@
...
+    @property
+    def output(self) -> Any:
+        return {"type": "chat_completion", "format": "openai_compatible"}
+
+    @property
+    def input_batch_size(self) -> int:
+        return 1
+
+
+def _cuda_visible_devices_from_device(device: Optional[str]) -> str | None:
```
jperez999: I am ok with checking for the gpu device... but we would error out if GPUs were not available. I would remove the portion of the exception language that talks about what GPU ids to pass.

nemo_retriever/src/nemo_retriever/models/local/agent_llm.py
jperez999: also why is this within the class and the other functions are outside?

nemo_retriever/src/nemo_retriever/models/local/agent_llm.py
```diff
@@ -0,0 +1,578 @@
...
+def _format_tool_message_content(message: Mapping[str, Any], tool_names_by_id: Mapping[str, str]) -> str:
+    tool_call_id = str(message.get("tool_call_id") or "")
+    tool_name = tool_names_by_id.get(tool_call_id, "tool")
+    content = str(message.get("content") or "")
+    if tool_call_id:
+        return f"Tool result for {tool_name} ({tool_call_id}):\n{content}"
+    return f"Tool result for {tool_name}:\n{content}"
+
+
+def _json_tool_prompt(tools: Sequence[Mapping[str, Any]]) -> str:
```
jperez999: This is only used once, and it is only one line, why not inline this?

nemo_retriever/src/nemo_retriever/query/agentic.py
```diff
@@ -250,6 +398,7 @@ def retrieve(self, query_ids: Sequence[str], query_texts: Sequence[str]) -> pd.D
         # small top_k.
         target_top_k = int(self._cfg.top_k)
         per_hop_top_k = max(AGENTIC_RETRIEVER_TOP_K, target_top_k)
+        chat_completion_fn = self._get_chat_completion_fn()
 
         pipeline = (
```
jperez999, line 401: Dont we have an all encompassing operator that pulls all these together? Shouldn't we use that?

nemo_retriever/src/nemo_retriever/query/agentic.py
```diff
@@ -150,14 +165,60 @@ class AgenticRetrievalConfig:
...
+        object.__setattr__(self, "local_llm_backend", local_llm_backend)
+
+        llm_model = str(self.llm_model or "").strip()
+        if not llm_model:
+            if llm_backend == "in_process":
+                llm_model = AGENTIC_LOCAL_LLM_MODEL
+            else:
+                raise ValueError("Agentic retrieval with invoke_url requires a non-empty llm_model.")
+
+        if llm_backend == "in_process":
```
jperez999, line 209: Is this to check that the local llm is part of the supported list?

nemo_retriever/src/nemo_retriever/query/agentic.py
```diff
@@ -178,16 +239,94 @@ def __post_init__(self) -> None:
...
 
 
+def _normalize_agentic_choice(value: object, valid: frozenset[str], *, field_name: str, default: str) -> str:
+    normalized = str(value or default).strip().lower()
+    if normalized not in valid:
+        raise ValueError(f"{field_name} must be one of {sorted(valid)}; got {value!r}")
+    return normalized
+
+
+def _agentic_float_range_value(
```
jperez999: This seems to be non-specific to agentic, its just checking float values and whether they are within a threshold right? If so, this should be generic and in the tools sub module.

jperez999 (changes requested):

jperez999:
Also found these two possible snags, would like to know if these are by design or actual problems (found by codex):

 - [P2] Preserve assistant tool calls for Super-49B follow-up turns — nemo_retriever/src/nemo_retriever/models/local/agent_llm.py:228

    _normalize_messages() leaves assistant messages containing only tool_calls unchanged. The adapter itself creates exactly that shape whenever a local model calls a tool. However, the pinned Super-49B chat template
    (https://huggingface.co/nvidia/Llama-3_3-Nemotron-Super-49B-v1/blob/387156d8d6868c19f3472fa607aa9bfc4f662333/tokenizer_config.json) reads only message["content"] and has no tool_calls handling. Consequently, subsequent ReAct iterations silently omit the previous tool name and arguments, so the
    advertised Super-49B profile receives incomplete conversation history. Serialize assistant tool calls into textual content for templates without native tool-call support.

  - [P2] Keep malformed argument strings malformed — nemo_retriever/src/nemo_retriever/models/local/agent_llm.py:452

    When a model emits invalid string arguments such as "query=foo", _arguments_to_json_string() wraps them with json.dumps(). Downstream json.loads() therefore succeeds but returns a string; both agent operators then call fn_args.get(...), raising AttributeError and aborting the query. Preserve
    the invalid string so existing JSONDecodeError handling can return a tool error, or validate that decoded arguments are mappings.

nemo_retriever/src/nemo_retriever/models/local/agent_llm.py
mahikaw: this one's a service specific config with fields like api_base, api_key etc. so won't be directly suitable for here. adding a similar config for localLLM config.

nemo_retriever/src/nemo_retriever/query/agentic.py
mahikaw: yes, it intentionally does that before loading vLLM. specially since we use trust_remote_code=True. also needed because tool-call formatting for some models might need special handling.

nemo_retriever/src/nemo_retriever/query/agentic.py
mahikaw: do you mean the agentic->React->RRF->selection operator?

mahikaw: we don't have that currently. tracking it as a tech debt follow-up

nemo_retriever/src/nemo_retriever/query/agentic.py
mahikaw: moving to `query/agentic_options.py` which houses other such helpers.

nemo_retriever/src/nemo_retriever/models/local/agent_llm.py
mahikaw: good point. removed the adapter-level `CUDA_VISIBLE_DEVICES` mutation and the subsequent restore helper entirely. vLLM path will not follow process-level assignment that can be set before process start

mahikaw:
> Also found these two possible snags, would like to know if these are by design or actual problems (found by codex):
> 
> * [P2] Preserve assistant tool calls for Super-49B follow-up turns — nemo_retriever/src/nemo_retriever/models/local/agent_llm.py:228
>   _normalize_messages() leaves assistant messages containing only tool_calls unchanged. The adapter itself creates exactly that shape whenever a local model calls a tool. However, the pinned Super-49B chat template
>   (https://huggingface.co/nvidia/Llama-3_3-Nemotron-Super-49B-v1/blob/387156d8d6868c19f3472fa607aa9bfc4f662333/tokenizer_config.json) reads only message["content"] and has no tool_calls handling. Consequently, subsequent ReAct iterations silently omit the previous tool name and arguments, so the
>   advertised Super-49B profile receives incomplete conversation history. Serialize assistant tool calls into textual content for templates without native tool-call support.
> * [P2] Keep malformed argument strings malformed — nemo_retriever/src/nemo_retriever/models/local/agent_llm.py:452
>   When a model emits invalid string arguments such as "query=foo", _arguments_to_json_string() wraps them with json.dumps(). Downstream json.loads() therefore succeeds but returns a string; both agent operators then call fn_args.get(...), raising AttributeError and aborting the query. Preserve
>   the invalid string so existing JSONDecodeError handling can return a tool error, or validate that decoded arguments are mappings.

Handling these in-process boundary issues as described below:
1. not preserving `tool_calls` when content is empty varies model to model due to different chat templates. We can add a defensive workaround here that serializes assistant `tool-call` metadata into textual `content` before local rendering. there will be a tradeoff here of possible redundant context for templates that handles tool_calls as expected though. but we do gain from not silently dropping prior tool history in templates that don't support content reads. 
2. incase of malformed tool args in llm response, the right behavior can be to fail at adapter/operator boundary rather than letting those errors trickle down into `.get()` calls. The adapter can avoid converting malformed args into valid json strings and operators should validate that decoded args are true json objects. if any of these boundaries detect failures, we send a clear tool error back to the agent so it can retry with valid args.

nemo_retriever/src/nemo_retriever/query/agentic.py
```diff
@@ -137,8 +144,15 @@ class AgenticRetrievalConfig:
     reranker_api_key: str = ""
     local_reranker_backend: str = "vllm"
     embed_modality: str = "text"
+    llm_backend: Optional[str] = None
     llm_model: str = ""
     invoke_url: Optional[str] = None
+    local_llm_backend: str = AGENTIC_LOCAL_LLM_BACKEND
+    local_hf_cache_dir: Optional[str] = None
```
jperez999, line 151: Shouldn't this be wired into the NEMO_RETRIEVER_HF_CACHE_DIR? and then shouldn't it fall back to just use the default HF_CACHE dir?

jperez999 (approved):

nemo_retriever/src/nemo_retriever/query/agentic.py
mahikaw: yeah, we have that fallback chain in practice. https://github.com/NVIDIA/NeMo-Retriever/blob/ac95fbba431b891e3bc4517e290d5c5e4898987f/nemo_retriever/src/nemo_retriever/models/local/agent_llm.py#L128

https://github.com/NVIDIA/NeMo-Retriever/blob/cde5fc4302c879e950e74383450de3c683482a39/nemo_retriever/src/nemo_retriever/models/hf_cache.py#L34-L45

PR #2384, opened 2026-07-20
Add inline text ingestion support

### LINKED ISSUE #2345: [FEA]: Add support for arbitrary text chunks passed as list

New Feature

Significant improvement

Some users have custom ingestion jobs for pulling chunks from JIRA, remote document stores, and web services of many kinds.

Instead of making a custom pipeline for those, it'd be convenient for NRL to make it super easy to ingest lists of text chunks. It should support go through the default text splitting, embedding, and writing to VDB tasks.

Something like:
```
def get_remote_data():
   return ["some chunk", "another chunk"]

ingestor.chunks(get_remote_data())
.embed()
.vdb_upload()
```

### Additional context

_No response_

jioffe502:
## Summary

- add `.texts()` for ingesting raw Python strings without temporary files
- infer each source modality automatically, including `inline://...` text sources
- allow inline text alongside files and buffers through the existing modality planner
- keep chunking configuration on the established `.extract(split_config=...)` interface; no separate `.extract_txt()` or Graph `.split()` API
- reuse the normal split, embed, VDB, and service paths in in-process, batch, and service modes

Closes #2345.

## Validation

- full local unit suite: 2,685 passed, 131 skipped, 12 deselected
- focused ingestion, manifest, splitter, actor, and service suite: 307 passed, 2 skipped, 4 deselected
- 2 local-Ray file/inline parity and mixed-source tests passed
- service worker test passed with default auto-routing and preserved inline source identity
- Black, Flake8, and diff whitespace checks passed on the changed files


nemo_retriever/src/nemo_retriever/ingestor/branch_extraction.py
```diff
@@ -77,7 +78,7 @@ def execute(self) -> Any:
         return self._execute_inprocess()
 
     def _execute_batch(self) -> Any:
-        _ray, cluster_resources = self.ensure_batch_runtime()
+        ray, cluster_resources = self.ensure_batch_runtime()
```
jperez999: why did you make this public? I think it should stay private no?

nemo_retriever/src/nemo_retriever/service/client.py
```diff
@@ -615,19 +641,20 @@ async def ingest_documents(
             upload_sem = asyncio.Semaphore(self._max_concurrency)
             upload_failures: list[tuple[str, str]] = []
 
-            async def _upload_one_file(fpath: Path) -> None:
+            async def _upload_one_file(source: UploadInput) -> None:
```
jperez999, line 669: is there anyway to refactor this internal logic and that from the aingest_document_stream internal into a single helper, that maybe calls a another function that handles the differences?

nemo_retriever/src/nemo_retriever/service/service_ingestor.py
```diff
@@ -207,6 +210,18 @@ def _normalize_files(files: Union[str, List[str], List[Path]]) -> list[Path]:
...
+def _empty_inline_text_dataframe(result_schema: ResultSchema) -> Any:
+    """Return the empty public result schema for inline text."""
+    import pandas as pd
+
+    columns = (
+        ["text", "source_id", "element_type", "page_number"]
+        if result_schema == "compact"
+        else ["text", "content", "path", "page_number", "metadata"]
+    )
+    return pd.DataFrame(columns=columns).astype({"page_number": "int64"})
```
jperez999: what is the difference between this and regular empty_text_chunks_df? Seems confusing to have both. Is this only because of the columns available, and the dtype conversion?

nemo_retriever/src/nemo_retriever/service/service_ingestor.py
```diff
@@ -1167,6 +1190,24 @@ def ingest(self, params: Any = None, **kwargs: Any) -> Any:
...
+            and not self._documents
+            and not self._buffers
+            and not any(text.strip() for text in self._inline_texts)
+        ):
+            self._document_ids.clear()
+            self._last_run_elapsed_s = 0.0
+            self._last_job_id = None
+            result = ServiceIngestResult()
+            if return_results:
+                result.dataframe = _empty_inline_text_dataframe(result_schema)
```
jperez999: Like here, why can't we use just empty dataframe.

jperez999 (approved):

PR #2397, opened 2026-07-23
docs(nemo-retriever): document Nemotron Parse hosted vs self-hosted contract selection

kheiss-uwzoo:
## Summary

Documents Nemotron Parse **hosted Build vs self-hosted** request-contract selection after library-side compatibility landed in #2398.

Hosted NVIDIA Build and self-hosted `nemotron-parse-v1.2` still use different request contracts, but the library now selects the matching contract automatically. Hosted Build is a supported path.

## Changes

- **`prerequisites-support-matrix.md`** — Default NVCF endpoints: annotate the `nemotron-parse` row and add a configuration note covering automatic model/contract selection, when to set `nemotron_parse_invoke_url` / `nemotron_parse_model`, and the mixed-endpoint rule. Uses public `method="nemotron_parse"` (not `extract_method`).
- **`troubleshoot.md`** — model/endpoint mismatch entry for the residual HTTP 400 / text-input failure when an incompatible tagged or `v1.2` model is sent to Build; resolution is omit `nemotron_parse_model` or set `nvidia/nemotron-parse`.

## Scope

Docs-only. Runtime contract selection is already in #2398.

## Self-review

- [x] Reflects post-#2398 supported Build path (not the old unavoidable failure mode)
- [x] Uses `method=` for public API guidance added by this PR
- [x] Page-role / link-CTA check passed (`check-nrl-doc-leakage.ps1`)
- [x] Cross-links and anchors consistent


kheiss-uwzoo:
Holding this as draft pending Charles's engineering compat changes so we handle each ``nemotron-parse`` version properly. Will refresh the doc wording against that PR once it lands (easier to contextualize then).

docs/docs/extraction/prerequisites-support-matrix.md
```diff
@@ -104,11 +104,19 @@ When you call [NVIDIA-hosted NIMs](deployment-options.md#when-to-use-nvidia-host
 | nemotron-ocr-v2 | `https://ai.api.nvidia.com/v1/cv/nvidia/nemotron-ocr-v2` | Chart default OCR SKU; library CPU actors default to this URL when no OCR invoke URL is set. **Local OCR language selectors (`--ocr-lang`, API `ocr_lang`) are not sent on remote requests** — hosted OCR v2 uses its own language behavior |
 | llama-nemotron-embed-vl-1b-v2 | `https://integrate.api.nvidia.com/v1/embeddings` with model ID `nvidia/llama-nemotron-embed-vl-1b-v2` | Core multimodal embedding |
 | llama-nemotron-rerank-vl-1b-v2 | `https://ai.api.nvidia.com/v1/retrieval/nvidia/llama-nemotron-rerank-vl-1b-v2/reranking` | Optional VL reranker |
-| nemotron-parse | `https://integrate.api.nvidia.com/v1/chat/completions` with model ID `nvidia/nemotron-parse` | Optional `extract_method="nemotron_parse"` |
+| nemotron-parse | `https://integrate.api.nvidia.com/v1/chat/completions` with model ID `nvidia/nemotron-parse` | Optional `extract_method="nemotron_parse"`. The hosted Build endpoint currently expects a different request contract than the self-hosted `nemotron-parse-v1.2` NIM; refer to [Nemotron Parse: hosted Build endpoint vs self-hosted NIM](#nemotron-parse-hosted-vs-self-hosted) |
```
charlesbluca: Could we use `method="nemotron_parse"` here? `ExtractParams` exposes `method`, and `.extract()` rejects unknown keyword arguments, so `extract_method` is not a valid public API parameter. This terminology already appears elsewhere, but this PR should avoid adding another customer-facing instance.

docs/docs/extraction/prerequisites-support-matrix.md
```diff
@@ -104,11 +104,19 @@ When you call [NVIDIA-hosted NIMs](deployment-options.md#when-to-use-nvidia-host
...
 
+<a id="nemotron-parse-hosted-vs-self-hosted"></a>
+
+!!! warning "Nemotron Parse: hosted Build endpoint vs self-hosted NIM"
+
+    The self-hosted Helm NIM (`nemotron-parse-v1.2`) and the NVIDIA-hosted Build endpoint (`https://integrate.api.nvidia.com/v1/chat/completions` with model ID `nvidia/nemotron-parse`) currently expect **different request contracts**. The library builds the request for the self-hosted `v1.2` NIM, which includes a text task prompt. The hosted Build endpoint rejects that request with `HTTP 400` (`Content cannot be a plain string. The model does not support text input.`).
+
+    Until the hosted Build endpoint is aligned with `nemotron-parse-v1.2`, run Nemotron Parse against a **self-hosted Helm NIM** or a **local Hugging Face model** rather than the hosted Build endpoint. Default **pdfium**-based PDF extraction is unaffected and works with hosted NIMs. For the error symptom, refer to [Nemotron Parse hosted Build endpoint returns HTTP 400](troubleshoot.md#nemotron-parse-hosted-build-http-400).
```
charlesbluca: Blocker: this describes the behavior before #2398. The deployment contracts still differ, but the library now supports both: Build selects `nvidia/nemotron-parse` and uses the image-only tool-call contract, while self-hosted chat endpoints default to `nvidia/nemotron-parse-v1.2` and use the tagged text-prompt contract.

Could we reframe this as a compatibility/configuration note instead of saying Build fails and must be avoided? It should explain that using Build requires setting `nemotron_parse_invoke_url`, that `nemotron_parse_model` can normally be omitted for automatic selection, and that an explicitly supplied model must match the endpoint. Mixed Build/self-hosted endpoint lists require an explicit model.

docs/docs/extraction/troubleshoot.md
```diff
@@ -133,6 +133,16 @@ Also refer to [What is NeMo Retriever Library?](overview.md) and [Pre-Requisites
...
 
+## Nemotron Parse hosted Build endpoint returns HTTP 400: model does not support text input { #nemotron-parse-hosted-build-http-400 }
+
+When you run PDF extraction with `extract_method="nemotron_parse"` against the NVIDIA-hosted Build endpoint (`https://integrate.api.nvidia.com/v1/chat/completions` with model ID `nvidia/nemotron-parse`), the request can fail with an error similar to the following:
+
+```text
+HTTP 400: Content cannot be a plain string. The model does not support text input.
+```
+
+The hosted Build endpoint currently expects a different request contract than the self-hosted `nemotron-parse-v1.2` NIM that the library targets, so the text task prompt the library sends is rejected. Until the two are aligned, run Nemotron Parse against a self-hosted Helm NIM or a local Hugging Face model, or fall back to the default **pdfium** PDF extraction path. For more information, refer to [Nemotron Parse: hosted Build endpoint vs self-hosted NIM](prerequisites-support-matrix.md#nemotron-parse-hosted-vs-self-hosted).
```
charlesbluca: After #2398, this is no longer the expected result when using the documented Build model. This 400 should now occur only when the caller explicitly selects an incompatible tagged/v1.2 model against Build or otherwise mismatches the model and endpoint contract; the runtime also replaces it with a targeted mismatch hint.

Could we either remove this entry or recast it as model/endpoint mismatch troubleshooting? The resolution for Build is to omit `nemotron_parse_model` for automatic selection or set it to `nvidia/nemotron-parse`, rather than avoid hosted inference.

charlesbluca (changes requested):
Thanks for holding this until #2398 landed. The hosted and self-hosted deployments still use different request contracts, but #2398 added library-side compatibility, making hosted Build a supported path. The inline comments identify the places that still describe the old failure mode, along with one invalid public parameter name.

The PR title and description also need to be refreshed: they currently state that hosted extraction fails and that dual-API support remains future work. They should instead describe the supported contract-selection behavior and any configuration guidance retained by this PR.

Once the docs and PR metadata reflect #2398, I did not find another content blocker in this diff; the anchors and cross-links are consistent.

docs/docs/extraction/prerequisites-support-matrix.md
kheiss-uwzoo: Addressed in ba7adec75 — reframed for post-#2398 contract selection, switched to `method=`, and refreshed the PR title/description.

docs/docs/extraction/prerequisites-support-matrix.md
kheiss-uwzoo: Addressed in ba7adec75 — reframed for post-#2398 contract selection, switched to `method=`, and refreshed the PR title/description.

docs/docs/extraction/troubleshoot.md
kheiss-uwzoo: Addressed in ba7adec75 — reframed for post-#2398 contract selection, switched to `method=`, and refreshed the PR title/description.

docs/docs/extraction/troubleshoot.md
```diff
@@ -133,6 +133,18 @@ Also refer to [What is NeMo Retriever Library?](overview.md) and [Pre-Requisites
 Currently, extraction with Nemotron parse doesn't support image files, only scanned PDFs.
 To work around this issue, convert image files to PDFs before you use `extract_method="nemotron_parse"`.
```
charlesbluca: The new section immediately below correctly uses `method="nemotron_parse"`, while this existing guidance still uses `extract_method`. Because `.extract()` rejects `extract_method` as an unknown argument, can we update this to `method="nemotron_parse"` before merging?

There are also pre-existing `extract_method` references elsewhere in the two touched docs. Since this PR is standardizing the post-#2398 Nemotron Parse guidance, could we normalize those occurrences in the touched pages as part of this change?

docs/docs/extraction/prerequisites-support-matrix.md
kheiss-uwzoo: Normalized remaining `extract_method` references to `method=` on the touched pages in the follow-up commit.

docs/docs/extraction/troubleshoot.md
kheiss-uwzoo: Normalized remaining `extract_method` references to `method=` on the touched pages in the follow-up commit.

charlesbluca:
Thanks for cleaning up the touched pages, and apologies—I was not clear that I meant the remaining Nemotron Parse/general extraction-selector references across the customer-facing docs as well. These can be updated directly from `extract_method` to `method`:

- `docs/docs/extraction/multimodal-extraction.md`: lines 27, 30, 55, 61
- `docs/docs/extraction/overview.md`: lines 26, 30
- `docs/docs/extraction/faq.md`: line 39
- `docs/docs/extraction/customize-extend.md`: line 26
- `nemo_retriever/README.md`: line 73
- `nemo_retriever/helm/README.md`: line 477

One caveat: `nemo_retriever/README.md:129` uses `extract_method="audio"`; please do not mechanically change that to `method="audio"`. Audio uses a different API path (`.extract_audio()`), so that occurrence should be handled separately rather than included in this replacement.

charlesbluca (approved):

PR #2018, opened 2026-05-12
Adding Agentic Retrieval as a new retrieveral mode

mahikaw:
# Agentic retrieval mode for `retriever query`

## Summary
Adds an **LLM-driven agentic retrieval strategy** as an opt-in alternative to the
single dense-retrieval pass on `retriever query`. **Additive** — the standard
retrieval path and outputs are unchanged; agentic mode reuses the existing
`Retriever`/vector DB and is enabled via `--agentic`.

## What's new
- **Agentic retrieval** — `ReActAgentOperator` runs a per-query ReAct loop (issues
  retrieval sub-queries, accumulates candidates across steps, decides when to stop)
  → `RRFAggregatorOperator` fuses across steps (RRF, k=60) → `SelectionAgentOperator`
  applies a source-priority chain (`final_results → RRF → selection → candidate_ranking`)
  and carries `result_source` through to the caller.
- **CLI flags** on `retriever query` — `--agentic`, `--agentic-llm-model`,
  `--agentic-invoke-url`, `--agentic-react-max-steps` (50), `--agentic-backend-top-k`
  (20), `--agentic-text-truncation` (0 = none), `--agentic-reasoning-effort` (high),
  `--agentic-temperature` (0.0). Existing `--rerank`/`--reranker-*` flags are plumbed
  through to the agent's retrieval backend (rerank applies per agent retrieval hop).
- **Docs & tests** — "Agentic retrieval" section in `docs/cli/README.md`;
  `test_agentic_eval.py`, `test_agentic_operators.py`, and `--agentic` CLI tests in
  `test_root_query_cli.py`.

## Results — ViDoRe v3
Benchmarked against the reference agentic pipeline (`retrieval-bench`) under an
**identical, controlled setup** so the comparison isolates the retrieval
framework: same page-level image+text index (`llama-nemotron-embed-vl-1b-v2`
embedder), same agent LLM (`llama-3.3-nemotron-super-49b-v1.5`), same agent
settings (`reasoning_effort=high`, retriever pool depth 20, target top-k 10,
max 50 ReAct steps), full query sets. The retrieval substrate is shared, so the
numbers reflect the agent framework only.

| Domain | recall@10 (ref / this PR) | nDCG@10 (ref / this PR) |
|---|---|---|
| computer_science | 0.7431 / 0.7234 | 0.7396 / 0.7182 |
| energy | 0.6975 / 0.6612 | 0.6369 / 0.6274 |
[...]

PR #2407, opened 2026-07-24
refactor(agentic): make agentic retrieval a self-contained module

Reza-esfandiarpoor:
## Description

Replaces the inline ReAct/selection loops in the graph operators with a
self-contained `nemo_agent` library (`src/nemo_retriever/_agentic/nemo_agent/`),
and reconciles it with main's CLI/config layer and in-process local-vLLM support.

## What's new
- **`nemo_agent` library**: `Agent` (ReAct) + `SelectionAgent`, retrieve/end/think
  tools, Jinja prompts, and an LLM backend layer behind one `_completion_impl`
  seam.
- **Backends**: `litellm` (remote) and `callable` — the latter adapts any
  OpenAI-compatible completion callable, used to wrap main's in-process vLLM
  (`VLLMAgentChatLLM`).
- **Config**: two axes — `llm_backend` (transport: `openai_compatible`/`in_process`,
  inferred from `--agentic-invoke-url`, similar to main, not changed) and `llm_client` (implementation), resolved
  in `AgenticRetrievalConfig.__post_init__` (detected automatically if not set).
- The operators are now thin adapters.

## CLI
- Removed `--agentic-backend-top-k` (`query.agentic_backend_top_k`) — the agent now
  sets retrieval breadth per tool-call.
- Added `--agentic-llm-client` (`callable` local default / `litellm` remote default) - Optional.
- `--agentic-temperature` now defaults to **unset** (endpoint/model default) instead of `0.0`.

## Checklist


nemo_retriever/src/nemo_retriever/query/agentic_options.py
```diff
@@ -15,6 +15,9 @@
 AGENTIC_OPENAI_COMPATIBLE_TEMPERATURE_MAX = 2.0
 AGENTIC_NVIDIA_TEMPERATURE_MAX = 1.0
 
+#: Default LLM client for remote (openai_compatible) agentic retrieval.
+AGENTIC_DEFAULT_CLIENT = "litellm"
```
mahikaw: can we have build inference endpoints as defaults instead.

nemo_retriever/src/nemo_retriever/_agentic/nemo_agent/agent.py
```diff
@@ -0,0 +1,374 @@
...
+    """Configuration for the agent.
+
+    Pure data — LLM and tool *instances* are constructor arguments, not
+    config. ``system_prompt`` is a packaged prompt name or a filesystem path;
+    ``None`` selects the mode default. The loop-policy fields (``max_steps``,
+    ``on_error``, logging and pacing knobs) are inherited from
+    :class:`~nemo_agent.loop.BaseAgentLoopConfig`.
+    """
+
+    mode: Literal["select", "answer"] = "select"
```
mahikaw: since `answer` mode is unused today, i would say we avoid increasing the api surface area here. especially since we want to try and reuse the existing QAGeneration Tasks that already exist in NRL today. Once this becomes a mode we agree with product to want to expose in public surface area, we can add it in a separate PR but until then let's keep this PR mostly for refactoring/clean up tasks.

nemo_retriever/src/nemo_retriever/_agentic/nemo_agent/prompts/templates/05_answer_lean_v1.j2
```diff
@@ -0,0 +1,44 @@
+You are a retrieval-augmented question-answering agent. Your goal is to answer the user's query as accurately and completely as possible, grounded in evidence you actively gather from a document corpus.
```
mahikaw: same as below, i would prefer if we keep this PR only for refactoring/clean up of existing functionality as that's a big chunk on its own.

nemo_retriever/src/nemo_retriever/operators/graph_ops/react_agent_operator.py
```diff
@@ -411,26 +160,81 @@ def __init__(
...
+            reasoning_effort=self._reasoning_effort or None,
+            temperature=self._temperature,
+            parallel_tool_calls=self._parallel_tool_calls,
+            max_completion_tokens=self._max_tokens,
+            drop_params=True,
+        )
+        kwargs = {"completion_fn": self._chat_completion_fn} if self._chat_completion_fn is not None else {}
+        return create_llm(config, **kwargs)
+
+    def _retrieve_adapter(self, query: str, top_k: int) -> List[Dict[str, Any]]:
```
mahikaw, line 205: nit: can we avoid get away without these dual keys for id across operator/agent core boundary. this way we can avoid adding this rename shim in the operators permanently.

nemo_retriever/src/nemo_retriever/operators/graph_ops/react_agent_operator.py
```diff
@@ -411,26 +160,81 @@ def __init__(
...
+            out.append(
+                {
+                    "id": doc_id,
+                    "score": float(doc.get("score", 0.0)),
+                    "text": str(doc.get("text", "")),
+                }
+            )
+        return out
+
+    def _ensure_agent(self) -> Agent:
```
mahikaw, line 222: would this mean each worker (incase of concurrent queries) build their own agent? can this be shared or will that be anti-pattern?

Reza-esfandiarpoor: Each Ray actor creates one instance of the agent but then all the queries in one thread pool across all workers share the same agent.

Reza-esfandiarpoor:
Evaluation results with changes in this PR. The reported numbers are average across three runs.

Dataset | nDCG@10
-- | --
HR | 71.3
Finance En | 73.4
Industrial | 62.3
Pharma | 73.8
CS | 82.8
Energy | 73.6
Physics | 46.9
Finance FR | 54.1
AVG | 67.3


Command to reproduce these results:

```bash
retriever harness run "${DS}_beir" \
    --set query.agentic=true \
    --set query.agentic_llm_model=openai/nvidia/nvidia/nemotron-3-ultra \
    --set query.agentic_invoke_url=${AGENTIC_INVOKE_URL} \
    --set query.agentic_num_concurrent=8 \
    --set query.agentic_reasoning_effort=high \
    --set query.agentic_react_max_steps=200 \
    --set query.agentic_text_truncation=0 \
    --set ingest.embed.embed_model_name=nvidia/nvidia/nemotron-3-embed-1b \
    --set ingest.embed.embed_invoke_url=${EMBED_INVOKE_URL} \
    --set ingest.embed.embed_modality=text \
    --set ingest.embed.embed_granularity=page \
    --set query.embed_model_name=nvidia/nvidia/nemotron-3-embed-1b \
    --set query.embed_invoke_url=${EMBED_INVOKE_URL} \
    --set dataset.path=${CORPUS_PATH} \
    --set evaluation.query_language=english \
    --set ingest.extract.method=pdfium_hybrid \
    --set ingest.extract.ocr_lang=multi \
    --set ingest.extract.use_table_structure=true
```

sahel-sh:
> Evaluation results with changes in this PR. The reported numbers are average across three runs.
> 
> Dataset	nDCG@10
> HR	71.3
> Finance En	73.4
> Industrial	62.3
> Pharma	73.8
> CS	82.8
> Energy	73.6
> Physics	46.9
> Finance FR	54.1
> AVG	67.3
> Command to reproduce these results:
> 
> ```shell
> retriever harness run "${DS}_beir" \
>     --set query.agentic=true \
>     --set query.agentic_llm_model=openai/nvidia/nvidia/nemotron-3-ultra \
>     --set query.agentic_invoke_url=${AGENTIC_INVOKE_URL} \
>     --set query.agentic_num_concurrent=8 \
>     --set query.agentic_reasoning_effort=high \
>     --set query.agentic_react_max_steps=200 \
>     --set query.agentic_text_truncation=0 \
>     --set ingest.embed.embed_model_name=nvidia/nvidia/nemotron-3-embed-1b \
>     --set ingest.embed.embed_invoke_url=${EMBED_INVOKE_URL} \
>     --set ingest.embed.embed_modality=text \
>     --set ingest.embed.embed_granularity=page \
>     --set query.embed_model_name=nvidia/nvidia/nemotron-3-embed-1b \
>     --set query.embed_invoke_url=${EMBED_INVOKE_URL} \
>     --set dataset.path=${CORPUS_PATH} \
>     --set evaluation.query_language=english \
>     --set ingest.extract.method=pdfium_hybrid \
>     --set ingest.extract.ocr_lang=multi \
>     --set ingest.extract.use_table_structure=true
> ```

Thank you @Reza-esfandiarpoor this result summary is really helpful to confirm consistency with what we currently have in https://docs.google.com/spreadsheets/d/1TPcVZa_1dxnwrNltkMq63lmgGtOE_cij7TTjkX2te8w/edit?gid=0#gid=0 and reported in the release, just a couple of questions:
1- IIRC, we agreed on switching the ultera nvfp4 model, these runs are using the nemotron-3-ultra?
2- would it be possible to run a case with vllm reranking to make sure it still works after this MR is merged?
3- the ingestion here is using ocr? would our ingestion pipeline still work for embedding the documents and queries?
4- Does the vllm adaptor for retriever still work? This is needed for running the nemotron nvfp4 embeding model

Reza-esfandiarpoor:
@sahel-sh

> Model used for the runs

Thanks for pointing it out. Yes, this is not the nvfp4 version. But it should be okay in this case. This run is to just show that the refactor in this PR is correct.

> Ingestion + implementation of retrieval and reranking

The scope of this PR is limited to the implementation of the agent itself. Ingestion and the implementation of retrieval methods are provided by NRL (this code just wraps whatever retriever function it gets from NRL in an agent tool). How and where different reranking and embedding models are implemented and using what backend (e.g., vLLM) is outside of the scope of this PR. We can discuss that for future changes.

nemo_retriever/src/nemo_retriever/operators/graph_ops/react_agent_operator.py
Reza-esfandiarpoor: This is the same check from before this PR. I prefer to keep NRL-specific details the same as before and just focus on the agent implementation.

mahikaw (approved):
lgtm. thank you @Reza-esfandiarpoor 
Assuming benchmark parity with current NRL agentic implementation as current numbers posted are using non-vl embedding models. previously reported benchmarks use the `llama-nemotron-embed-vl-1b-v2` here:  https://github.com/NVIDIA/NeMo-Retriever/pull/2018

Reza-esfandiarpoor:
> lgtm. thank you @Reza-esfandiarpoor Assuming benchmark parity with current NRL agentic implementation as current numbers posted are using non-vl embedding models. previously reported benchmarks use the `llama-nemotron-embed-vl-1b-v2` here: #2018

Please find the evaluation results for the agentic retrieval with the agent from the main branch (e.g., same LLM, embedding model, etc.) below.  
For a fair comparison, both pipelines (main branch and refactored agent) are run with the default temperature of the LLM provider (i.e., no explicit temperature is set).

The results for this PR are very close and even slightly higher than the main branch pipeline on average.


Dataset | nDCG@10
-- | --
HR | 70.9
Finance En | 72.9
Industrial | 62.1
Pharma | 73.1
CS | 82.3
Energy | 73.5
Physics | 44.9
Finance FR | 54.1
AVG | 66.7

PR #2416, opened 2026-07-30
Expose agentic retrieval via MCP and HTTP service endpoints

mahikaw:
## Description

- Add server-owned AgenticConfig and thin service adapters so existing agentic_query_documents() can run in service mode without changing AgenticRetriever internals.

- Expose POST /v1/agentic/query on the VectorDB app (execution) and gateway proxy; add config-gated MCP tool agentic_query alongside unchanged plain query.

- Wire Helm/retriever-service.yaml so gateway gets enablement + timeout from ConfigMap, and the VectorDB deployment gets --agentic* CLI flags + API key when a remote LLM URL is set.

- Service mode requires remote invoke_url + llm_model (and remote embed); request body stays {query, top_k} with server-owned ReAct knobs. Hits remain slim (rank / doc_id / result_source).

## Checklist

### DIFF
--- a/nemo_retriever/src/nemo_retriever/service/agentic_query.py
+++ b/nemo_retriever/src/nemo_retriever/service/agentic_query.py
@@ -2,25 +2,57 @@
 # All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-"""Service boundary for the existing in-process agentic retrieval workflow."""
+"""Service boundary for in-process agentic retrieval over the VectorDB query API."""
 
 from __future__ import annotations
+
+from typing import Any
 
 from nemo_retriever.query.options import (
     QueryAgenticOptions,
     QueryEmbedOptions,
-    QueryRequest,
+    QueryRequest as WorkflowQueryRequest,
     QueryRetrievalOptions,
     QueryStorageOptions,
 )
 from nemo_retriever.query.workflow import agentic_query_documents
 from nemo_retriever.service.config import AgenticConfig
-from nemo_retriever.service.query_schema import AgenticQueryRequest, AgenticQueryResponse
+from nemo_retriever.service.query_schema import QueryResponse, QueryResult
+
+
+def agentic_ranked_to_hits(ranked: list[dict[str, Any]]) -> list[dict[str, Any]]:
+    """Map agentic document ranks onto the ``/v1/query`` hits envelope.
+
+    Agentic retrieval selects documents (``doc_id``), not chunks. Each hit places
+    ``doc_id`` in ``source``, records ``result_source`` / ``rank`` under
+    ``metadata``, and leaves chunk-level fields (``text``, ``page_number``,
+    scores, …) unset.
+    """
+    hits: list[dict[str, Any]] = []
+    for item in ranked:
+        doc_id = str(item.get("doc_id") or "") or None
+        hits.append(
+            {
+                "text": None,
+                "metadata": {
+                    "result_source": item.get("result_source"),
+                    "rank": item.get("rank"),
+                },
+                "source": doc_id,
+                "source_id": None,
+                "path": None,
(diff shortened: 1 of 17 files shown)


nemo_retriever/src/nemo_retriever/service/query_schema.py
```diff
@@ -36,6 +41,23 @@ def hits_by_query(self, *, expected_results: int | None = None) -> list[list[dic
         return [result.hits for result in self.results]
 
 
+class AgenticQueryRequest(BaseModel):
```
jperez999: How is this different from a regular query response? If its not different why do we need another one?

nemo_retriever/src/nemo_retriever/service/query_schema.py
```diff
@@ -36,6 +41,23 @@ def hits_by_query(self, *, expected_results: int | None = None) -> list[list[dic
...
+    """One query for the server-configured agentic retrieval pipeline."""
+
+    query: str = Field(min_length=1, max_length=MAX_AGENTIC_QUERY_CHARS)
+    top_k: int = Field(default=5, ge=1, le=1000)
+
+
+class AgenticQueryResult(BaseModel):
```
jperez999: Same question as above here, why do we need these new models, if the query is still producing a ranked list of top_k chunks?

nemo_retriever/src/nemo_retriever/service/vectordb_app.py
```diff
@@ -472,6 +499,82 @@ async def query(req: QueryRequest) -> QueryResponse | EvidenceQueryResponse:
 
         return QueryResponse(results=[QueryResult(hits=hits) for hits in hits_per_query])
 
+    @app.post("/v1/agentic/query", response_model=AgenticQueryResponse, tags=["query"])
+    async def agentic_query(req: AgenticQueryRequest) -> AgenticQueryResponse:
```
jperez999: explain the need for a separate api here. Why cant I just use the regular query api with the agentic flag and have that use the other graph and have it execute?

jperez999 (changes requested):

nemo_retriever/src/nemo_retriever/service/vectordb_app.py
mahikaw: added this to make backend selection explicit in the api call itself without making it combinatorial using flags. 
if we decide to converge it to a single endpoint, i propose we still keep the option of not serving agentic backend open. this will be in favor of slim service spin ups when agentic is not needed.

nemo_retriever/src/nemo_retriever/service/query_schema.py
mahikaw: right now its slightly different shape as agentic hits are not enriched with metadata. even though aim is to converge both hits to the classic nrl shape, PR is not checked in yet: https://github.com/NVIDIA/NeMo-Retriever/pull/2274

nemo_retriever/src/nemo_retriever/service/agentic_query.py
```diff
@@ -0,0 +1,116 @@
...
+    """Map agentic document ranks onto the ``/v1/query`` hits envelope.
+
+    Agentic retrieval selects documents (``doc_id``), not chunks. Each hit places
+    ``doc_id`` in ``source``, records ``result_source`` / ``rank`` under
+    ``metadata``, and leaves chunk-level fields (``text``, ``page_number``,
+    scores, …) unset.
+    """
+    hits: list[dict[str, Any]] = []
+    for item in ranked:
+        doc_id = str(item.get("doc_id") or "") or None
```
jperez999: In what scenario are we allowed to hit None here? Have you seen that before? Should we maybe raise an error here if that happens?

jperez999 (approved):

PR #2417, opened 2026-07-31
fix(service): wire Nemotron Parse endpoint into extraction workers

girihemant19:
## Summary

Service-mode clients can select `method="nemotron_parse"`, but the Python
service configuration does not currently expose or inject the remote Nemotron
Parse endpoint and model into worker `ExtractParams`.

Without the remote URL, the Parse operator receives an empty invoke URL and
falls back to local vLLM loading, even when the operator intends to use an
NVIDIA-hosted or self-hosted Parse NIM.

This change:

- adds server-owned `nemotron_parse_invoke_url` and
  `nemotron_parse_model` fields to `NimEndpointsConfig`;
- exposes the fields in the bundled service YAML;
- injects the configured method, endpoint and model into worker
  `ExtractParams`;
- prevents clients from overriding the server-owned Parse model; and
- adds tests for configuration propagation, policy enforcement and
  server-owned worker merging.

## Testing

- Focused service tests pass.
- Self-hosted service-mode validation completed successfully against
  `nvidia/nemotron-parse-v1.2`.
- The service client sent only `method="nemotron_parse"` and extraction flags.
- The server injected the Parse URL/model.
- One-page end-to-end extraction completed with one result row, no failures,
  and `nemotron_parse_v1_2.error = None`.

### DIFF
--- a/nemo_retriever/helm/templates/configmap.yaml
+++ b/nemo_retriever/helm/templates/configmap.yaml
@@ -11,6 +11,7 @@ inherits the NIMService resource name, so the mapping is fixed:
   page_elements                          -> nemotron-page-elements-v3                /v1/page-elements
   table_structure                        -> nemotron-table-structure-v1              /v1/table-structure
   ocr                                    -> nemotron-ocr-v2                          /v1/ocr
+  nemotron_parse                         -> nemotron-parse                           /v1/chat/completions
   vlm_embed                              -> llama-nemotron-embed-vl-1b-v2            /v1/embeddings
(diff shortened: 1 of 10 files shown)


girihemant19:
Addressed all three review findings:

- Parse configuration no longer breaks pdfium, pdfium_hybrid, or ocr requests.
- Helm now renders and auto-wires the Nemotron Parse endpoint/model in standalone and split topologies.
- Model-only Parse configuration is rejected during service configuration.

Validation:
- 125 focused regression tests passed
- Edge-case notebook: 9 PASS, 0 FAIL, 0 SKIP
- Hosted and self-hosted service-mode notebooks validated

jdye64 (approved):

charlesbluca (approved):

PR #2429, opened 2026-08-04
ci(greptile): add documentation-stays-current review rule

kheiss-uwzoo:
## Summary
- Add a single Greptile rule (`documentation-stays-current`) so PR review flags concrete stale, incorrect, contradictory, or missing docs when user-facing code, configuration, examples, or documentation change.
- Scope covers `nemo_retriever/src/**`, Helm, package/root READMEs, `docs/docs/extraction/**`, and `examples/**`.
- Compare documented APIs, CLI, config keys/defaults, formats, errors, and examples against checked-in source and tests; do not request docs for internal-only changes or speculate without a specific mismatch.

## Why this shape
Simplified per review feedback: keep the change focused on **Greptile PR review behavior** only (no repository-wide `AGENTS.md` / DORI coding-agent orchestration).

## Test plan
- [ ] Confirm Greptile picks up the new rule on this PR (or a follow-up docs/code PR that touches scoped paths)
- [ ] Spot-check that an internal-only change does not get a generic "please update docs" request
- [ ] Spot-check that a user-visible API/CLI/config drift can produce a concrete page-level flag


sosahi (approved):

KyleZheng1284 (commented):
I really like the idea of having a more in depth review for docs but I was wondering if we could simplify this PR to focus specifically on Greptile’s documentation review?

The main goal, as I understand it, is for an agent to catch documentation that becomes inaccurate or stale when user-facing code, configuration, examples, or documentation changes. 

I feel like instead of configuring this to be more agent side during development it might better serve if we just add this in depth review as a single structured rule in `.greptile/config.json`:

```json
{
  "id": "documentation-stays-current",
  "rule": "When this PR changes user-facing code, configuration, examples, or documentation, verify that the affected documentation remains accurate. Compare documented APIs, CLI commands and flags, configuration keys and defaults, supported formats, error behavior, and code examples against the PR's checked-in source and tests. Flag concrete stale, incorrect, contradictory, or missing documentation, and identify the affected page. Do not request documentation for internal-only changes or speculate without a specific mismatch.",
  "scope": [
    "nemo_retriever/src/**",
    "nemo_retriever/helm/**",
    "nemo_retriever/README.md",
    "nemo_retriever/docs/**",
    "docs/docs/extraction/**",
    "examples/**",
    "README.md"
  ],
  "severity": "medium"
}
```

Scoping the rule to both code and documentation would let Greptile:

- Review changed documentation against the implementation.
- Detect documentation made stale by user-visible code or configuration changes.
- Avoid generic documentation requests when there is no concrete mismatch.

With that approach, I think we could remove the new `AGENTS.md`, `docs/AGENTS.md`, and `docs/DORI_SETUP.md` files, along with the duplicate rule and context entries in `.greptile/rules.md` and `.greptile/files.json`.

My concern with the current approach is that a root AGENTS.md becomes repository-wide guidance for compatible coding agents. That can introduce documentation analysis, subagent coordination, and DORI-related decision-making during normal agent-assisted development, including tasks where documentation work may not be necessary.

Keeping this as a Greptile rule would provide the desired review-time protection without adding development-time overhead for contributors using coding agents.

kheiss-uwzoo:
Thanks @KyleZheng1284 — updated to match your suggestion.

This PR now only adds the single `documentation-stays-current` rule in `.greptile/config.json` (your wording + the shared code/docs/examples scope). Removed `AGENTS.md`, `docs/AGENTS.md`, `docs/DORI_SETUP.md`, and the duplicate Greptile `rules.md` / `files.json` entries.

.greptile/config.json
```diff
@@ -179,6 +179,20 @@
...
+    {
+      "id": "documentation-stays-current",
+      "rule": "When this PR changes user-facing code, configuration, examples, or documentation, verify that the affected documentation remains accurate. Compare documented APIs, CLI commands and flags, configuration keys and defaults, supported formats, error behavior, and code examples against the PR's checked-in source and tests. Flag concrete stale, incorrect, contradictory, or missing documentation, and identify the affected page. Do not request documentation for internal-only changes or speculate without a specific mismatch.",
+      "scope": [
+        "nemo_retriever/src/**",
+        "nemo_retriever/helm/**",
+        "nemo_retriever/README.md",
+        "nemo_retriever/docs/**",
+        "docs/docs/extraction/**",
+        "examples/**",
```
KyleZheng1284: One small scope suggestion: published documentation also exists under docs/docs/reference/**, including the collection-management API and CLI quickstart pages. Since the goal is to keep general user-facing documentation current, I think we should broaden it to if you agree:

- "docs/docs/extraction/**",
+ "docs/docs/**",

Otherwise, LGTM thanks Kurt

kheiss-uwzoo:
@KyleZheng1284 Agreed — broadened the rule scope from `docs/docs/extraction/**` to `docs/docs/**` so reference pages are covered too.

.greptile/config.json
kheiss-uwzoo: @KyleZheng1284 Agreed — broadened the rule scope from `docs/docs/extraction/**` to `docs/docs/**` so reference pages are covered too.

kheiss-uwzoo:
Following up now that this has merged, to correct the record on my end.

My original PR conflated two separate things, and that was a mistake in how I scoped it. The intent I pitched was the **composition-time** agentic docs workflow — root `AGENTS.md` / `docs/AGENTS.md` guidance that helps coding agents recognize user-visible changes and draft documentation to NVIDIA writing standards during development. I bundled that together with review-time enforcement, and @KyleZheng1284 was right that repository-wide `AGENTS.md` guidance does not belong in a Greptile config change — it affects every contributor's agent session and needs its own review.

What merged here is the **review-time** half, and it stands on its own: Greptile flags concrete documentation drift on PRs.

The composition-time half is now open as a separate draft PR — #2445 — where the original intention belongs and can get a proper review. The two layers are complementary: this rule catches drift after the fact; `AGENTS.md` prevents it at draft time.

Thanks for the catch, Kyle. Cleaner split.

PR #2439, opened 2026-08-06
Batch vLLM reranker scoring across queries

jioffe502:
## Summary
- submit aligned query/document pairs to one vLLM `LLM.score()` call
- let vLLM memory-schedule candidates across queries while preserving output and multimodal alignment
- disable per-call progress output for the batched path

## Why
The vLLM reranker grouped candidates by query and made one synchronous engine call per query. vLLM supports memory-aware N-to-N scoring and recommends submitting all inputs together, so those round trips prevented cross-query scheduling.

## Validation
- `135 passed`: VL reranker, Retriever batch-query, and actor/operator tests
- `23 passed`: final focused VL reranker suite
- Ruff and `git diff --check` pass
- Full ViDoRe v3 Energy A/B, 1,848 queries on one H100 80GB: 2.792 to 2.934 QPS (+5.08%) with NRB query batch size 4
- nDCG@10: 0.651244 to 0.651541; Recall@10: 0.716534 to 0.717165

BF16 scores near rank boundaries are batch-sensitive: 51/1,848 Energy queries changed top-10 membership at batch size 4. Aggregate quality moved slightly upward. The paired NRB change keeps the generic default at one and opts only the offline nightly into the measured batch-size-four throughput/latency knee.

### DIFF
--- a/nemo_retriever/src/nemo_retriever/models/local/nemotron_rerank_vl_v2.py
+++ b/nemo_retriever/src/nemo_retriever/models/local/nemotron_rerank_vl_v2.py
@@ -297,27 +297,20 @@ class NemotronRerankVLV2VLLM(BaseModel):
         if not pairs:
             return []
 
-        # Group consecutive pairs by query so each (query, [docs...]) goes to
-        # vLLM in a single batched score() call instead of one round-trip per pair.
-        scores: List[float] = [0.0] * len(pairs)
-        i = 0
-        while i < len(pairs):
-            q = pairs[i][0]
-            j = i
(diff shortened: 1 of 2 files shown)


jioffe502:
Closing. Measured alone against its upstream baseline, cross-query scoring was throughput-neutral (end-to-end query QPS +0.17% at query-batch 1, -0.33% at query-batch 4), and there is no offline batching consumer that needs the interface yet. The separately qualified tokenizer fast path (11-17% QPS with identical rankings) will land as a standalone change if reranker work is picked back up. Branch is preserved for reference.

PR #2444, opened 2026-08-07
add rerank to mcp

jperez999:
## Checklist


nemo_retriever/src/nemo_retriever/service/routers/ingest.py
```diff
@@ -1802,80 +1914,179 @@ async def answer(req: ServiceAnswerRequest, request: Request) -> Response | Answ
...
     agentic = False
+    rerank_request: QueryRequest | None = None
+    query_body = body
     try:
-        parsed = json.loads(body.decode("utf-8") if isinstance(body, (bytes, bytearray)) else body)
+        parsed = json.loads(
+            body.decode("utf-8") if isinstance(body, (bytes, bytearray)) else body
+        )
         agentic = bool(parsed.get("agentic")) if isinstance(parsed, dict) else False
+        if isinstance(parsed, dict) and parsed.get("rerank"):
```
mahikaw: nit: we should gate on `rerank_request.rerank` here

nemo_retriever/src/nemo_retriever/service/mcp_server.py
```diff
@@ -542,16 +577,25 @@ async def get_document(job_id: str, document_id: str) -> dict[str, Any]:
...
         ) -> dict[str, Any]:
-            return await service.query(query, top_k=top_k, format=format, payload=payload)
+            return await service.query(
+                query,
+                top_k=top_k,
+                format=format,
+                rerank=rerank,
+                rerank_top_k=rerank_top_k,
+                payload=payload,
+            )
```
mahikaw, line 576: we are passing rerank here and in payloads too above: . they can be passed conflicting values leading to unexpected behaviour.

nemo_retriever/src/nemo_retriever/service/routers/ingest.py
```diff
@@ -1802,80 +1914,179 @@ async def answer(req: ServiceAnswerRequest, request: Request) -> Response | Answ
...
+        use_remote_reranker = bool(
+            (config.nim_endpoints.rerank_invoke_url or "").strip()
         )
+        rerank_lock: asyncio.Lock | None = None
+        local_reranker = None
+        if not use_remote_reranker:
+            # The local model is service-owned and lazily loaded once. Serialize
+            # scoring too: HF/vLLM model instances are not safe to invoke from
+            # simultaneous request threads, and this avoids duplicate GPU work.
+            rerank_lock = getattr(request.app.state, "local_reranker_lock", None)
```
mahikaw, line 1926: the lock should be _created_ before this point. concurrent requests would create separate locks otherwise

jperez999: I don’t believe this is a race in the current async path: there is no await between reading and assigning app.state.local_reranker_lock, so another request cannot interleave before the shared lock is stored. Initialization and scoring then both occur under that lock. Separate service processes necessarily own separate local model instances and locks. I’d prefer to retain lazy initialization so reranker-disabled deployments do not allocate unnecessary state.

mahikaw (approved):

PR #2454, opened 2026-08-11
fix: parallelize known file-backed text ingestion

### LINKED ISSUE #2476: Bound LanceDB ingestion memory with an Arrow-native transactional sink
## Summary

Batch ingestion currently treats `IngestVdbOperator` as a global-batch operator. Ray repartitions the complete embedded dataset into one block, converts that block for `VDB.run()`, and only then begins the LanceDB table-write lifecycle.

That handoff is not bounded by rows or bytes, so peak host memory grows with the complete embedded result. This issue is about making the terminal storage path bounded and observable. It is **not** a claim that an unmodified full-corpus run has been observed OOMing on a 251 GiB host.

## Current evidence

| Control | Production code | Result |
| --- | --- | --- |
| Full BrowseComp-Plus validation | Validation stack containing #2454 and #2474 | 100,195 files produced 781,927 embedded rows and one 33.2 GiB Ray object-store block. The store actor reached 173.34 GiB RSS; host-available memory fell to 24.16 GiB. The run was manually stopped after about 930 seconds with zero rows committed. No OOM exception or kernel kill occurred. |
| BrowseComp-Plus 10k sink replay | Detached upstream `main` at `c80f4a51` | Replaying 75,065 already-embedded rows through `LanceDB.run()` under an 8 GiB/no-swap cgroup reached the sink and was OOM-killed (`OOMKilled=true`, exit 137). |
| Same BrowseComp-Plus replay | Detached upstream `main` at `c80f4a51` | The identical 75,065 rows completed under a 24 GiB/no-swap cgroup with a 12.378 GiB observed peak. |
| BRIGHT Biology 10k replay | Detached upstream `main` at `c80f4a51` | 10,014 already-embedded rows completed under 8 GiB at roughly 1.5 GiB process RSS. |

The controlled replay disables vector-index construction and does not repeat extraction, embedding, or Ray repartition. It isolates table creation/write and demonstrates a real memory-capacity failure at that boundary. It does **not** prove that full BrowseComp-Plus naturally OOMs on a 251 GiB host, nor does it reproduce a separate failure during vector/FTS index construction.

Related history should remain separate:

- The original BRIGHT nightly failure was serialized CPU text extraction and six-hour harness timeouts, not OOM.
- PR #2474 fixes Arrow/pandas representation correctness; it does not implement storage batching or memory admission. It allowed the full embedded result to reach the existing sink, where this downstream behavior became measurable.

## Desired outcome

Replace the global materialization boundary with a terminal LanceDB sink whose in-flight memory is bounded by configuration and whose lifecycle is explicit:

1. **Prepare**
   - Resolve the canonical Arrow schema and vector dimension before writing.
   - Open or initialize the target once.
   - Select overwrite or append semantics before consuming any batch.
2. **Consume with backpressure**
   - Project only canonical VDB columns plus compact accounting/result columns.
   - Consume bounded Arrow batches using a byte-based in-flight budget and small prefetch count.
   - Use one logical writer/coordinator; do not turn every Ray block into an independent LanceDB commit.
3. **Commit and finalize**
   - Validate the committed table version and row count.
   - Build or update vector/FTS indices once after the data write succeeds.
   - Apply compaction/optimization only through an explicit policy informed by table statistics.
4. **Abort and retry**
   - Do not report success or build indices after a failed data write.
   - Define overwrite and append retry semantics explicitly. Append must not silently duplicate rows.

Evaluate the public LanceDB paths before selecting a low-level implementation:

1. one `table.add()` supplied by a bounded Arrow `RecordBatch` iterator;
2. bounded Arrow/Parquet staging exposed as a `pyarrow.dataset.Dataset`, followed by one coordinated `table.add()`;
3. Lance's two-phase distributed-write model only if the public iterator/dataset paths cannot satisfy the gates.

## Prior art

PR #2041 showed that streaming storage can reduce memory and wall time, but its per-Ray-block `table.add()` design should remain an experimental comparison rather than the default:

- every block can become a table version/commit and create small fragments;
- correctness depends on mutable “first batch overwrites, later batches append” state;
- retries can duplicate or partially publish work;
- increasing writer concurrency creates competing committers rather than backpressure;
- result materialization and public result types became coupled to the storage change.

Reference: https://github.com/NVIDIA/NeMo-Retriever/pull/2041

## Required ablations

Use immutable inputs and the same hardware:

| Control | Purpose |
| --- | --- |
| Current global-batch sink | Establish peak memory, versions/fragments, and end-to-end baseline |
| PR #2041-style per-block append, experiment only | Quantify commit/version/fragment amplification |
| Bounded `RecordBatch` iterator | Evaluate the preferred public-API path |
| Staged `pyarrow.dataset.Dataset` | Evaluate the bounded-memory throughput fallback |

Tune by bytes, not only rows. Test at least 64 MiB, 256 MiB, and 512 MiB in-flight budgets with prefetch 1 and 2 on the fixed 10k BrowseComp-Plus control. Use one representative heterogeneous PDF corpus to validate the Arrow schema path. Do not repeat the full BrowseComp-Plus run until a bounded candidate wins these controls.

### Correctness

- Exact file and stored-row parity on the fixed BrowseComp-Plus 10k control: 10,000 files / 75,065 rows.
- Exact expected stored rows on a representative heterogeneous PDF control.
- Full BrowseComp-Plus eventually commits all 781,927 expected rows.
- Vector and FTS index coverage includes every committed row before the job reports success.
- Failure injection leaves no visible partial overwrite, builds no index, and retry behavior is explicit and tested.
- Schema validation and bad-vector handling match the existing production contract.
- No changes to the deprecated NRL benchmark harness.

### Resource bounds

- No global `repartition(num_blocks=1)` for the VDB sink.
- Peak sink RSS is bounded by the configured in-flight byte budget plus measured Arrow/LanceDB overhead, rather than corpus size.
- Ray object-store spill remains zero on the bounded controls.
- Table version and fragment growth do not scale linearly with the number of Ray input blocks.
- Peak RSS, object-store bytes/spill, versions, fragments, committed rows, and index duration are recorded.

### Performance

- The 10k BrowseComp-Plus control remains within 10% of the current 854.510-second end-to-end result while materially reducing sink memory.
- The winning design completes the full BrowseComp-Plus sink without approaching the host safety boundary.
- Record rows/s, bytes/s, write duration, and index duration are first-class telemetry.

### Tests

- Arrow batch iteration preserves schema and enforces the configured byte/prefetch bound.
- A real local LanceDB multi-batch write produces one complete table and correct indices.
- Failure/retry coverage proves visible-version and duplicate-row behavior.
- One bounded Ray integration test verifies backpressure and avoids global repartition.
- Existing ingestion and retrieval suites remain green.

## Non-goals

- Diagnosing unrelated reports that fail specifically during vector-index construction; first capture their exact phase and memory evidence.
- Redesigning extraction routing or actor admission.
- Changing service-mode model residency.
- Adding behavior to the deprecated NRL benchmark harness.
- Solving legacy full-result client materialization in the same production diff.
- Selecting the low-level distributed-fragment API before the public LanceDB paths are measured.

## References

- LanceDB large-ingest and iterator guidance: https://docs.lancedb.com/tables/create
- LanceDB concurrent-write limits: https://docs.lancedb.com/faq/faq-oss
- LanceDB `Table.add`, compaction, and optimization API: https://lancedb.github.io/lancedb/python/python/
- Lance distributed writes: https://lancedb.github.io/lance/distributed_write.html
- Prior experiment: https://github.com/NVIDIA/NeMo-Retriever/pull/2041

jioffe502:
## Purpose

This fixes library ingestion of large corpora that are already known to be text, especially file-backed `.txt` datasets such as BRIGHT and BrowseComp-Plus.

Input classification already identifies those files as text before graph construction. However, the resolved branch still used the generic `MultiTypeExtractOperator`, whose Ray actor pool defaulted to one worker. That serialized CPU text extraction and underfed downstream embedding.

```text
Before: ReadBinary -> MultiTypeExtractCPUActor(concurrency=1, dispatches to text splitting)
After:  ReadBinary -> TxtSplitCPUActor(concurrency<=8)
```

Routing is the implementation mechanism; scalable known-text extraction is the user-visible fix. This is not a replacement for the general multi-type router.

## What changes

- Builds a direct `TxtSplitActor` graph when the resolved extraction mode is `text`.
- Sizes the automatic batch text pool to at most eight CPU workers.
- Leaves `MultiTypeExtractOperator` in place for `auto` and other composite/unknown input paths.
- Reserves CPU for file readers and shared schema normalization before admitting actor pools.
- Uses one internal allocation implementation for both single-graph and shared preflight, so sizing and compatibility rules cannot diverge between those paths.
- Reduces only automatically derived pools; explicit `BatchTuningParams` and `node_overrides` remain authoritative and infeasible explicit plans fail before execution.
- Admits explicit Ray `(min, max)` and `(min, max, initial)` pools at their minimum/initial size without changing their upper bound.
- Keeps automatic sizing enabled when a node override changes only `batch_size` or another non-concurrency setting.

The admission changes are part of the fix: increasing text concurrency without reserving reader and normalization capacity can recreate a zero-progress Ray plan on constrained clusters.

## Scope boundaries

- Direct graph construction applies to known text inputs in library batch and in-process modes.
- Automatic pool sizing and shared resource admission apply only to library batch mode.
- `.texts([...])` remains inline text; this PR does not reinterpret file contents as inline inputs or change the `inline://` contract.
- PDF, image, audio, video, HTML, and mixed/unknown routing are unchanged.
- Service-mode actor residency is unchanged.
- Embedding and LanceDB behavior are unchanged; large final writes are tracked separately in #2476.
- The deprecated NRL benchmark harness is unchanged. Workload construction and telemetry remain NRB-owned.

## Evidence

| Check | Result |
| --- | --- |
| BrowseComp-Plus 10k direct-text run | 10,000/10,000 paths produced exactly 75,065 rows with 0 errors; Ray extraction completed in 33.18 s and the public call in 76.42 s including worker-environment setup |
| BrowseComp worker ablation | Increasing the same extraction workload from 1 to 8 workers reduced runtime from approximately 377 s to 66 s while preserving all 75,065 rows |
| Full BRIGHT acceptance | All 12 domains completed on the validation stack |
| Mixed text/PDF control | 268 files completed in 80.914 s and stored 3,145 valid rows |
| Classic-nightly non-regression | All 12 datasets passed: 1,859 files, 142,967 pages, and 202,646 rows; matched full BO767 was 1,850.989 s on the exact base versus 1,856.410 s on this PR (+0.29%, parity) |
| Current-main BRIGHT smoke check after rebase | Physical plan was `ReadBinary -> TxtSplitCPUActor`; 100/100 real files produced 100 rows; Ray dataset completed in 4.46 s after cold startup |

The direct path removes the serial CPU extraction bottleneck. Embedding and the final store remain independent downstream limits for full BrowseComp ingestion.

## Latest focused validation (September 15)

- Graph, planner, manifest, ingest interface, inline-text, and text-splitter suites: **289 passed, 3 skipped, 6 deselected** (integration cases run separately).
- Real Ray, 4 CPUs / 0 GPUs: **4 integration cases passed**, covering file/inline parity and combined inputs with default or batch-size-only text overrides.
- Separate bounded public-API checks: both `(1, 8)` and `(1, 8, 2)` explicit text pools completed on the same 4-CPU resource limit. Planner tests cover both admission paths and reject fixed/minimum/initial counts that exclude reader capacity.
- Consolidation parity tests cover constrained CPU/GPU budgets and preserve the distinction between single-executor and shared-preflight bookkeeping.
- Pre-commit on all nine changed files: **all applicable hooks passed**.
- Published performance guidance updated; MkDocs build not run because MkDocs is unavailable in this environment.
- Runtime acceptance evidence and NRB harness changes: [NRB MR !13](https://gitlab-master.nvidia.com/charlesb/nemo-retriever-benchmark/-/merge_requests/13)

## Reviewer focus

- Does known text bypass only redundant multi-type dispatch while `auto` remains unchanged?
- Does the direct graph preserve `TextChunkParams` and `split_config["text"]` precedence?
- Does batch admission reserve reader/normalization work without overriding explicit actor pools?


nemo_retriever/src/nemo_retriever/ingestor/branch_extraction.py
```diff
@@ -151,7 +152,12 @@ def _execute_batch(self) -> Any:
             source_cpu_reservation=0,
         )
         if hasattr(cluster_resources, "available_cpu_count"):
-            preflight_executors([*branch_executors, post_executor], cluster_resources)
+            normalization_cpus = float(len(branch_inputs)) if len(branch_inputs) > 1 else 0.0
```
jperez999: [P1] A batch with one file-backed text input and one `.texts()` input creates two branch datasets. On a 4-CPU cluster, this reserves 1 CPU for reading plus 2 CPUs here, while both automatic `TxtSplitActor` pools have a mandatory minimum of 1 CPU each: 1 + 2 + 2 = 5 CPUs. Preflight therefore rejects an otherwise default, supported ingestion before any work starts. Reserve only concurrently required normalization capacity, or combine identical extraction inputs before planning; add a real constrained-resource regression test.

nemo_retriever/src/nemo_retriever/graph/executor.py
```diff
@@ -191,33 +191,49 @@ def _requires_stable_pandas_blocks(nodes: list[Node]) -> bool:
...
-        minimum, _maximum, initial = (int(value) for value in concurrency)
         return (minimum, max(minimum, initial, planned), initial)
+    if isinstance(concurrency, tuple) and len(concurrency) == 2:
+        return (min(minimum, planned), planned)
     return planned
 
 
-def preflight_executors(executors: list[Any], cluster_resources: ClusterResources) -> None:
-    """Plan all lazy executor pools against one shared Ray resource snapshot."""
+def preflight_executors(
```
jperez999, line 225: Non-blocking design note: `_preflight_resources()` now duplicates the same entry collection, fixed/automatic accounting, fair allocation, mutation, and logging implemented by `preflight_executors()`. This change expands both copies; consider extracting a shared private planner in a follow-up to prevent the two paths from drifting. Keeping that refactor separate would preserve this PR's focused scope.

jperez999 (changes requested):

nemo_retriever/src/nemo_retriever/ingestor/branch_extraction.py
jioffe502: Fixed in 5418ee1c. Preflight now reserves one shared normalization CPU when multiple datasets are combined. The regression test uses the real shared planner with one file reader, one inline dataset, and 4 available CPUs; both automatic pools are admitted at their minimum instead of failing at 5 requested CPUs.

nemo_retriever/src/nemo_retriever/graph/executor.py
jioffe502: Agreed. I kept the shared-planner extraction out of this PR so this review fix remains focused. The duplicated planner implementation remains follow-up cleanup.

nemo_retriever/src/nemo_retriever/graph/executor.py
```diff
@@ -191,33 +191,49 @@ def _requires_stable_pandas_blocks(nodes: list[Node]) -> bool:
...
+        return minimum, maximum, initial
+    return 1, int(concurrency), 1
+
+
 def _concurrency_target(concurrency: Any) -> int:
     """Return the largest actor-pool size that resource planning can permit."""
-    if isinstance(concurrency, tuple):
-        return int(concurrency[1] if len(concurrency) == 3 else concurrency[0])
-    return int(concurrency)
+    return _concurrency_bounds(concurrency)[1]
```
jperez999, line 207: [P2] Preserve elastic actor-pool upper bounds

Ray's `(min, max)` and `(min, max, initial)` tuples describe autoscaling ranges, not a request to reserve `max` actors immediately. Returning the upper bound here makes every explicit tuple a fixed-capacity admission requirement: for example, `(1, 8)` with `num_cpus=1` now raises on a healthy 4-CPU cluster even though Ray can run the pool at 1-4 actors. This changes all direct `node_overrides` callers, not just known-text ingestion, and appears unnecessary for the focused fix. Please keep feasibility based on the startup/minimum requirement and only reduce automatically derived pools, or separate and justify this global compatibility change independently.

jperez999 (changes requested):

nemo_retriever/src/nemo_retriever/graph/executor.py
jioffe502: Fixed in 2c9d96c0. Both preflight paths now charge explicit elastic pools at their minimum/initial size and leave the upper bound unchanged; fixed integer counts still count in full. Also corrected partial overrides so batch_size alone does not disable automatic sizing. Added shared/single planner regressions and real-Ray four-CPU file/inline tests. Validation: 286 focused tests passed, 4 integration cases passed, and separate public ingestion checks completed with both (1, 8) and (1, 8, 2) on four CPUs. All applicable pre-commit hooks passed. Updated the PR description and performance guidance to match.

jperez999 (approved):

PR #2484, opened 2026-08-13
fix(store): store batch image assets and return valid rows

mahikaw:
## Description
 Fixes the two open items on the bug: the ArrowInvalid: Invalid null value crash at the _stored_image_uri assignment, and the malformed result frame that could not be consumed with standard pandas iteration.

**Root cause:** Two distinct issues. _stored_image_uri is created upstream in the UDF stage as None for every row, so Arrow types it null and rejects a string scalar. Separately, _store_nested_image_payloads only recursed into list, so NumPy-backed element cells were returned unstored, and writing a NumPy array back via DataFrame.at collapses it to a 0-d array, which is what breaks Arrow conversion of the returned batch.

**Fix:** Include _stored_image_uri in the object-dtype normalization, and convert one-dimensional NumPy element cells to lists before persisting nested payloads.

**Testing:** Two new cases in test_store_pipeline_stages.py for the null-typed Arrow column and for NumPy-backed element images, both asserting the result survives pa.Table.from_pandas. Verified end-to-end through a Ray Data UDFOperator(explode) → StoreOperator plan: 3 images written, URIs returned, base64 stripped, rows readable via iterrows; on main the same run writes 2 images and returns 0-d array cells. Full suite matches the main baseline.

## Checklist

### DIFF
--- a/nemo_retriever/tests/test_store_pipeline_stages.py
+++ b/nemo_retriever/tests/test_store_pipeline_stages.py
@@ -12,6 +12,7 @@ import io
 from pathlib import Path
 from urllib.parse import urlparse
 
+import numpy as np
 import pandas as pd
 import pytest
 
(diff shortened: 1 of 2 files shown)


nemo_retriever/tests/test_store_pipeline_stages.py
```diff
@@ -166,6 +167,55 @@ def test_store_operator_writes_top_level_image_b64_video_frame(self, tmp_path: P
...
+        assert isinstance(df["_stored_image_uri"].dtype, pd.ArrowDtype)
+
+        result = StoreOperator(params=StoreParams(storage_uri=str(tmp_path))).process(df)
+
+        stored_uri = result.iloc[0]["_stored_image_uri"]
+        assert isinstance(stored_uri, str)
+        assert stored_uri.startswith("file://")
+        assert Path(urlparse(stored_uri).path).read_bytes() == base64.b64decode(b64)
+        assert result.iloc[0]["page_image"]["stored_image_uri"] == stored_uri
+        pa.Table.from_pandas(result)
```
jioffe502, line 193:
```suggestion
        list(result.iterrows())
        pa.Table.from_pandas(result)
```

jioffe502 (commented):
This is the right StoreOperator fix. Adding `_stored_image_uri` to the existing object-column helper is the crash QA hit on RC2 (writing a path string into a column that was still typed as empty/null).

One small test ask below — the rest looks good to me.

nemo_retriever/tests/test_store_pipeline_stages.py
```diff
@@ -166,6 +167,55 @@ def test_store_operator_writes_top_level_image_b64_video_frame(self, tmp_path: P
...
+        df = arrow_table.to_pandas()
+        assert isinstance(df.iloc[0]["images"], np.ndarray)
+
+        result = StoreOperator(params=StoreParams(storage_uri=str(tmp_path))).process(df)
+
+        element = result.iloc[0]["images"][0]
+        assert element["image_b64"] is None
+        assert Path(urlparse(element["stored_image_uri"]).path).read_bytes() == base64.b64decode(element_b64)
+        # Ray Data re-serializes each returned batch through Arrow.
+        pa.Table.from_pandas(result)
```
jioffe502, line 218:
```suggestion
        list(result.iterrows())
        pa.Table.from_pandas(result)
```

jioffe502 (approved):

PR #2514, opened 2026-08-17
fix(helm): propagate global NGC secret names to NIM CRs

jdye64:
## Summary
- Fix Helm so `ngcImagePullSecret.name` and `ngcApiSecret.name` propagate into every rendered `NIMCache` / `NIMService` (not just Retriever Pods).
- Per-NIM `image.pullSecrets` / `authSecret` now default to empty and inherit the chart-wide names; non-empty per-NIM values still override for that NIM only.
- Document the inheritance contract and add regression coverage for the RC4 reproduction.

## Problem
Overriding the global Secret names updated direct Pods, but NIM Operator CRs kept hard-coded `ngc-secret` / `ngc-api`. Model-download Jobs and NIM Pods then referenced missing Secrets while the Retriever deployment looked healthy.

## Test plan
- [x] `helm template` with `ngcImagePullSecret.name=rc4-global-pull` and `ngcApiSecret.name=rc4-global-api` shows those names on every NIMCache/NIMService
- [x] `uv run --with pyyaml python -m unittest discover -s nemo_retriever/tests -p 'test_helm_nim_global_secret_names.py' -v`
- [x] Related suites: `test_helm_answer_llm_generation.py`, `test_helm_nimcache_model_profile.py`, `test_helm_nimservice_resources.py`
- [x] `mkdocs build --strict` from `docs/` (api-keys.md note)
- [ ] Reviewer: confirm per-NIM override still wins when both global and per-NIM names are set


nemo_retriever/helm/openshift.md
```diff
@@ -103,6 +103,8 @@ imagePullSecrets:
   - name: default-dockercfg-xxxxx   # replace with your SA secret (section below)
 ```
 
+When `ngcImagePullSecret.name` is empty and NIM Operator CRs are enabled, set a non-empty `nimOperator.<key>.image.pullSecrets` for each enabled NIM (or keep the chart-wide name and use the "Both NGC NIMs..." pattern below). An empty global name with empty per-NIM `pullSecrets` fails chart render.
```
charlesbluca: `git diff --check upstream/main...HEAD` still fails here because these newly added lines use CRLF endings, which Git treats as trailing whitespace. Please normalize this addition (or the file) to LF.

jperez999 (approved):

nemo_retriever/helm/openshift.md
jdye64: Fixed in 5e02944d — the two added lines now use LF, so `git diff --check upstream/main...HEAD` is clean.

I normalized only the addition rather than the whole file: `openshift.md` is entirely CRLF on `main`, and #2531 already converts the full file to LF. Normalizing it here too would duplicate that 261-line change and conflict between the two PRs, so this keeps the diff at the intended two lines.

PR #2551, opened 2026-08-18
fix(embed): fail the run when embed rows arrive with no embedding

rhdong:
## Problem
Rows whose embedding failed were silently excluded from the index, and the run still reported success.
- When several embed workers share one GPU, vLLM can refuse to start an engine. A bare `except` caught the refusal and turned the whole batch into `{"embedding": []}`.
- Those rows never reached the index, yet the run reported `success: true` with `exit_code 0`.
- Measured on one 44 GiB card with the shipped `embed_workers: 3`: the index held 41,181 of 79,234 rows, and bo767 nDCG@10 was 0.2598 against an expected ~0.75. Every log line looked healthy.

## Root cause
The engine is refused permission to start. This is a capacity check, not an allocation failure.
- Only library mode is affected, because only library mode starts an embed engine inside the ingest process. It runs the model in-process on vLLM, so the engine competes for one GPU with the extraction stages of the same run. Pointing `embedding_endpoint` at a separate embedding service starts no in-process engine and makes the failure disappear.
- vLLM admits an engine only if free memory exceeds `gpu_memory_utilization x total`. That threshold scales with card size; the memory the run already holds when engines start does not.
- This model is encoder-only and allocates no KV cache, so an engine's real footprint is ~8 GiB against a check demanding ~20 GiB.
- The two stages do not fit together. Measured on one 44.39 GiB card, one run, sampled every 10 s:

| Resident on the card | At peak (2 engines admitted) | After extraction ends (3 admitted) |
|---|---|---|
| Extraction actors (OCR, page-element) | ~28.0 GiB | none |
| vLLM embed engines | ~16.3 GiB (2 x ~8.2) | 24.9 GiB (3 x ~8.3) |
| **Total** | **44.3 GiB (98.6% of card)** | **24.9 GiB (55.4%)** |

- Free memory at the refused startups ranged 0.11-19.08 GiB against a demand that stayed constant at 19.98 GiB. Once extraction ends the pressure lifts and the refused engines start, so the damage is bounded by how much of the run elapsed first.
- The refusal is usually `ValueError: Free memory ... is less than desired GPU memory utilization`. When free memory collapses to ~0.16 GiB, `cudaMalloc` fails before that check and vLLM reports `CUDA error: out of memory` instead. Both end in `RuntimeError: Engine core initialization failed`, which is the string the classifier matches, so both are covered.
- The same config on an 80 GB card loses nothing.

## Fix
Five guards, each covering a failure shape the others cannot see.

| Failure shape | Guard |
|---|---|
| Engine raises: admission refused, or engine death | `models/inference/runtime.py` re-raises instead of returning empty embeddings |
| vLLM returns `[]` for every row, raising nothing | `models/inference/main_text_embed.py` treats an empty batch result as fatal |
| Multimodal batch answered short | `main_text_embed._multimodal_callable_runner` compares vectors received against rows submitted with an image, instead of padding the shortfall with `None` |
| Some rows fail, zero-padded to the right length | `_finalize_vectors` stops discarding its own count of failed rows |
| Anything the above miss | `common/vdb/lancedb.py` refuses to build an index from rows with an empty embedding |

The writer guard carries the correctness guarantee for the shape it owns: it keys on the row, not the cause, so an empty embedding is refused whatever produced it - any backend, any GPU, endpoint path included. It is not a guarantee over every way a row can lose its vector. A missing embedding delivered as `None` stays a counted, non-fatal drop, because `operators/embed/text_embed.py:107` writes `None` on purpose for a blank-text row it chose not to embed; making `None` fatal would fail ingests that work today. The runtime classifier buys no correctness at all and only shortens time-to-failure, because the vector-store write is a single terminal global-batch stage and no earlier write exists to fail at. Do not delete either one as redundant.

`on_bad_vectors` is untouched. It governs malformed vectors, while an empty embedding means the embed stage produced nothing: a separate category, counted as `empty_embedding`.

## Verification
| Scenario | Result |
|---|---|
| Runtime classifier | `exit_code 10` in ~107 s, no index directory, 0 rows reached the writer |
| Writer guard alone | `exit_code 10` after ~1 h 48 m, `lancedb/` left empty: no table, no `.lance` files |
| 80 GB card, same config | completes normally, all three engines admitted, nothing lost |

The full suite passes on Linux with the same failure set as `main`, and all `pre-commit` hooks pass.

## Residual risk
- Bare `OutOfMemoryError` is deliberately not fatal, because on the HuggingFace local backend an OOM is per-batch and recoverable. That backend was never exercised on hardware, so the protection rests on code reasoning and unit tests.
- Two `BaseException` handlers convert the new error to `{"embedding": None}`. Neither is on the shipped route, but one exports an identically named function, so future rewiring could lose the protection. Tests pin the gap.
- Classification matches exception types and message fragments. If upstream rewords a message, the run fails later at the writer rather than silently succeeding.
- The multimodal guard has not run on hardware. It was added after the last GPU leg, in response to review, and is covered by unit tests only.
- `None` is not covered. Four `BaseException` handlers in `operators/embed/text_embed.py` and `common/modality/pipeline/embedding.py` rewrite a failed batch to `{"embedding": None}`, which the writers deliberately ignore. Neither route has a shipped importer, and tests pin that; closing it properly means separating "not embedded on purpose" from "embedding lost" at the source, which is out of scope here.
- Most measurements are single runs. The two failing runs disagree on how much was lost, so no per-configuration accuracy delta is claimed.

## Known gaps, not addressed here
Out of scope, and the reason one missing check became a silent 48% loss.
- `drain_errors()` has no production callers. The failure path calls `report_error()`, so the exception is stored, but nothing in `nemo_retriever/src` reads it back.
- `has_embedding` is written but never read as a condition. It is set in six places, including correctly to `False` here, and no code branches on it.
- Metric gates count quantity, not quality. The failing run passed `files==767`, `pages==54730`, `query_count==991`, and nothing asked whether those pages produced usable vectors.

- The writer guard has not fired on hardware at this commit. In every end-to-end run the
  runtime classifier aborted first, and the one run that did reach the writer used an
  earlier revision of `lancedb.py`. That path is covered by unit tests only.

## Checklist


rhdong:
Covered by https://github.com/NVIDIA/NeMo-Retriever/pull/2599

PR #2567, opened 2026-08-20
feat(vdb): add optional streaming ingest with LanceDB

jioffe502:
## Why

`RayDataExecutor` historically sends VDB ingestion through one global Ray batch. That preserves backend compatibility, but it creates a corpus-sized handoff before LanceDB can write.

## What changed

- `VDB` has one optional, non-abstract `stream_ingest()` capability over a lazy iterable of canonical NRL record dictionaries.
- `IngestVdbOperator` owns canonical conversion. `RayDataExecutor.ingest()` owns Ray iteration, prefetch, cleanup, retained results, and downstream ordering; it has no LanceDB dependency.
- LanceDB owns bounded Arrow packing, one native table mutation, schema/index/optimization, validation, and separate process-shared mutation and finalization locks for embedded local stores.
- Default streaming stores no durable retry history. A caller-persisted `stream_operation_id` explicitly enables durable retry markers and reconstructed-backend recovery; ambiguous append outcomes fail closed.
- Backends without the capability retain the global-batch path. `PutVdbOperator`, service/in-process execution, and `build_dataset()` retain their historical behavior. Stream-only settings are rejected on legacy paths rather than ignored.
- Removed the experimental policy/report surface, redundant failure-test module, and research document.

## Validation

- Final source: `c86676e471151a93928aa3254f03a407ccdaadbe`; clean source snapshot: `138772b0e9165741c4a5c0d72f0f3dfbf41fa8030bfe65fc8059840a096cdb97`.
- Merge-resolution validation passed: 298 conflict-focused VDB/Lance/executor/graph tests; 71 real-Lance behavior, recovery, concurrency, index, and write-policy tests; 8 real Ray/service/Lance integration tests; and 77 ingest planning tests.
- Repository Black, Flake8, compile, and commit hooks passed on the merged source. Fresh GitHub CI is the authoritative final-head CI result.
- The final-source frozen 12-dataset replay passed 24/24 baseline/candidate arms: 197,796 stored rows per arm with exact input, content digest, schema, index coverage, version, and retained terminal-result audits. Baseline was `6cdbc1daca379fccbd94fb46ad6ef963ca4cacb1`; candidate was the final source above. Harnesses and artifacts remain outside NRL.
- Counterbalanced full-ingest JP20 control passed all four baseline/candidate runs, including result and semantic audits: 1,940 pages, 3,201 stored rows, and 115 queries per run. This performance comparison used baseline `6cdbc1daca379fccbd94fb46ad6ef963ca4cacb1` and candidate `964db8e83ff5c19fdaafb4d7629690bb6941656b` (tree `5168a398cccaf93ee9a7e1d714cf1b21fdbd632b`). Later changes are limited to local coordination, capability fallback, recovery error handling, upstream conflict reconciliation, and tests; their exact sink correctness is covered by the final-source replay above.

| Full-ingest JP20 measurement | Baseline mean | Streaming mean | Change |
| --- | ---: | ---: | ---: |
| Ingest time | 103.34 s | 87.66 s | -15.17% |
| Launcher wall time | 134.98 s | 120.09 s | -11.03% |
| Ray object-store peak | 1.84 GB | 1.15 GB | -37.19% |
| `/dev/shm` peak | 2.24 GB | 1.34 GB | -40.08% |
| Cgroup peak memory | 65.00 GB | 65.02 GB | +0.04% |

Baseline ingest runs were 99.976-106.695 seconds; streaming runs were 85.514-89.796 seconds. The 12-dataset replay is a sink correctness/resource check, not an end-to-end performance claim.

Relates to #2476 and #2570.


nemo_retriever/src/nemo_retriever/common/vdb/lancedb.py
```diff
@@ -1230,14 +1295,79 @@ def write_to_index(
...
+            unsupported = ", ".join(sorted(kwargs))
+            raise TypeError(f"Unsupported LanceDB stream_ingest setting(s): {unsupported}")
+
+        from nemo_retriever.common.vdb.sink import write_lancedb_records
+
+        with self._stream_lock:
+            generated_operation_id = self.stream_operation_id is None
+            if generated_operation_id and self._pending_stream_operation_id is None:
+                self._pending_stream_operation_id = str(uuid.uuid4())
+            operation_id = self.stream_operation_id or self._pending_stream_operation_id
```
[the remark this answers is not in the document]
jioffe502: This is a real limitation of an omitted caller identity, and the implementation now makes the contract explicit rather than pretending a generated ID is restart-safe. A generated ID is retained for retry after a caught failure on the same LanceDB object. For backend reconstruction or process restart, callers must set and durably retain stream_operation_id; the consolidated real-Lance tests cover that recovery path. Incomplete markers fail closed so a fresh identity cannot mutate or expose uncertain data. We also prototyped persisting generated identities, but once only a success marker remains there is no sound way to distinguish a retry from a new intentional identical append. The published docs and API docstring now state the same-instance versus caller-persisted restart boundary.

nemo_retriever/tests/test_ray_vdb_sink_integration.py
```diff
@@ -0,0 +1,235 @@
...
+def test_ray_streams_three_blocks_into_real_lancedb_and_preserves_contract(tmp_path, monkeypatch) -> None:
+    """Streaming ingest avoids the global sink barrier and executes in order once."""
+
+    assert ray.__version__ == "2.56.1"
+    monkeypatch.setenv("CUDA_VISIBLE_DEVICES", "")
+    monkeypatch.setenv("RAY_ACCEL_ENV_VAR_OVERRIDE_ON_ZERO", "0")
+
+    if ray.is_initialized():
+        ray.shutdown()
+    ray.init(num_cpus=4, num_gpus=0, include_dashboard=False, log_to_driver=False)
```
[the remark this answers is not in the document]
jioffe502: These tests are intentionally real local integrations, as required by this PR acceptance criterion. Ray is an in-process local cluster and LanceDB is an embedded filesystem library here; neither test contacts an external service. Mocks cannot prove Ray native multi-block iteration or LanceDB RecordBatchReader mutation/recovery behavior. External services remain mocked/absent, while the heavy Ray case is kept to one focused integration test and the Lance behavior is consolidated in one module.

nemo_retriever/pyproject.toml
```diff
@@ -87,6 +87,7 @@ dependencies = [
   "universal-pathlib>=0.2.0",
   "numpy>=1.26.0",
   "debugpy>=1.8.0",
+  "filelock",
```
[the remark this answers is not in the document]
jioffe502: Fixed in 68ef3f0b. The direct requirement is now filelock>=3.12.1, matching NRL’s Python >=3.12 support floor; filelock 3.12.1 is the upstream release that added Python 3.12 support. The lockfile change is limited to that requires-dist bound, and the uv 0.12.10 lock check plus repository pre-commit hook pass.

nemo_retriever/src/nemo_retriever/common/vdb/lancedb.py
```diff
@@ -1230,14 +1362,553 @@ def write_to_index(
...
+        ValueError
+            If records or configured streaming controls are invalid.
+        RuntimeError
+            If mutation, recovery, validation, indexing, or optimization fails.
+        """
+
+        if not self.supports_stream_ingest:
+            raise UnsupportedVDBOperation("LanceDB.stream_ingest() is unavailable; use the legacy global-batch path.")
+
+        with self._stream_lock, FileLock(_stream_lock_path(self.uri, self.table_name)):
```
jperez999: **[P1] Coordinate legacy writes with the same process-shared lock**

Only `stream_ingest()` acquires this `FileLock`; `run()` and `put()` take only the per-instance `_write_lock`. Those legacy paths are still public/documented, and `build_dataset()` intentionally continues to execute `run()`, so another instance or process can mutate the same local table during the streaming mutation/finalization window. With the default `stream_operation_id=None`, there is no pending tag for `_assert_legacy_table_ready()` to detect, so an append can commit against a moving base (or an overwrite can replace the stream's data) despite the cross-process serialization guarantee. Could the filesystem lock be factored into a shared table-mutation guard and used by `stream_ingest()`, `run()`, and `put()` for local fixed-table URIs? A mixed stream/legacy concurrency test would pin this down.

nemo_retriever/src/nemo_retriever/graph/executor.py
```diff
@@ -553,10 +561,81 @@ def _linearize(graph: Graph) -> List[Node]:
...
+            for index, node in enumerate(nodes)
+            if isinstance(node.operator, IngestVdbOperator) and node.operator._supports_stream_ingest()
+        ]
+        return positions[0] if len(positions) == 1 else None
+
     def ingest(self, data: Any, **kwargs: Any) -> Any:
         """Build, execute, and materialize a Ray Data pipeline from the graph."""
 
-        return ray_dataset_to_pandas(self.build_dataset(data, **kwargs))
+        if kwargs:
```
jperez999, line 579: **[P2] Keep the public executor behavior scoped to this feature**

Before this PR, `ingest(..., **kwargs)` forwarded to `build_dataset()`, and `build_dataset(..., **kwargs)` accepted and ignored extra kwargs. These guards make both public methods raise even for legacy/non-streaming graphs. That is an unrelated compatibility break and conflicts with the PR description's statement that `build_dataset()` retains its historical behavior. The private `_build_dataset(segment=...)` already prevents callers from selecting internal segments, so please remove these guards (or deprecate kwargs separately) to keep this PR minimal.

jperez999 (changes requested):

nemo_retriever/src/nemo_retriever/graph/executor.py
jperez999: Dont fix this here, but we will need to fix this in a follow up PR.

nemo_retriever/src/nemo_retriever/operators/vdb.py
```diff
@@ -155,6 +168,46 @@ def process(self, data: Any, **kwargs: Any) -> Any:
             self._vdb.run(records)
         return data
 
+    def _supports_stream_ingest(self) -> bool:
+        """Return whether the configured VDB opts into canonical record streaming."""
+
+        return self._vdb.supports_stream_ingest
```
jperez999: **[P2] Treat a missing optional capability as opt-out**

`IngestVdbOperator(vdb=...)` does not require the injected object to inherit the current `VDB` base class; historically, a contract-compatible object could work by implementing the legacy methods. This direct attribute access makes such an existing backend raise `AttributeError` during Ray capability discovery if it predates `supports_stream_ingest`, rather than taking the promised legacy `run()` fallback. Since streaming is explicitly optional, could this use `bool(getattr(self._vdb, "supports_stream_ingest", False))` so a missing capability is treated as opt-out?

nemo_retriever/src/nemo_retriever/common/vdb/lancedb.py
jioffe502: Fixed in ccf75387. Local stream, run, and put writes now use the same table-scoped process lock. stream_ingest holds it through mutation and finalization; run and put acquire it before their legacy mutations, while non-filesystem paths remain unchanged. Added a real spawned-process regression with stream_operation_id=None: a stream is paused before native add, a legacy run cannot enter, the holder is terminated, and the legacy writer then completes with exactly the intended rows. The existing real legacy concurrency/index tests also remain green.

nemo_retriever/src/nemo_retriever/operators/vdb.py
jioffe502: Fixed in ccf75387 with the suggested bool(getattr(..., False)) fallback. A compact contract test injects a legacy duck-typed backend with no supports_stream_ingest attribute and verifies that capability discovery produces the normal UnsupportedVDBOperation fallback rather than AttributeError.

nemo_retriever/src/nemo_retriever/graph/executor.py
jioffe502: Acknowledged. Per your follow-up, I left the executor kwargs behavior unchanged in this PR; that compatibility change is deferred to a follow-up.

nemo_retriever/src/nemo_retriever/graph/executor.py
```diff
@@ -553,10 +561,81 @@ def _linearize(graph: Graph) -> List[Node]:
...
+        finally:
+            close = getattr(batch_iterator, "close", None)
+            if callable(close):
+                close()
+
+        if has_downstream_nodes:
+            import ray.data as rd
+
+            if terminal_frames:
+                continuation_input = rd.from_pandas(terminal_frames)
```
jperez999, line 682: **[P1] Preserve pandas blocks before recreating the Dataset**

`rd.from_pandas()` performs tensor-extension casting immediately using the global `DataContext` (the default is enabled). Passing `input_preserves_pandas_output=True` only afterward changes the copied Dataset context and downstream `map_batches`; it is too late to stop this conversion. For heterogeneous ndarray/bbox/object columns, this can reintroduce the tensor/schema promotion that `_requires_stable_pandas_blocks()` is specifically avoiding, so a graph with nodes after the VDB sink can fail or observe different row types than the same unsplit graph. Please construct the continuation Dataset under the pandas-preserving context (ideally reuse the context-management path already used in `_build_dataset`) and add a downstream regression with heterogeneous object columns.

nemo_retriever/src/nemo_retriever/graph/executor.py
```diff
@@ -219,7 +222,12 @@ def preflight_executors(executors: list[Any], cluster_resources: ClusterResource
     available_cpus = cluster_resources.available_cpu_count()
     available_gpus = cluster_resources.available_gpu_count()
     for executor in executors:
-        for node in executor._linearize(resolve_graph(executor.graph, cluster_resources)):
+        nodes = executor._linearize(resolve_graph(executor.graph, cluster_resources))
+        sink_index = executor._stream_ingest_index(nodes)
+        planned_nodes = (
+            [node for index, node in enumerate(nodes) if index != sink_index] if sink_index is not None else nodes
```
jperez999: **[P2] Preflight the two streaming segments as mutually exclusive**

Removing only the sink still charges every actor pool before and after it at the same time. However, `ingest()` fully drains/closes the upstream Dataset before it constructs the downstream Dataset, so those pools are mutually exclusive. A graph with one 1-GPU stage on each side of the sink now runs sequentially on one GPU when invoked directly, but `preflight_executors()` sums them to 2 GPUs and rejects the same graph. Please plan each segment independently (then reserve the per-executor maximum that can be live), so shared preflight agrees with the segmented execution path.

nemo_retriever/src/nemo_retriever/common/vdb/lancedb.py
```diff
@@ -105,6 +182,33 @@ def _normalize_on_bad_vectors(value: str) -> str:
     return normalized
 
 
+def _stabilize_fill_vectors(
```
jperez999, line 186: **[P3] Keep the legacy fill-policy change out of this sink patch**

This helper is also called from the legacy `create_index`/`run` and `put` paths, so on newer LanceDB releases this PR changes their existing behavior from preserving valid vector elements to replacing the entire vector. That is a separate compatibility policy change, not required to make the streamed Arrow path bounded, and it is not called out in the PR description's legacy-path guarantees. To keep this patch minimal, please either limit normalization to the new stream path and handle the legacy compatibility decision separately, or explicitly add this legacy semantic change to the PR scope and reuse one shared bad-vector policy helper instead of maintaining `_stabilize_fill_vectors` alongside `_apply_deferred_bad_vector_policy`.

jperez999 (approved):

PR #2591, opened 2026-08-26
Make Nemotron 3.5 Lightning the default answer and judge model

jioffe502:
## Description

Make Nemotron 3.5 Lightning the default answer generator and LLM judge across runtime configuration, evaluation tools, and examples. Answer generation remains opt-in.

- Update Helm and Compose to NIM image `2.0.9-variant`, one GPU, automatic profile selection, and the `nemotron_v3` reasoning parser.
- Keep answer-model tool calling opt-in; update the existing judge tool parser to `qwen3_coder`.
- Update matching tests and documentation. Retain Nano as an optional example and preserve local Super-49B compatibility and historical benchmark attribution.

Production Python changes are model strings and docstrings only. No new runtime logic or abstractions.

## Validation

- 261 tests and 38 subtests passed; the example follow-up passed 17 targeted tests. Pre-commit hooks, Compose rendering, and documented Helm tool-call overrides passed.
- All CI checks passed on `175ab4df`, including Docker build/test, full unit tests, and platform installs.
- Self-hosted smoke passed on one RTX PRO 6000 Blackwell (96 GB), including NIM startup, generation, judging, reasoning controls, and opt-in tool calling/agentic retrieval.
- Hosted validation passed with a working NVIDIA API key: generation with reasoning enabled/disabled; correct-answer score 1.0 and incorrect-answer score 0.0; both judge prompts individually; the skill-eval config's 4096-token judge; a three-case retrieval/generation/judge pipeline; tool calling; and agentic selection of the correct document.
- The hosted pipeline completed all three cases with scores 1.0, 0.5, and 1.0 and no reported generation/judge errors. These synthetic checks establish functionality, not benchmark quality. Long, variable request durations and an observed NFS package-import stall make this run unsuitable for performance conclusions.
- Strict MkDocs has the same pre-existing embedded CLI link failure as `main`. Kubernetes reconciliation and other GPU families were not exercised.

## Review note

The GPU-variable finding compares an earlier PR revision rather than `main`. The existing Compose configuration and README use `NIM_ANSWER_GPU_ID_0`; this PR preserves it. Rendering with `NIM_ANSWER_GPU_ID_0=1` correctly selects GPU `1`.


nemo_retriever/src/nemo_retriever/service/config.py
```diff
@@ -247,7 +247,7 @@ class LLMConfig(RichModel):
     model_config = ConfigDict(extra="forbid")
 
     enabled: bool = False
-    model: str = "openai/nvidia/llama-3.3-nemotron-super-49b-v1.5"
+    model: str = "openai/nvidia/nemotron-3.5-lightning-30b-a3b"
```
jperez999, line 250: [P1] Increase the output budget with the reasoning-model default

This changes the default to a reasoning model but leaves the service default at `max_tokens=512` while `reasoning_enabled=True`. For Lightning, `max_tokens` covers reasoning plus visible output; the NVIDIA guide warns that a request can exhaust the budget before producing a final answer and uses 4,096 tokens with a 2,048-token thinking budget. Because `/v1/answer` returns only parsed content, a default request can therefore produce an empty or truncated answer even though the NIM call succeeds. Please make the coupled defaults safe. The smallest option is to raise the service budget, ideally aligning it with the existing 4,096-token `LiteLLMClient` default, or explicitly cap or disable thinking. Reference: https://docs.nvidia.com/nim/large-language-models/2.0.10/get-started/advanced/get-started-nemotron-3.5-lightning.html#control-thinking-budget.

jperez999 (changes requested):

jperez999 (approved):

PR #2599, opened 2026-08-27
fix(embedding): split overlength text before inference

jioffe502:
## Summary

One oversized page should not prevent nearby pages from becoming searchable or silently lose text at the end.

This PR splits overlength text before embedding, preserves every piece's document/page information, and rejects incomplete results before a vector-database write. Text that fits does not get split. This is protection for rare oversized inputs, not a throughput change.

## What we observed

Earlier dataset qualification, not rerun on the latest head:

- **JP20:** 0 of 3,201 text rows exceeded the 8,192-token formatted limit; maximum 4,555.
- **BO767:** 2 of 83,341 text rows exceeded it (0.0024%); maximum 10,647. Those parents became four children, adding two rows.
- **ViDoRe V3:** 19,252 embedding rows primarily use image-bearing VL input, outside the text-splitting policy.
- JP20 retrieval metrics and attribution were unchanged. Full BO767 showed no material retrieval regression (nDCG@10 delta -0.001064; Recall@5 and Recall@10 delta +0.001009).

## Core behavior

- Count the exact pinned tokenizer's formatted input, including prefix and special tokens, against the smaller configured/checkpoint limit.
- Split only overlength text into deterministic, contiguous ranges that reconstruct the original text and token IDs exactly. Unicode boundaries are reversible.
- Keep neighboring rows, source/page fields, and stable child IDs/order through embedding, storage, and retrieval. Split provenance persists in JSON metadata; collection storage keys retain their existing semantics.
- Require one embedding per admitted row; reject missing searchable coverage before VDB dispatch.
- Preserve image-bearing inputs' existing truncation policy. Remote admitted text uses `NONE`; images retain `END`. Mixed requests are separated only when these settings differ, and results return in original row order.

<details>
<summary>Implementation, validation, and limits</summary>

### Scope

A pure admission planner handles token counting/splitting; a pandas adapter handles row expansion/provenance. All embedding actors use one canonical runtime. Warmed and cold local construction use the same resolved checkpoint specification. The backend-neutral VDB adapter checks completeness.

Local batching, batch-size limits, Ray scheduling, model configuration, extraction chunking, retrieval settings, and index construction are unchanged. The narrow exception is remote request separation to preserve distinct text/image truncation settings. Unpinned custom remote models must supply an immutable revision or use a registered/local checkpoint rather than guessed limits.

### Latest validation

- Expanded embedding/storage suite: 449 passed; two index-statistics tests fail identically on the pre-fix commit because the installed LanceDB IndexConfig lacks num_unindexed_rows. The focused changed-area subset passed all 171 tests. New write/read regressions cover ordinary, service-schema, and collection paths; a local-checkpoint test changes metadata after actor construction and verifies admission still uses the original resolved spec.
- Real pinned tokenizers: 12 Unicode cases across legacy text, legacy VL, and Nemotron3 BF16, at configured limits 16 and 8192. Exact reconstruction, child fit, and deterministic plans passed.
- Earlier real single-H100 HF batch actor, run twice: short / 10,003-formatted-token Unicode / short produced four complete rows, including two children of 8,191 and 1,815 formatted tokens. Every vector was finite and 2,048-dimensional. Text/token reconstruction, IDs/order, and source/page attribution passed.
- The canonical VDB adapter accepted the complete result and rejected a deliberately missing child embedding.
- Latest real-H100 HF-to-storage round trip: configured limit 16 forced a 39-formatted-token synthetic parent into three children, including one whitespace-only child. Both short neighbors and all children (five 2,048-dimensional vectors) survived ordinary, service-schema, and collection writes and reads. Exact text/token reconstruction, complete split metadata, vectors, and source/page attribution passed; collection storage-key derivation was unchanged.
- Remote transport tests inspect serialized requests for mixed batches and image-only batches with a cached policy, including row/vector association.
- Independent simplicity and correctness reviews cleared the Unicode/remote-policy corrections and the subsequent storage/checkpoint-reuse fixes. Black, Flake8, and diff checks passed.

### Remaining qualification limits

Hosted inference has not been exercised for the mixed-request correction. Real vLLM qualification remains blocked by the host's NVML/driver mismatch; the successful GPU smoke uses HF. Dataset retrieval evaluations above are historical, not rerun on this head. Strict MkDocs validation is blocked by an existing CLI quickstart link and a missing environment template. CI and human approval remain separate review gates.

### Suggested review order

1. Formatted-input limits and lossless Unicode splitting.
2. Child provenance and neighboring-row preservation.
3. Remote text/image request policy and result ordering.
4. VDB completeness checks.

</details>


nemo_retriever/src/nemo_retriever/common/schemas/embedding.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+def embedding_split_metadata(
+    *,
+    content: str,
+    parent_id: str,
+    chunk_id: str,
+    chunk_index: int,
+    chunk_count: int,
+    start_token: int,
```
jperez999, line 25: Could we place this under one namespaced mapping, such as `metadata["embedding_split"]`? These top-level fields include the generic `content` key and expand the record-wide metadata contract, which makes collisions with extraction and VDB metadata more likely. A nested object would keep split provenance cohesive and easier to evolve.

nemo_retriever/src/nemo_retriever/models/inference/embedding_input.py
```diff
@@ -0,0 +1,373 @@
...
+                    parent_id=parent_id,
+                    chunk_id=chunk_id,
+                    chunk_index=chunk_index,
+                    chunk_count=len(chunks),
+                    start_token=start,
+                    end_token=end,
+                )
+            )
+            expanded.append(child)
+        return expanded
```
jperez999: Could we split pure admission/planning from DataFrame expansion and provenance mutation? `EmbeddingInputPolicy` currently owns tokenizer policy, token planning, row copying, ID derivation, metadata mutation, and pandas construction. Separating a pure split plan from its pandas adapter would make it easier to test and reuse outside this runtime.

nemo_retriever/src/nemo_retriever/common/vdb/records.py
```diff
@@ -25,6 +30,15 @@
     "table_caption": "table",
 }
 
+_CONTENT_PROVENANCE_METADATA_KEYS = (
+    "chunk_index",
```
jperez999, line 30: The generic VDB record adapter now recognizes embedding-specific split children and validates embedding coverage. Is there an existing embedding-boundary helper this can reuse or move behind? Keeping VDB conversion ingestion-agnostic avoids future embedding-policy changes requiring synchronized edits in record conversion.

nemo_retriever/src/nemo_retriever/models/inference/embedding_input.py
```diff
@@ -0,0 +1,373 @@
...
+            content_metadata.get("bbox_xyxy_norm"),
+        ),
+        "source_chunk_index": metadata.get("chunk_index"),
+        "text_sha256": hashlib.sha256(text.encode("utf-8")).hexdigest(),
+    }
+    return hashlib.sha256(_stable_json(identity).encode("utf-8")).hexdigest()
+
+
+@dataclass(frozen=True)
+class EmbeddingPreparationResult:
```
jperez999, line 110: Prefix formatting is duplicated between this admission policy and the local embedder implementations that prepare query and passage strings. Since admission counts tokens after prefixing, these need to stay byte-for-byte aligned with actual inference. Could we extract a shared embedding-input formatter so a future prefix-rule change cannot make preflight accounting disagree with the backend?

jperez999 (changes requested):

nemo_retriever/src/nemo_retriever/common/schemas/embedding.py
jioffe502: Addressed in fdf6dd11. Split provenance now lives in one `metadata["embedding_split"]` mapping containing exact content, parent/child IDs, order/count, and token range. Because this contract is introduced by this PR, I updated the producer, consumers, docs, and tests directly without a legacy flat-key path.

nemo_retriever/src/nemo_retriever/models/inference/embedding_input.py
jioffe502: Addressed in fdf6dd11. `EmbeddingInputPolicy.plan()` now returns immutable, side-effect-free `EmbeddingSplitPlan` / `EmbeddingSplitChild` values. The separate `prepare_embedding_inputs()` pandas adapter owns row copying, stable ID derivation, provenance attachment, and DataFrame construction.

nemo_retriever/src/nemo_retriever/common/vdb/records.py
jioffe502: Addressed in fdf6dd11. `records.py` no longer imports split-key lists or interprets/copies individual split fields. It calls the single `embedding_record_content()` helper, which keeps the split layout behind the embedding schema seam. I retained the generic missing-embedding check at VDB dispatch because that is the point that prevents an incomplete partial write.

nemo_retriever/src/nemo_retriever/models/inference/embedding_input.py
jioffe502: Addressed in fdf6dd11. `format_embedding_input()` is now the shared formatter used by admission, local vLLM prefix application, the local HF text embedder, and the legacy local runtime adapter. The explicit `prefix_if_missing` argument preserves the HF backend rule while keeping the formatted string identical to what admission counts.

nemo_retriever/src/nemo_retriever/models/inference/runtime.py
```diff
@@ -117,6 +123,7 @@ def embed_text_main_text_embed(
     nim_http_max_concurrent: int = 32,
     input_type: str = "passage",
     request_timeout_s: float | None = None,
+    embedding_input_policy: EmbeddingInputPolicy | None = None,
```
jperez999, line 126: **[P2] Consolidate the duplicate embedding runtime**

`nemo_retriever.common.modality.pipeline` still exports a second `embed_text_main_text_embed` implementation. That path calls `create_text_embeddings_for_df()` directly, never accepts or applies `embedding_input_policy`, and hard-codes `truncate="END"`, so callers of the exported compatibility entry point can still silently truncate the overlength text this PR is intended to preserve. It also duplicates the grouping, prefixing, error, and output-column logic being changed here. Could we make that module a thin compatibility wrapper around this implementation (or remove its export if it is truly dead) so there is one admission/runtime path?

jperez999 (changes requested):

nemo_retriever/src/nemo_retriever/models/inference/runtime.py
jioffe502: Addressed in 36566241. I confirmed repository-wide that no caller imports the old `common.modality.pipeline` embedding export; the CPU/GPU actors and tests already use `models.inference.runtime`. I removed the export and deleted the duplicate 188-line implementation rather than adding a compatibility wrapper. There is now one embedding runtime, so admission policy, no-truncation behavior, grouping, errors, and output cardinality cannot diverge between these two paths. Validation: 353 focused tests passed, plus Black, Flake8, import smoke, and `git diff --check`.

nemo_retriever/src/nemo_retriever/common/modality/pipeline/__init__.py
```diff
@@ -8,11 +8,9 @@
     collapse_content_to_page_rows,
     explode_content_to_rows,
 )
-from nemo_retriever.common.modality.pipeline.embedding import embed_text_main_text_embed
```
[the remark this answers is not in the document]
jioffe502: This removal is intentional and follows the human reviewer’s explicit “remove its export if it is truly dead” option. This is an internal pipeline helper, not a FastAPI endpoint, client method, CLI argument, or API Pydantic model covered by the compatibility rule. A repository-wide search found no callers of either old import path; all active embedding actors already import the canonical runtime. Retaining a shim would preserve the second public seam the review asked us to consolidate, so no compatibility layer is being added.

nemo_retriever/src/nemo_retriever/common/schemas/embedding.py
```diff
@@ -0,0 +1,162 @@
...
+        return None
+    requested_modality = embedding_runtime_modality(row, default_modality=default_modality)
+    if selected.is_split_child or requested_modality == "text_image":
+        return selected.content
+    return selected.content.strip()
+
+
+def embedding_record_content(row: Mapping[str, Any], *, text_column: str = "text") -> str | None:
+    """Return exact searchable text without exposing split metadata layout to callers."""
+    selected = select_embedding_text(row, text_column=text_column)
```
jperez999: **[P2] Keep VDB content selection scoped to its previous contract**

`embedding_record_content()` delegates to `select_embedding_text()`, whose fallback order includes `chunk` and `page_text`. Before this PR, VDB conversion only accepted `text`, `content`, or `metadata["content"]`. A normal row with an auxiliary `page_text`/`chunk` value can now become uploadable—or, if it lacks an embedding, cause the new fail-closed check to reject an otherwise valid mixed batch—even though that row is unrelated to overflow splitting. I reproduced the latter with one valid embedded row plus one unembedded `page_text`-only row. This contradicts the stated “normal inputs are unchanged” scope. Can we preserve the old VDB fallback order here and add only the `metadata["embedding_split"]` special case needed by this PR?

nemo_retriever/src/nemo_retriever/models/inference/embedding_input.py
```diff
@@ -0,0 +1,412 @@
...
+    for row, selected, plan in zip(rows, selected_inputs, plans):
+        first_output_position = len(prepared)
+        expanded = _expand_row(row, selected, plan)
+        prepared.extend(expanded)
+        if plan is not None and plan.requires_split:
+            split_parent_positions.add(first_output_position)
+            split_child_positions.update(range(first_output_position, first_output_position + len(expanded)))
+
+    return EmbeddingPreparationResult(
+        frame=pd.DataFrame(prepared).reset_index(drop=True),
```
jperez999: **[P2] Preserve the input DataFrame schema when one row splits**

Rebuilding the entire frame from row dictionaries makes pandas infer every dtype again. In a mixed short/overflow batch I reproduced `category -> object`, nullable `Int64 -> int64`, and nullable `boolean -> bool`, including on the fitting neighbor. The PR description says fitting inputs retain their dtypes, but that currently holds only when no row in the batch splits. Could we expand rows with pandas operations that retain the original column dtypes (or explicitly restore compatible dtypes) and cover a mixed batch with extension/categorical columns?

nemo_retriever/src/nemo_retriever/models/warmup_registry.py
```diff
@@ -63,18 +70,7 @@ def build_warmup_spec(
...
-                "device": kwargs.get("local_hf_device") or kwargs.get("device"),
-                "hf_cache_dir": kwargs.get("hf_cache_dir"),
-                "gpu_memory_utilization": float(kwargs.get("gpu_memory_utilization", 0.45)),
-                "enforce_eager": bool(kwargs.get("enforce_eager", False)),
-                "dimensions": kwargs.get("dimensions"),
-                "normalize": bool(kwargs.get("normalize", True)),
-                "max_length": int(kwargs.get("max_length", 8192)),
-                "query_max_length": int(kwargs.get("query_max_length", 128)),
-            }
+            embed_spec = LocalEmbedderSpec.from_config(kwargs).as_dict()
```
jperez999: **[P2] Keep warmup-spec construction free of model downloads**

This turns `build_warmup_spec()` from normalization/serialization into a Hugging Face config lookup. In a fresh cache with network unavailable, the otherwise unit-level `test_build_warmup_spec_full_local_stack` now fails here. The same `LocalEmbedderSpec.from_config()` resolution already runs in `warm_local_models()`, and `LocalEmbedderSpec.create()` calls `create_local_embedder()`, which resolves the checkpoint again, so this also repeats work without changing the worker's final identity. Can we keep the normalized raw settings here and resolve once in the worker, or pass/reuse one already-resolved checkpoint spec through construction?

nemo_retriever/src/nemo_retriever/common/schemas/embedding.py
jperez999: Check that splitting does a row, that might get corrupted does not prevent us from uploading the other rows in the batch.

nemo_retriever/src/nemo_retriever/models/warmup_registry.py
jperez999: Lets not redo any of these things. They are costly as is.

jperez999 (changes requested):

nemo_retriever/src/nemo_retriever/common/schemas/embedding.py
jioffe502: Fixed in 86d7b34f. VDB text selection now uses only text, content, then metadata.content, with the single embedding_split special case. Tests cover auxiliary-only chunk/page_text rows both with and without embeddings beside a valid row, plus field precedence. The previous code rejects the valid neighbor in the reported reproduction; the corrected adapter uploads it.

For the follow-up about corrupted split rows: an unrelated non-searchable row no longer vetoes valid records. A real split child missing its embedding still rejects the write, as required by the existing no-partial-write contract; this correction does not silently discard that child.

nemo_retriever/src/nemo_retriever/models/inference/embedding_input.py
jioffe502: Fixed in 86d7b34f. Expansion now repeats source rows with pandas iloc before updating child text and metadata. Unrelated columns retain their original values and dtypes. The mixed short/overflow/short regression covers category, nullable Int64, nullable boolean, and string columns, input immutability, and child page attribution. Text categories retain their categorical representation and add the new child strings.

nemo_retriever/src/nemo_retriever/models/warmup_registry.py
jioffe502: Fixed in 86d7b34f. build_warmup_spec now returns normalized settings without resolving checkpoint assets. The worker resolves one EmbedModelSpec; LocalEmbedderSpec retains it and construction consumes that same object without another config lookup. Removed the unused resolved-spec serialization method. Regression coverage forbids resolution during planning and checks that worker construction resolves metadata once and loads the model once. 395 focused tests passed with HF_HUB_OFFLINE=1.

nemo_retriever/src/nemo_retriever/common/vdb/records.py
```diff
@@ -341,7 +343,7 @@ def _client_record_from_graph_row(row: dict[str, Any], *, require_embedding: boo
     record_metadata = dict(metadata)
     if embedding is not None:
         record_metadata["embedding"] = embedding
-    record_metadata["content"] = text
+    record_metadata["content"] = "" if text is None else text
```
jperez999, line 346: **[P2] Preserve whitespace children through the actual collection write**

This adapter now intentionally preserves a whitespace-only split child, but the next collection boundary still executes `if not text.strip() and content_type != "image": continue` in `_collection_rows()`. I reproduced `to_client_vdb_records()` returning both children and `_collection_rows()` silently returning no row for the whitespace child. That leaves the completeness check passing even though the backend write drops admitted content, contradicting the PR's lossless/complete-coverage contract. Please either prevent standalone whitespace children in the split plan or update the collection writer and add an end-to-end adapter-to-storage regression.

nemo_retriever/src/nemo_retriever/common/schemas/embedding.py
```diff
@@ -0,0 +1,162 @@
...
+    parent_id: str,
+    chunk_id: str,
+    chunk_index: int,
+    chunk_count: int,
+    start_token: int,
+    end_token: int,
+) -> dict[str, Any]:
+    """Return canonical metadata for one lossless split child."""
+    return {
+        EMBEDDING_SPLIT_METADATA_KEY: {
```
jperez999, line 30: **[P2] Persist the advertised stable child identity**

The new split mapping survives `to_client_vdb_records()`, but `_collection_rows()` serializes only `content_metadata` and generates `chunk_id` from `document_id/document_version/row_index`; it drops this `embedding_split` mapping and its `chunk_id` entirely. I reproduced a canonical child with `chunk_id=stable-child` being stored with a different row-index hash and metadata containing only the content type. As a result, callers cannot identify/order split children after retrieval as the docs and PR description promise, and IDs are no longer stable when row composition changes. Please thread the split provenance/ID through the collection adapter (or narrow the stated contract if it is only intended for the transient DataFrame).

nemo_retriever/src/nemo_retriever/operators/embed/gpu_operator.py
```diff
@@ -36,41 +38,28 @@ def __init__(self, params: EmbedParams) -> None:
...
-            query_max_length=int(self._kwargs.get("query_max_length", 128)),
-            revision=self._kwargs.get("embed_model_revision"),
-        )
+        self._model = local_spec.create()
 
     def preprocess(self, data: Any, **kwargs: Any) -> Any:
         return data
 
     def process(self, data: Any, **kwargs: Any) -> Any:
+        ensure_embedding_input_policy_for_batch(self._kwargs, data)
```
jperez999, line 63: **[P3] Reuse the already-resolved checkpoint for admission**

`LocalEmbedderSpec.from_config()` has already resolved the checkpoint metadata above, but the first text batch calls `resolve_embedding_input_policy()`, which resolves the same config/prompt/limit metadata again from the IDs copied into `_kwargs`. Besides duplicate cache/config work per actor, a mutable local checkpoint can change between these resolutions, so construction and admission need not use the same spec. Could the policy factory accept `local_spec.checkpoint` (or install the policy from it here) so the claimed single resolved checkpoint is actually shared?

nemo_retriever/src/nemo_retriever/common/schemas/embedding.py
```diff
@@ -0,0 +1,162 @@
...
+    if not isinstance(split, Mapping):
+        return None
+    chunk_id = split.get("chunk_id")
+    content = split.get("content")
+    if isinstance(chunk_id, str) and chunk_id.strip() and isinstance(content, str):
+        return content
+    return None
+
+
+def embedding_split_id(metadata: Any) -> str | None:
```
jperez999: **[P3] Avoid adding an unused public helper**

`embedding_split_id()` is exported but has no caller anywhere in the repository; even the collection writer that would need the stable ID does not use it. Since this change is already broad and the goal is the smallest possible surface, please either use this helper at the persistence boundary or remove it until a consumer exists.

jperez999 (changes requested):

nemo_retriever/src/nemo_retriever/common/vdb/records.py
jioffe502: Fixed in c5eeb719. Ordinary dense, service-schema, and collection writes now retain valid whitespace-only split children; ordinary blank records still follow the existing filter. Read normalization also preserves exact child text. Three real LanceDB write/read regressions cover this path. A separate H100 HF-to-storage probe produced three children (one whitespace-only) plus both neighbors; all five rows and vectors survived storage and retrieval in all three paths.

nemo_retriever/src/nemo_retriever/common/schemas/embedding.py
jioffe502: Fixed in c5eeb719. Both writers persist the complete embedding_split mapping in JSON metadata, including parent/child IDs, order, count, content, and token ranges. The stable embedding identity is metadata.embedding_split.chunk_id. I kept the collection's existing document/version/row-index storage key unchanged and clarified that distinction in the docs. Write/read tests compare the complete split mapping and verify existing storage-key derivation.

nemo_retriever/src/nemo_retriever/operators/embed/gpu_operator.py
jioffe502: Fixed in c5eeb719. The local actor retains local_spec.checkpoint and passes it to admission instead of resolving metadata again. Tokenizer loading remains lazy, and remote resolution is unchanged. The regression changes actual local prompt/limit files after construction and verifies that admission still uses the original resolved checkpoint; warm-model reuse coverage also forbids a second admission resolution.

nemo_retriever/src/nemo_retriever/common/schemas/embedding.py
jioffe502: Removed embedding_split_id and its export in c5eeb719. No replacement helper or new abstraction was added.

jperez999 (approved):

PR #2632, opened 2026-09-03
feat: use Nemotron 3 Embed as the default embedding model

ChrisJar:
## Summary

- change the default NIM and local embedding model to Nemotron-3-Embed-1B
- select NVFP4 on Blackwell GPUs and use BF16 for other supported local configurations
- update Helm, Compose, service configuration, tests, and documentation for the text-only default
- set NIM_ENGINE_COUNT=1 by default and document NIM_PERFORMANCE_MODE=1 as an optional embed-NIM setting
- persist embedding-model identity on LanceDB legacy and collection indexes
- reject mismatched or untagged dense/hybrid indexes with explicit rebuild guidance; sparse retrieval remains exempt
- document the public local-model resolver contract and persistent-index migration path

## Validation

- pre-commit: passed
- focused Helm option test: 1 passed, 3 skipped because Helm is unavailable
- focused persistence/model suite: 188 passed, 7 deselected
- broader model, Helm, service, Retriever, and LanceDB suite: 321 passed, 59 skipped, 7 deselected, 17 subtests passed
- Python compile and YAML parsing: passed
- git diff --check: passed
- Helm rendering and strict MkDocs validation unavailable because those executables are not installed in this environment


nemo_retriever/src/nemo_retriever/models/__init__.py
```diff
@@ -25,19 +28,20 @@
     }
 )
 
-# Short name → full HF repo ID.
+# Short name to canonical model ID.
 _EMBED_MODEL_ALIASES: dict[str, str] = {
+    "nemotron-3-embed-1b": NEMOTRON_3_EMBED_MODEL,
```
jperez999, line 33: [P1] Canonicalize the BF16 and NVFP4 checkpoint IDs here before the new persistent-index checks compare them. `resolve_local_embed_model()` deliberately selects either concrete ID for the same logical default, and the new embedding docs explicitly say they share an embedding space. However, neither concrete ID is an alias in this map, so an index created with `nvidia/Nemotron-3-Embed-1B-BF16` cannot be opened by a service configured with the logical name (or the NVFP4 variant): `_production_vdb`, collection access, and `Retriever` all compare the raw, different strings and force a rebuild. Add both IDs as aliases of `NEMOTRON_3_EMBED_MODEL` (and a regression test) so the documented precision switch remains usable.

nemo_retriever/docs/cli/README.md
```diff
@@ -381,6 +396,8 @@ These options apply to `retriever ingest`, `retriever ingest local`, and
 | `--table-name` | `nemo-retriever` | LanceDB table name. Must match query-time storage flags. Python `.vdb_upload()` and default `Retriever()` use the same default. |
 | `--overwrite/--append` | overwrite | Overwrite the table by default; use `--append` to add rows. |
 | `--index-mode` | `auto` | Recommended: leave this unset. `auto` creates a hybrid vector + BM25/FTS configuration for new tables and preserves an existing table on append. Use `dense`, `hybrid`, or `sparse` only for explicit experiments or specialized deployments. |
+| `--embed-model-name` | `nvidia/nemotron-3-embed-1b` | Logical default embedding model. Local vLLM resolves it to the NVFP4 checkpoint on Blackwell and BF16 otherwise. |
```
jperez999, line 399: [P2] This now disagrees with the public CLI help. `cli/ingest/options.py` and `cli/query/options.py` still import `VL_EMBED_MODEL` and expose it as `DEFAULT_EMBED_MODEL`, so `retriever ingest --help` advertises the old VL model and `retriever query --help` says legacy tables fall back to it. The runtime resolver has changed, but the stale help makes the documented default and the command interface contradictory. Update those shared CLI constants/imports as part of this default-model change.

jperez999, line 399: [P1] The `retriever recall` deployment path is still left on the old embedding family. Its `RecallConfig`, `recall-with-main`, and `run` commands default to legacy model IDs and the HF backend. Running recall against an index created with the newly documented default therefore embeds queries in an incompatible space and produces misleading metrics. Please migrate these defaults (or explicitly make recall read the index metadata) and cover the path in the default-model tests.

jperez999 (changes requested):

nemo_retriever/src/nemo_retriever/models/__init__.py
```diff
@@ -51,6 +57,41 @@ def is_vl_embed_model(model_name: str | None) -> bool:
...
+            return False
+        return all(torch.cuda.get_device_capability(index)[0] >= 10 for index in range(torch.cuda.device_count()))
+    except (ImportError, RuntimeError, AssertionError):
+        return False
+
+
+def resolve_local_embed_model(model_name: str | None, *, backend: str) -> str:
+    """Resolve an embedding model to a backend-compatible checkpoint.
+
+    Args:
```
jperez999: [P2] Preserve explicitly selected precision checkpoints. Because the BF16 and NVFP4 IDs are aliases, resolving the name before checkpoint selection means an explicit BF16 request on Blackwell returns NVFP4, while an explicit NVFP4 request with `backend="hf"` returns BF16. These concrete IDs are advertised as selectable checkpoints in the CLI docs, so silently replacing one can change memory and compatibility characteristics. Please auto-select precision only for an omitted/logical model ID, keep concrete checkpoint IDs intact for loading, and use canonicalization separately for embedding-space equality. A pair of explicit-ID tests would capture the distinction.

nemo_retriever/src/nemo_retriever/service/vectordb_app.py
```diff
@@ -256,8 +259,27 @@ def _production_vdb(
...
+        embedding_model_name=resolved_embed_model,
         expiration_cleanup_enabled=expiration_cleanup_enabled,
     )
+    if existing_mode is None:
+        return backend
+
+    stored_embed_model = backend.get_index_metadata("embedding_model_name")
+    if not stored_embed_model:
+        raise ValueError(
+            f"Existing LanceDB table {table_name!r} at {lancedb_uri!r} does not record its embedding "
```
jperez999: [P1] Keep startup acceptance consistent with the append guard. This comparison accepts an existing BF16- or NVFP4-tagged table as compatible, then returns a backend configured with the canonical logical ID. On the next `backend.run(...)`, `_validate_append_embedding_model` compares the stored and configured strings without canonicalizing and rejects the append (I reproduced startup succeeding followed by `uses embedding model ... BF16; cannot append vectors from ... nemotron-3-embed-1b`). That makes the documented compatible precision migration fail every subsequent ingest. Please reuse one shared model-identity compatibility helper in `_production_vdb`, the LanceDB append validator, and the collection validator so all three paths apply the same policy.

nemo_retriever/src/nemo_retriever/graph/retriever.py
```diff
@@ -454,22 +509,27 @@ def queries(
...
                 resolved_index_model = resolve_embed_model(index_model)
                 if resolved_explicit_model != resolved_index_model:
-                    logger.warning(
-                        "The explicitly configured query embedding model %r differs from the model %r "
-                        "recorded on the index. Results may be unreliable because different embedding "
-                        "models can use incompatible vector spaces. Use the index model or rebuild the "
-                        "index with the query model. Continuing with the explicitly configured model.",
-                        resolved_explicit_model,
-                        resolved_index_model,
+                    raise ValueError(
```
jperez999: [P1] Migrate the remaining callers that force the old model before making mismatches fatal. `common/io/export.py::query_lancedb` and `export_retrieval_json`, plus `FileRetriever.from_lancedb` / the `retriever eval` configuration path, still default to `nvidia/llama-nemotron-embed-1b-v2`. I reproduced creating a table with the new default and calling `query_lancedb(...)` with its defaults; it now fails at this line before retrieval. The more reusable fix is to make the embedder override optional and omit explicit embed kwargs when absent, letting `Retriever` use the table metadata; otherwise these public defaults all need to move together.

jperez999 (approved):

PR #2633, opened 2026-09-03
Expose cache-read token usage for agentic queries

mahikaw:
## Description
- Normalize provider-reported cache-read tokens as `usage.cache_tokens`.
- Expose cache-read usage through Python, CLI, service, and MCP responses.
- Preserve the original provider breakdown under `usage.stages`.
- Keep cache tokens as a subset of input tokens so they are not added again to totals.
- Update API schemas, CLI documentation, workflow documentation, and tests.
## Semantics
`input_tokens` represents total input processed, including cached input.
`cache_tokens` is the observed cache-read portion of that input.
`total_tokens` remains `input_tokens + output_tokens`.
Cache creation details remain available in the provider-specific stage breakdown.

## Checklist

### DIFF
--- a/nemo_retriever/src/nemo_retriever/_agentic/nemo_agent/llm/usage.py
+++ b/nemo_retriever/src/nemo_retriever/_agentic/nemo_agent/llm/usage.py
@@ -218,6 +218,34 @@ def sum_usage_breakdown(usage_by_stage: Optional[Mapping[str, Any]]) -> Dict[str
     return total
 
 
+def usage_integer(usage: Mapping[str, Any], *keys: str) -> Optional[int]:
+    """Return the first integer usage value found under ``keys``."""
+    for key in keys:
+        value = usage.get(key)
+        if isinstance(value, int) and not isinstance(value, bool):
+            return int(value)
+    return None
+
+
+def cache_read_tokens(usage: Mapping[str, Any]) -> Optional[int]:
+    """Return provider-reported cache-read input tokens, when available."""
+    cached = usage_integer(
+        usage,
+        "cached_tokens",
+        "cached_input_tokens",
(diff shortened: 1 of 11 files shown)


nemo_retriever/src/nemo_retriever/_agentic/nemo_agent/llm/usage.py
```diff
@@ -264,11 +267,25 @@ def _token_count(
             ("input_tokens", "prompt_tokens"),
             additive_keys=("cache_creation_input_tokens", "cache_read_input_tokens"),
         )
+        stage_cache = _integer(
+            usage,
+            "cached_tokens",
+            "cached_input_tokens",
+            "cache_read_input_tokens",
+        )
+        if stage_cache is None:
```
jperez999: [P2] Keep cache-read parsing shared with ATIF. This path now recognizes `input_tokens_details.cached_tokens`, but the existing `_atif_metrics()` helper only checks `prompt_tokens_details`. A provider that reports only `input_tokens_details` will therefore expose cache reads in CLI, service, and MCP usage while omitting them from ATIF trajectory metrics. Extract the integer/cache-read lookup into a shared helper and use it from both paths so provider support cannot drift.

jperez999 (changes requested):

nemo_retriever/src/nemo_retriever/_agentic/nemo_agent/llm/usage.py
mahikaw: addressed

jperez999 (approved):
Approved after re-review. The follow-up commit shares cache-read parsing between normalized usage and ATIF, including input_tokens_details coverage, resolving the prior finding.

PR #2637, opened 2026-09-03
fix(service): return complete result payloads

ChrisJar:
## Description

- Preserve complete `text` and `content` strings in legacy service result rows while retaining transport sanitization for incidental values.
- Honor `return_embeddings=True` for compact result rows by returning a normalized top-level `embedding` field.
- Document the service result schema behavior and add regression coverage for both embedding storage shapes.

## Checklist

## Validation
- `pre-commit run --all-files`
- `pytest nemo_retriever/tests/test_ingest_results.py nemo_retriever/tests/test_service_pipeline_spec.py nemo_retriever/tests/test_service_ingest_async.py -q` (136 passed)


jperez999 (changes requested):
looks good, just need to cover the other possible cases.

jperez999 (approved):

PR #2638, opened 2026-09-04
Update NeMo Retriever CLI and MCP skills

mahikaw:
## Description
- update the NeMo Retriever CLI skill for the 26.8.1 release
- use a project-local `.venv` instead of a Harbor-specific path
- add the environment-agnostic Retriever MCP skill
- remove evaluation-specific MCP guidance
- generalize citation handling and query retry behavior

## Checklist


skills/nemo-retriever/SKILL.md
```diff
@@ -1,52 +1,62 @@
 ---
 name: nemo-retriever
-description: "Use when the user wants to search, query, extract, transcribe, describe, quote, filter, or aggregate across documents — PDFs, scanned forms / images (`.jpg` `.png` `.tiff`), Office (`.docx` `.pptx`), text (`.html` `.txt`), audio (`.mp3` `.wav` `.m4a`), or video (`.mp4` `.mov`). Prefer this over native Read / Grep for multi-file or non-PDF corpora. Not for: editing files, web browsing, single-file plain-text lookups, fine-tuning."
-license: Apache-2.0
-allowed-tools: Bash Write Read
+description: Use when searching, extracting, ingesting, or querying a document collection with the NeMo Retriever 26.8.1 CLI, including local LanceDB indexes and deployed Retriever services. Use for PDFs, images, Office files, HTML, text, audio, and video; not for editing documents or web search.
 ---
 
-# nemo-retriever
+# NeMo Retriever 26.8.1
```
jperez999, line 6: Lets remove the version 26.8.1 here.

skills/nemo-retriever/SKILL.md
```diff
@@ -1,52 +1,62 @@
...
-The `retriever` CLI indexes a folder of PDFs into LanceDB (`retriever ingest`) and serves vector search over it (`retriever query`). For any task about searching/answering questions across a folder of PDFs, use this CLI — do not write a custom RAG.
+Use the `retriever` CLI. Prefer it over hand-built retrieval code.
 
-**Beyond PDFs and beyond semantic search.** `retriever ingest` also handles images, Office, HTML, TXT, audio, and video — see `references/setup.md` for the per-format recipe and `references/install.md` for the install extras (`[multimedia]`, libreoffice, ffmpeg). The query turn is two retrieval passes — see **§Query turn** below (inline, no reference read needed); `references/cli/query.md` holds only the fallback detail (exact-term, chart text-extract, compose-reply). Don't fall back to native Read/Grep/Python on non-PDF inputs.
+## Install only when missing
 
-## Install (if `retriever` is missing)
+```bash
+uv venv .venv --python 3.12
+uv pip install --python .venv/bin/python "nemo-retriever==26.8.1"
```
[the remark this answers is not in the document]
jperez999: I think this is valid, we are probably going to want to setup the local version for service mode. Just in case the user wants to use hf models instead of endpoints.

skills/nemo-retriever/SKILL.md
```diff
@@ -1,52 +1,62 @@
...
+export PATH="$PWD/.venv/bin:$PATH"
+```
 
-If `command -v retriever` returns nothing, follow `references/install.md` to install the NeMo Retriever Library before proceeding. It prints `RETRIEVER_VENV=<path>`; substitute that path for `<RETRIEVER_VENV>` in every example in this skill (setup, query, troubleshooting, and the CLI references).
+Do not clone NeMo Retriever or install from a Git URL. If `retriever` is already
```
jperez999, line 18: Should this say to install the aforementioned version, Wdyt?

jperez999: What I mean to say is that the version should be updated to the version assigned in the skill, in this case 26.8.1, but open to opinions here.

skills/retriever-mcp/SKILL.md
```diff
@@ -0,0 +1,32 @@
...
+- Use `ingest_documents` only when required documents are not yet searchable
+  and server-readable paths are explicitly available. Do not assume that paths
+  visible to the agent are also readable by the Retriever server.
+- After ingestion completes, use `query` to retrieve evidence from the added
+  documents.
+- Start with `top_k=5` and `format="hits"`. Request more only when the first
+  successful response does not contain enough evidence to answer the task.
+- Read evidence hits from the tool response and ground your answer in them.
+- Preserve `doc_id` and page numbers from MCP hit metadata when the task
+  requests citations or structured output.
```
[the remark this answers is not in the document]
jperez999: Do we feel this is valid?

jperez999 (changes requested):

skills/nemo-retriever/SKILL.md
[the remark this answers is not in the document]
mahikaw: agreed, this is useful to add.

skills/nemo-retriever/SKILL.md
mahikaw: it could be a nice feature to have from the skills capabilities POV, but can definitely lead to unintended/surprise updates that can potentially break things based on version diffs. I would say we observe this behavior for some time if it seems useful and then promote it here.

skills/retriever-mcp/SKILL.md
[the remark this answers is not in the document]
mahikaw: this is fair, reduces the burden on the agent.

skills/nemo-retriever/SKILL.md
```diff
@@ -1,52 +1,81 @@
 ---
 name: nemo-retriever
-description: "Use when the user wants to search, query, extract, transcribe, describe, quote, filter, or aggregate across documents — PDFs, scanned forms / images (`.jpg` `.png` `.tiff`), Office (`.docx` `.pptx`), text (`.html` `.txt`), audio (`.mp3` `.wav` `.m4a`), or video (`.mp4` `.mov`). Prefer this over native Read / Grep for multi-file or non-PDF corpora. Not for: editing files, web browsing, single-file plain-text lookups, fine-tuning."
-license: Apache-2.0
-allowed-tools: Bash Write Read
+description: Use when searching, extracting, ingesting, or querying a document collection with the NeMo Retriever 26.8.1 CLI, including local LanceDB indexes and deployed Retriever services. Use for PDFs, images, Office files, HTML, text, audio, and video; not for editing documents or web search.
```
jperez999, line 3: **[P1] Regenerate or remove stale signed artifacts**

`skill.oms.sig` is unchanged: its DSSE payload still signs the previous `SKILL.md` digest and enumerates references/scripts deleted here. `skill-card.md` also still identifies evaluated SHA `b331d0f7`, and `BENCHMARK.md` reports the May workflow. Shipping this rewritten skill with the old signature and evaluation claims leaves the bundle unverifiable and misrepresents which behavior passed evaluation. Please regenerate the signature/evaluation/card for this content, or remove the stale publication artifacts until that happens.

skills/nemo-retriever/SKILL.md
```diff
@@ -1,52 +1,81 @@
...
+uv pip install --python .venv/bin/python \
+  "nemo-retriever[service,local]==26.8.1"
 
-- **Semantic pass** — the full question, hybrid (dense + lexical fusion):
-  `<RETRIEVER_VENV>/bin/retriever query "<question>" --format evidence --retrieval-mode hybrid --top-k 10`
-- **Lexical pass** — the EXACT term/figure/code/proper-noun the question targets (just the term, not the whole question — that's what makes BM25 precise):
-  `<RETRIEVER_VENV>/bin/retriever query "<exact term, e.g. Management VaR / Level 3 / a code>" --format evidence --retrieval-mode sparse --top-k 10`
+# Local audio or video ingestion
+uv pip install --python .venv/bin/python \
+  "nemo-retriever[local,multimedia]==26.8.1"
```
jperez999, line 34: **[P1] Keep the required system prerequisites**

The `[local,multimedia]` extra installs Python packages only; the checked-in audio/video docs explicitly say it does not install the `ffmpeg`/`ffprobe` binaries. DOCX/PPTX ingestion likewise requires the `libreoffice` binary. Because this PR deletes `references/install.md`, users following the advertised Office/audio/video workflows now have no prerequisite step and will fail at runtime. Please retain one concise prerequisite sentence (or link directly to the published prerequisite sections) for FFmpeg and LibreOffice.

skills/nemo-retriever/SKILL.md
```diff
@@ -1,52 +1,81 @@
...
+  --lancedb-uri lancedb --table-name nemo-retriever
+```
 
-Spend the calls you need to get the figures right — accuracy matters more than minimizing calls here. Only avoid genuinely wasteful loops (re-running identical queries, reading whole PDFs, 15+ calls). **A fully-supported answer beats a cheap partial one.**
+Query it:
+
+```bash
+retriever query "<question>" \
+  --lancedb-uri lancedb --table-name nemo-retriever \
+  --top-k 5 --format evidence
```
jperez999, line 54: **[P1] Do not cap exhaustive tasks at five results**

The retained evaluations require “every mention” and “every clause,” but this command returns at most five items and the rewrite removes the previous exact-term/re-query guidance. That makes the skill silently incomplete for tasks it still claims to support. A concise shared policy with the MCP skill would work: start at five for ordinary Q&A, but increase `top-k` and re-query exact terms until the requested coverage is satisfied.

skills/retriever-mcp/SKILL.md
```diff
@@ -0,0 +1,33 @@
...
+available; an empty MCP resource list is expected.
+
+## Workflow
+
+- Use `query` to search documents already available to Retriever.
+- Never ingest documents that are already searchable, including documents in a
+  prebuilt index.
+- Use `ingest_documents` only when required documents are not yet searchable
+  and server-readable paths are explicitly available. Do not assume that paths
+  visible to the agent are also readable by the Retriever server.
```
jperez999, line 22: **[P1] Preserve remote-agent uploads**

`MCPDocumentInput` explicitly accepts either `path` or `content_base64`, with `content_base64` intended for remote agents. Requiring a server-readable path here forbids ingestion in exactly the case the inline upload path was designed to solve. Please allow `content_base64` plus `filename` when the agent can read the bytes but the Retriever server cannot see its filesystem; keep the current warning only for `path` inputs.

skills/retriever-mcp/SKILL.md
```diff
@@ -0,0 +1,33 @@
...
+- After ingestion completes, use `query` to retrieve evidence from the added
+  documents.
+- Start with `top_k=5` and `format="hits"`. Request more only when the first
```
jperez999, line 25: **[P2] Reuse the citation-ready evidence projection**

The MCP `query` tool supports `format="evidence"`, and the service routes that through the same `build_evidence_result` implementation as the CLI, yielding normalized `source`, `locator`, and `citation` fields. Recommending raw `hits` and then teaching the agent to branch across `document_id`/`source_id`/`pdf_basename` duplicates that product logic and can still produce inconsistent citations. Use `evidence` when citations or grounded answers are requested, and reserve `hits` for consumers that actually need the raw response.

skills/nemo-retriever/scripts/doctor.py
```diff
@@ -1,243 +0,0 @@
-#!/usr/bin/env python
```
jperez999, line 1: **[P2] Keep or separately justify the reusable contract check**

The PR description covers skill guidance changes, but not removal of the CLI contract, evidence schema, and live compatibility probe. The rewritten skill still depends on exact CLI flags and the evidence response shape, so making the agent-facing prose shorter does not make this maintainer validation obsolete. Can we retain and update a trimmed doctor/contract, or split this deletion into a separately explained change with replacement validation?

edknv (approved):

jperez999 (approved):

PR #2640, opened 2026-09-04
Add agentic answer generation functionality

Reza-esfandiarpoor:
This PR adds the agentic answer generation functionality to the `_agentic` module.

The interface is very similar to the existing document selection mode. The difference is that the agent should be instantiated with `mode='answer'` and then in the results returned by the agent, there will be two fields (`answer` and `citations`) that contain the agent-generated data.

This PR just adds the feature to `_agentic` module but does not expose it to NRL users. Future changes should add the proper graph operators and interfaces to make use of the agentic answer generation mode in NRL.

It is best to add the answer generation mode in addition to the existing chunk selection (instead of replacing it), especially with MCP tools (i.e., defining a new MCP tool for agentic answer generation). That allows us to experiment with both methods side by side.


nemo_retriever/src/nemo_retriever/_agentic/nemo_agent/results.py
```diff
@@ -44,22 +44,24 @@ class AgentError:
...
+    ``error`` still says why the run failed and ``succeeded`` stays ``False``.
+    It is ``None`` only when no usable end attempt was ever made.
 
     The verbose fields (``trajectory``, ``retrieval_log``, ``extra_data``)
     are always populated. ``atif_trace`` is the lightweight ATIF rendering of
     the run when trace construction succeeded.
     """
 
     final_doc_ids: List[str] = field(default_factory=list)
+    answer: Optional[str] = None
```
mahikaw: Adding these fields before the existing dataclass fields silently changes positional construction: AgentRunResult(ids, payload, error) now binds payload to answer and error to citations. Internal callers use keywords, but this class is exported. Could we append answer and citations after the existing fields to preserve constructor compatibility?

mahikaw (changes requested):
lgtm. @Reza-esfandiarpoor could you address the greptile comments and another small comment, before merging it.

mahikaw (approved):

PR #2641, opened 2026-09-04
Improve description of MCP tools

Reza-esfandiarpoor:
Update the descriptions of the `query` and `agentic_query` MCP tools to clarify for the consumer what each tool's capabilities are and remove unnecessary details.


nemo_retriever/src/nemo_retriever/service/mcp_server.py
```diff
@@ -581,11 +580,10 @@ async def query(
         @mcp.tool(
             name="agentic_query",
             description=(
-                "Run the configured agentic (ReAct) retrieval workflow over ingested "
-                "documents via POST /v1/query with agentic=true. Returns the standard "
-                "hits envelope with the same chunk-level fields as classic retrieval, "
-                "plus top-level doc_id, rank, and result_source for the selecting "
-                "stage; rank and result_source also remain under metadata for compatibility."
+                "A smart, high-level retrieval tool that understands complex queries. "
```
oliverholworthy, line 583: I think moving away from endpoint details and describing the retrieval behaviour makes these descriptions more useful to the agent selecting a tool.

One thing i wonder is if it would be helpful to make the expected calling pattern more explicit? Since agentic_query performs decomposition and iterative retrieval internally, one call should usually be sufficient for a complex or multi-part query. This would distinguish it from query and help avoid redundant decomposition or repeated calls by the outer agent.

Reza-esfandiarpoor: This is a great point. There have been some concerns that such instructions could limit what the agent can/would do in practice. That is why I did not include such guidelines here.

jperez999 (approved):
Looks like you just need to fix formatting, then feel free to merge.

PR #2685, opened 2026-09-14
Batch remote OCR crops across page rows

jioffe502:
## Description

Batch remote OCR crops across page rows instead of waiting for each page to finish before submitting the next. This covers inprocess, Ray batch, and service ingestion through the shared OCR path. Tracks NVBug 6642584.

- Reuse the existing NIM request pool, with bounded crop windows and ordered page/region output.
- Forward the effective OCR batch size and remote settings through PDF and automatic extraction.
- Keep transport retries in the HTTP client. Only split multi-page input rejections when every concurrent failure is HTTP 400/413/422; do not replay authentication, exhausted throttling, or transport failures per page.
- Discard earlier-window OCR results when a remote request or response processing fails a page.

No new dependencies, worker pools, configuration fields, model changes, or pipeline-stage overlap. Six changed files: four runtime files, tests, and the existing performance guide.

## Validation

- 237 focused tests passed after the review fixes in `2c3b43f6`; all 76 OCR batching tests also passed with socket connections/listeners blocked. New regressions cover mixed concurrent failures in both exception-observation orders and later-window parsing/stitching failures.
- Fourteen targeted failure regressions fail against the earlier implementation and pass after review fixes.
- Pinned Black/Flake8 and whitespace checks passed; all commits are signed off.
- Real-NIM success-path validation, before the failure-handling-only revisions: inprocess PDF/auto, Ray batch PDF/auto, service whole-document PDF, and service SDK auto on RTX PRO 6000 Blackwell Server Edition. The same synthetic four-page PDF returned identical ordered text and crop hashes in all six routes. Four sequential three-crop requests became two overlapping requests of eight and four crops.

- Full BO767 inprocess extraction on `0b0b4675`, before these failure-path-only review fixes: 767 PDFs / 54,730 pages completed with zero errors, identical per-page content/counts, and identical 30,560 OCR crop hashes versus the PR base. Same four document workers and locally hosted HTTP NIMs. OCR POSTs fell 24,597 → 4,152; wall time fell 25.3 → 23.7 minutes; sampled peak VRAM stayed at 4,880 MiB. Cross-page HTTP overlap within one document was confirmed. Individual request latency increased with fuller batches and higher concurrency. This was one instrumented extraction pair, not the retrieval benchmark.

HTTP overlap is not proof of concurrent GPU kernels or customer throughput. Customer-document and externally hosted API validation remain outstanding. The locked environment lacks MkDocs, so the strict documentation build was not run successfully.

```bash
uv run --locked pytest -q \
  tests/test_ocr_cross_page_batching.py tests/test_ocr_version_selection.py \
  tests/test_service_pipeline_spec.py tests/test_nim_tracing.py \
  tests/graph/test_multi_type_grouping.py tests/test_nim_chat_completion_step.py \
  tests/test_remote_embedding_http.py tests/test_table_structure_nim_empty_bbox.py
```

## Related changes

#2658 independently fixes the named `inference_batch_size` binding. This PR overlaps that small correction, but adds cross-page scheduling, bounded submission, failure handling, and mode-specific wiring. Reconcile that overlapping hunk when deciding merge order. The local-only batching change in #2421 is unchanged.

## Checklist


nemo_retriever/src/nemo_retriever/common/modality/ocr/shared.py
```diff
@@ -809,55 +821,106 @@ def _run_remote_ocr(
...
+            return
+        try:
+            response_items = _invoke(jobs)
+        except Exception as exc:
+            jobs_by_page: Dict[int, List[_RemoteOCRCropJob]] = {}
+            for job in jobs:
+                jobs_by_page.setdefault(job.row_index, []).append(job)
+            # Only input/payload rejections can benefit from splitting a
+            # multi-page call. Transport retries belong to the NIM client.
+            isolate_pages = (
```
jperez999, line 888: [P1] Preserve all concurrent failure categories before page fallback

`_invoke()` can have several HTTP futures outstanding, but `NIMClient` raises only the first exception observed and, after waiting for the others, discards their exceptions. A window can therefore have one chunk return 400/413/422 while another times out or exhausts 429/5xx retries. If the input rejection wins the race, this condition retries every represented page, including work whose actual failure category must not be replayed. That contradicts the retry policy stated in the PR and can multiply load during a partial outage. Please retain the per-range/all failure categories and allow page isolation only when no concurrent non-isolatable failure occurred. A regression with `max_batch_size=1`, a pool of 2, and mixed failures would cover this.

nemo_retriever/src/nemo_retriever/common/modality/ocr/shared.py
```diff
@@ -809,55 +821,106 @@ def _run_remote_ocr(
...
+                    label_name=job.label_name,
+                    bbox=job.bbox,
                     preds=preds,
                     crop_hw=crop_hw,
                     use_table_structure=use_table_structure,
                 )
-        except BaseException as exc:
-            _record_ocr_error(row_result, exc)
+            except Exception as exc:
+                _record_ocr_error(row_results[job.row_index], exc)
```
jperez999, line 868: [P2] Clear prior output when stitching a page fails

This records a post-processing error but leaves any results accumulated for the same page in earlier windows intact, so `_build_ocr_page_elements_output()` can publish partial table/chart/text output alongside the page error. During page-isolation fallback this catch also prevents the outer handler at line 900 from calling `_fail_page()`. Please reset this page here (or propagate the failure so it is reset) and add a regression where a later-window response fails parsing or table stitching.

jperez999 (changes requested):

nemo_retriever/tests/test_ocr_cross_page_batching.py
```diff
@@ -182,6 +256,355 @@ def test_local_actor_batches_compatible_crops_across_page_rows() -> None:
...
+        actor.ocr_kwargs.update(invoke_url=url, api_key="test-only")
+        actor._remote_retry = RemoteRetryParams(remote_max_pool_workers=2, remote_max_retries=1)
+        bbox = [0.0, 0.0, 1.0, 1.0]
+        pages = [_page(str(i), i, [_detection(str(i), "chart", bbox)]) for i in (11, 22, 33, 44)]
+        result = actor(pd.DataFrame(pages))
+    finally:
+        client.shutdown()
+    assert state["peak"] == 2
+    assert len(state["calls"]) == 4 // batch_size
+    assert all(len(ids) == batch_size and authorized for ids, authorized in state["calls"])
```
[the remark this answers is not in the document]
jperez999: You can decide if you want to fix this, its a loopback address so, I am on the fence about needing a mock here.

nemo_retriever/src/nemo_retriever/common/modality/ocr/shared.py
jioffe502: Fixed in 2c3b43f6. NIMClient now retains concurrent failures in an ExceptionGroup after draining the pool; a single failure keeps its original exception type. OCR allows page isolation only when every observed failure is HTTP 400/413/422. The regression uses batch size 1 and pool size 2, forces both exception-observation orders, and covers mixed auth, exhausted 429/503, timeout, connection, and response-count failures. It checks that neither page is replayed and both failure categories are retained. Pure input-rejection groups still allow isolation. All 237 focused tests pass.

nemo_retriever/src/nemo_retriever/common/modality/ocr/shared.py
jioffe502: Fixed in 2c3b43f6. Stitching failures now call the existing _fail_page helper, clearing all accumulated output for that page. Added regressions for later-window parsing and table-stitching failures, both during normal batching and page-isolation fallback. They verify earlier results are removed, detection counts reset to zero, and the other page retains its complete output.

nemo_retriever/tests/test_ocr_cross_page_batching.py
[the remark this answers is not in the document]
jioffe502: Replaced the loopback server in 2c3b43f6 with a mocked requests.post transport. NIMClient, response normalization, retries, and the thread pool remain real. Event synchronization forces overlap and out-of-order completion without sleeps. Auth, chunking, ordering, status handling, and drain coverage are retained. All 76 OCR batching tests also pass with socket connect/connect_ex/bind explicitly blocked.

jperez999 (approved):

PR #2694, opened 2026-09-15
fix(ingest): add caption dedup opt-out

KyleZheng1284:
## Description

- Adds `--dedup/--no-dedup` support across local, batch, and service ingestion.
- Preserves automatic caption dedup by default while allowing users to retain every extracted image crop.
- Reports effective dedup settings during dry runs without mutating pipeline state.
- Skips the dedup operator when both dedup mechanisms are disabled.
- Adds regression tests and updates captioning documentation.

## Checklist

- [x] I am familiar with the [[Contributing Guidelines](https://github.com/NVIDIA/NeMo-Retriever/blob/main/CONTRIBUTING.md)](https://github.com/NVIDIA/NeMo-Retriever/blob/main/CONTRIBUTING.md).


nemo_retriever/src/nemo_retriever/cli/ingest_workflow.py
```diff
@@ -94,21 +96,50 @@ def _ingest_plan_to_dry_run_data(plan: ResolvedIngestPlan) -> dict[str, Any]:
...
 
+def _service_dedup_dry_run_data(
+    request: ServiceIngestRequest,
+) -> tuple[dict[str, Any] | None, dict[str, Any] | None]:
+    """Return effective service dedup parameters and mixed-input scope."""
+
+    dedup_params = request.dedup_params
+    dedup_scope: dict[str, Any] | None = None
+    if dedup_params is None and request.caption_params is not None:
+        families = {input_type_for_path(document) for document in request.documents}
```
jperez999: Could this reuse a shared service input-family/dedup-scope resolver instead of classifying paths again in the dry-run formatter? `_split_config_for_auto_documents()` already performs nearly the same service-side family discovery, while `resolve_effective_dedup_params()` owns the policy for the other paths. Keeping a dry-run-only copy of both decisions makes drift likely and adds more surface than this flag needs; a single helper that returns the effective params plus optional scope would keep reporting aligned with execution.

jperez999 (approved):

PR #2719, opened 2026-09-17
feat(agentic): expose answer mode across public APIs

mahikaw:
## Description
Exposes agentic answer generation through NeMo Retriever’s public interfaces while preserving the existing agentic document-selection behavior as the default.
## Changes
- Adds `AgenticRetriever.answer()` and `answer_with_usage()`.
- Adds `--agentic-mode answer` to the query CLI.
- Exposes answer mode through:
  - `/v1/query`
  - `/v1/answer` with `mode="agentic"`
  - Python service client
  - MCP
- Returns integrated answers, validated citation IDs, rehydrated citation hits, status metadata, and optional token usage.
- Rejects duplicate query IDs to prevent result and usage collisions.
- Validates proxied VectorDB responses as `AgenticAnswerResponse`.
- Uses the configured agentic timeout for service requests.
- Prevents select-mode `top_k` constraints from affecting answer prompts.
- Documents the public API and select/answer usage-output differences.
## Design
Answer mode terminates inside the ReAct loop using `LogAnswer`. It intentionally bypasses RRF and `SelectionAgentOperator`, which produce ranked document selections rather than integrated answers.
The change is additive and backward compatible:
- `agentic_mode="select"` remains the default.
- Existing agentic retrieval output is unchanged unless answer mode is explicitly selected.

## Checklist


nemo_retriever/src/nemo_retriever/_agentic/nemo_agent/agent.py
```diff
@@ -287,6 +290,20 @@ async def _dispatch_tool_call(
             except Exception as e:
                 raise ToolExecutionError(fn_name, e) from e
             return content, False
+        if isinstance(tool, LogAnswer):
+            citations = fn_kwargs.get("citations")
```
jperez999, line 294: [P1] Validate citations before salvaging malformed end calls

This guard runs only when every citation is already a string. For `citations=["ghost", 42]`, execution falls through to the base end-tool handler: strict validation rejects `42`, but `salvage_payload` retains `"ghost"`; after `max_steps`, the result exposes `citations=["ghost"]` even though that ID was never retrieved. I reproduced this with `max_steps=1`. That breaks the advertised validated-citation contract on failed/best-effort answers. Please validate every string citation against `state.retrieved_docs` before delegation, or filter/clear salvaged citations against that set.

nemo_retriever/src/nemo_retriever/service/client.py
```diff
@@ -597,6 +597,19 @@ async def aquery(
             return [self._query_hit(hit) for hit in parsed[0]]
         return parsed
 
+    def agentic_answer(self, query: str, *, top_k: int = 5) -> AgenticAnswerResponse:
+        """Research and answer a query with the service's integrated ReAct agent."""
+        return self._run(self.aagentic_answer(query, top_k=top_k))
+
+    async def aagentic_answer(self, query: str, *, top_k: int = 5) -> AgenticAnswerResponse:
+        """Asynchronously run integrated agentic answer mode."""
+        body = await self._arequest(
```
jperez999, line 668: [P2] Give the Python client the agentic timeout

`aagentic_answer()` uses `_arequest`, whose `AsyncClient` timeout is fixed at 300 seconds, while `AgenticConfig.request_timeout_s` defaults to 1,800 seconds and the gateway/MCP paths use the longer agentic timeout. A legitimate run lasting more than five minutes therefore fails only through this newly advertised Python interface while the service continues processing it. Please allow a per-call timeout (or a client-level agentic timeout) and use it here.

nemo_retriever/docs/cli/README.md
```diff
@@ -348,8 +354,20 @@ fusion) -> SelectionAgentOperator -> ranked results`:
   emits ranked document IDs. Those IDs are then rehydrated from the retrieval-hop
```
jperez999, line 354: [P2] Update the published agentic workflow docs

This updates only the in-repo CLI reference, not `docs/docs/extraction/`, which the repository marks as the published source of truth. In particular, `workflow-agentic-retrieval.md` and `agentic-retrieval-concept.md` still describe agentic mode as document ranking only, list only `agentic_query` for MCP, and describe `/v1/answer` as a separate non-tool-calling path. Since this PR description says the public API is documented and adds REST, MCP, and Python-client contracts, those published pages need the answer-mode behavior and response differences in the same change.

nemo_retriever/src/nemo_retriever/query/agentic.py
```diff
@@ -457,6 +471,112 @@ def unload(self) -> None:
...
+        if len(set(caller_query_ids)) != len(caller_query_ids):
+            raise ValueError("query_ids must be unique.")
+
+        with self._hit_cache_lock:
+            self._hit_cache.clear()
+
+        from nemo_retriever.operators.graph_ops.react_agent_operator import ReActAgentOperator
+
+        per_hop_top_k = max(AGENTIC_RETRIEVER_TOP_K, int(self._cfg.top_k))
+        react_operator = ReActAgentOperator(
```
jperez999: [P3] Reuse the ReAct operator construction

This block duplicates the near-identical `ReActAgentOperator` construction in `retrieve_with_usage`, including every LLM, retrieval, and concurrency setting. Answer mode should ideally add only the mode/output branch; keeping two constructor lists makes future configuration changes easy to apply to one path but not the other. A small private builder such as `_build_react_operator(mode=...)`, used by both methods, would shrink this addition and keep select/answer configuration aligned.

jperez999 (changes requested):

nemo_retriever/src/nemo_retriever/_agentic/nemo_agent/agent.py
```diff
@@ -287,6 +290,24 @@ async def _dispatch_tool_call(
...
+            if isinstance(citations, list):
+                cited_ids = {
```
jperez999: [P2] Apply the Black-generated formatting

The required pre-commit job is currently failing because Black rewrites this set comprehension onto one line. Please apply the formatter output and rerun the check; the unit-test and Docker checks are otherwise green.

jperez999 (approved):

PR #2727, opened 2026-09-18
fix(extraction): validate OCR dependencies while preserving audio opt…

KyleZheng1284:
Bring the follow-up fixes from PR #2723 to main. Reject Page Elements opt-outs that would leave document OCR without detections, preserve audio and Nemotron Parse exemptions, and retain main-specific Parse model validation. Includes regression tests and documentation.

## Checklist


sylvesterkaczmarek (commented):
The audio exemption is global to ExtractParams, but extraction_mode="auto" can use the same params for a PDF in a mixed batch. With method="audio" and use_page_elements=False, _run_pdf_pipeline takes the non-Parse path and default table/chart extraction can reach OCR without detections. The image path has a runtime guard for this, but the PDF path does not. Could the audio exemption be scoped to audio inputs or guarded in the PDF path too?

iamdevdhanush:
@sylvesterkaczmarek  
Thanks, that's a good catch. You're right that the exemption currently lives at the `ExtractParams` level, so with `extraction_mode="auto"` the same params could be reused for a PDF in a mixed batch. That would allow the PDF path to bypass the validation while still reaching OCR-dependent stages without Page Elements.

I'll scope the exemption to audio inputs (or add the equivalent runtime guard in `_run_pdf_pipeline()` if that's the cleaner place to enforce it) so the PDF path matches the image-path behavior.

PR #2730, opened 2026-09-21
feat(pdf): add opt-in nested Form image extraction

ChrisJar:
## Summary

- add extract_nested_images and matching CLI flags as an opt-in PDFium extraction mode
- preserve default rendered-region output while appending each nested raster IMAGE occurrence
- decode intrinsic source bitmaps, preserve repeated placements, and transform Form coordinates into normalized page space
- document the rendered-region versus PDF IMAGE XObject contract

## Validation

- pre-commit run --all-files
- 35 focused extraction, SDK, CLI, planning, model, and service tests
- fixture verification: 1 default composite region; 18 nested occurrences; 5 distinct source bitmaps at 256x256 or 512x512; normalized page bounding boxes
- git diff --check


nemo_retriever/src/nemo_retriever/common/api/util/pdf/pdfium.py
```diff
@@ -327,8 +333,171 @@ def extract_simple_images_from_pdfium_page(page, max_depth):
...
+        raise RuntimeError(f"Nested PDF image source is {source_size:,} bytes; limit is {max_source_bytes:,} bytes")
+    if source_size <= 0:
+        return None
+
+    source_data = image_obj.get_data(decode_simple=False)
+    if len(source_data) != source_size:
+        raise RuntimeError(f"Nested PDF image source size changed from {source_size:,} to {len(source_data):,} bytes")
+    metadata = image_obj.get_metadata()
+    digest = hashlib.sha256(memoryview(source_data)).digest()
+    return (
```
jperez999: [P1] Do not cache decoded pixels by this incomplete stream fingerprint

The decoded bitmap is not determined by `(size, filters, bits_per_pixel, colorspace enum, raw bytes)` alone: `/Decode`, `/DecodeParms`, indexed/ICC color-space definitions, and `/Mask`/`/SMask` can change PDFium's output. I reproduced this with two nested 10x10 DeviceGray images containing identical raw bytes but opposite `/Decode` arrays; this function returned identical payloads because the second image hit this cache, although PDFium should decode it inverted. Please use an identity/render-state key that guarantees equivalent decoding or, to keep this change small and correct, remove cross-object payload reuse. A regression test with differing `/Decode` arrays would cover the failure.

nemo_retriever/src/nemo_retriever/common/api/util/pdf/pdfium.py
```diff
@@ -327,8 +333,171 @@ def extract_simple_images_from_pdfium_page(page, max_depth):
...
+    max_source_bytes: int,
+) -> tuple[tuple[int, int], tuple[str, ...], int, int, bytes] | None:
+    """Build a source-image key without decoding its bitmap."""
+    source_size = int(pdfium_c.FPDFImageObj_GetImageDataRaw(image_obj, None, 0))
+    if source_size > max_source_bytes:
+        raise RuntimeError(f"Nested PDF image source is {source_size:,} bytes; limit is {max_source_bytes:,} bytes")
+    if source_size <= 0:
+        return None
+
+    source_data = image_obj.get_data(decode_simple=False)
```
jperez999: [P1] Bound the aggregate raw-stream work

`max_source_bytes` caps one occurrence, but `get_data()` allocates and copies the full stream for every placement before the payload-cache lookup. With the current defaults, 256 repeated placements of a 64 MiB source can cause roughly 16 GiB of copying and hashing on one page even though the image is decoded only once; `max_decoded_pixels` does not constrain this path. Please add an aggregate source-byte budget or avoid rereading the source for repeated placements.

nemo_retriever/src/nemo_retriever/operators/extract/pdf/extract.py
```diff
@@ -390,6 +393,25 @@ def pdf_extraction(
                         except Exception:
                             pass  # Image extraction failure should not crash the pipeline.
 
+                        if extract_nested_images:
+                            nested_images = extract_nested_simple_images_from_pdfium_page(page)
```
jperez999: [P1] Preserve the base page when supplemental extraction fails

This helper intentionally raises for a count/pixel/source budget, deep Forms, or an unsupported bitmap. Those exceptions reach the outer page boundary, which replaces the entire row with `_error_record` and discards text, the page image, and the default rendered-region images already extracted above. That conflicts with the PR description's promise to preserve default output while appending nested occurrences. Please report the nested-image failure without throwing away the successfully extracted base page (or explicitly define and document a different partial-result contract).

nemo_retriever/src/nemo_retriever/operators/extract/pdf/extract.py
```diff
@@ -390,6 +393,25 @@ def pdf_extraction(
...
+                            for img in nested_images:
+                                max_w = float(img.max_width) if img.max_width else 1.0
+                                max_h = float(img.max_height) if img.max_height else 1.0
+                                x0, y0, x1, y1 = img.bbox
+                                detected_images.append(
```
jperez999: [P3] Reuse the image-record conversion

This normalization/schema block duplicates lines 377-392. A small local helper such as `_base64_image_to_record(img)` used by both paths would reduce this feature diff and prevent the two image-output contracts from drifting.

jperez999 (changes requested):

jperez999 (approved):

